PR #85, opened 2017-01-12
Fixes unwanted resizing of application window

darko-mijic:
Application window was resizing to default size of 1024 x 768 pixels when navigating from wallets to settings and the other way back. This PR fixes that issue.


DominikGuzei:
Short and sweet, thanks! 👍

PR #87, opened 2017-01-13
Feature/improve backend transaction integration

DominikGuzei:
This PR finishes the backend integration for sending and receiving money.

**Last thing missing:**
Currently the api endpoint `searchHistory` only returns the array of search results, but we need the total available amount of transactions to calculate if "there are any transactions" for a wallet without fetching all records. As soon as @akegalj has this implemented we can finalize this.


DominikGuzei:
@darko-mijic this PR is now ready for final review & merge

PR #89, opened 2017-01-16
Feature/demo

DominikGuzei:
This PR was intended for preparation of the demo but i think we should probably keep the current changes so that the app is always deployable. E.g: login / accounts does not exist yet, so we should not have it enabled at all.


darko-mijic:
Yes, we should have only deployable, backend supported features in the UI.

domenkozar:
I will need this branch further on to generate newer installers.

Could we rename it to `cardano-sl-0.1` so it matches the backend branch?

DominikGuzei:
@domenkozar the branch is not gone but merged into `develop`, you can branch off with `cardano-sl-0.1` and continue there 😉 

In general we tend to avoid to make other branches than develop / master the main branches.

domenkozar:
As soon as we start doing releases (as of yesterday), we need to move closer to git-flow branching, which allows one branch per release cycle.

PR #91, opened 2017-01-17
Feature/wallet backup

darko-mijic:
This PR implements the wallet backup feature.


webpack/webpack.config.base.js
```diff
@@ -37,7 +37,7 @@ module.exports = validate({
 
   plugins: [
     new webpack.DefinePlugin({
-      CARDANO_API: process.env.CARDANO_API || 1,
+      CARDANO_API: process.env.CARDANO_API || 0,
```
darko-mijic: We have to revert this. `CARDANO_API=false` is not working since it has falsy value.

DominikGuzei: you have to use numbers for boolean env flags like `CARDANO_API=0`

PR #100, opened 2017-02-02
CSL-705: fix icon path passed to electron-package

jmitchell:
Manually ran `npm run package` locally with the updated path and the resulting
Daedalus.exe in the release directory ends up with the specified icon.
Without this change the program had the default Electron icon.

In case there's concern that the path doesn't correspond to an actual file, the
.ico extension (in Windows) is appended in webpack/package.js.

daedulus-windows-icon

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -28,7 +28,7 @@ install:
 test_script:
   - SET DAEDALUS_VERSION=%APPVEYOR_BUILD_VERSION%
   # Package frontend
-  - npm run package -- --icon icons/64x64
+  - npm run package -- --icon installers/icons/64x64
   - cd installers
   # Install stack
   - curl -sSL -o stack.zip --insecure http://www.stackage.org/stack/windows-x86_64


darko-mijic:
Looks good. :)

domenkozar:
@jmitchell ah, yes. 

Icon is not taken in effect for the shortcut.

This is the installer built for this PR: https://ci.appveyor.com/project/jagajaga/daedalus/build/artifacts

And here is the code for generating the shortcut: https://github.com/input-output-hk/daedalus/blob/develop/installers/WindowsInstaller.hs#L32

I suspect the problem is in variable substitution, we could try hardcoding the full path to the iconfile.

To see the icon used: right click -> properties -> change icon.

jmitchell:
Thanks for bringing that to my attention, @domenkozar. https://github.com/input-output-hk/daedalus/pull/102 may resolve it, but it could potentially break other things. For now this will wait on the Appveyor build to finish.

PR #101, opened 2017-02-02
Make it easier to build the backend

domenkozar:
## Changes

- pinpoint cardano-sl in git submodule
- generated the bridge as a file instead of npm dance
- update README how to use
- add dev:backend to also launch backend with frontend in dev
- optionally install electron if it's not availabile (Nix provides
electron itself)

## Usage

```
  $ git submodule update --init
  $ nix-shell --run "npm run build-bridge"
  $ nix-shell --run "CARDANO_API=1 npm run dev:backend"
```

You will get Daedalus running with the backend as well.

## Future work

Once we get to [build the bridge using Nix](https://github.com/ElvishJerricco/nix-purescript-skeleton),
we can build Nix binaries for cardano-sl and the bridge, meaning you won't have to build locally anymore.

We could make this one command using npm run-in-all.

## Notes

We now require to use Nix for building the bridge, to be sure the build is reproducible :)

I wonder if we should write some docs how to use submodules correctly?

This depends on https://github.com/input-output-hk/cardano-sl/pull/24

cc @DominikGuzei @darko-mijic


jagajaga:
Awesome.

.gitignore
```diff
@@ -40,3 +40,6 @@ main.js.map
 
 # Generated translation files
 locales
+
+# Generated by the backend
+app/api/ClientApi.js
```
DominikGuzei: I think something like `CardanoWalletApi` (how will it work in the future with other currencies etc.?) would be more meaningful in the context of the frontend app. (E.g: we already have a `CardanoClientApi` in the same folder … which could be confusing).

package.json
```diff
@@ -17,7 +18,9 @@
     "package": "cross-env NODE_ENV=production node -r babel-register -r babel-polyfill webpack/package.js",
     "package-all": "npm run package -- --all",
     "postinstall": "node node_modules/fbjs-scripts/node/check-dev-engines.js package.json",
+    "preinstall": "which electron || npm install electron@1.4.12",
```
DominikGuzei: Hmm … I don't think it's a good idea to hard-code the dependency version of `electron` here!

.gitignore
domenkozar: Maybe `CardanBridge`?

package.json
domenkozar: Yeah, I couldn't find a better way. If we put it in other dependency lists, npm will just install it.

.gitignore
DominikGuzei: Yeah i like `CardanoBridge` 👍

package.json
DominikGuzei: What was the problem with letting it install with the other dependencies?

domenkozar: If you want to use system `electron` executable, npm doesn't allow for a way to specify that. Sadly we need this when using Nix, since electron is provided  by it.

The only way is to optionally install it if `electron` is missing.

Maybe I've missed some better method while googling :)

DominikGuzei: Ok understood, so in which situations could `electron` be missing?

domenkozar: If you install Daedalus on a system without electron already installed, which is the case quite often.

DominikGuzei: I thought nix is providing the `electron` executable? 😆

domenkozar: Well if you don't use Nix with daedalus, then it still needs to install `electron`. If you do use Nix, `preinstall` is a noop.

DominikGuzei: Ok, isn't the point of this PR to avoid having to support multiple ways to run Daedalus? In the PR description you're saying: "We now require to use Nix for building the bridge, to be sure the build is reproducible", so what is the scenario of not needing Nix for development?

domenkozar: We're going to recommend using Nix to install everything, but I still want others not using Nix to be able to install it. Trying to avoid Nix lock-in just in case :)

DominikGuzei: Im still against putting a version number here … this is not expected and will lead to confusion down the road (e.g: nix electron version might be different than this one). It's easy to forget to update it here. The only option i see is to put the electron version in a config file that can be loaded in both environments or behind an environment flag.

domenkozar: Nix provides only a specific version of electron. Currently `1.4.14`. I'm not sure how we can keep that in sync really with Nix except manually.

domenkozar: You specify it when specifying the whole package set here: https://github.com/input-output-hk/daedalus/pull/101/files#diff-5712e736e0de6ba170577f8472c398e9R1

We could add a note there to also bump electron in `pacakge.json`. What do you think? Sadly `package.json` doesn't allow comments.

DominikGuzei: Hmm … seems like the only option then. I just hope we won't run into "the latest electron version of nix is behind the real one but has a bug and we cannot upgrade" problems 😉 e.g: we are using `1.4.15` at the moment.

domenkozar: I'm NixOS core dev so easy peasy :)

domenkozar:
@darko-mijic @DominikGuzei pushed!

domenkozar:
@darko-mijic you need to follow the instructions in README (specifically `git submodule` step).

You'll need to switch the branch to `bridge-automation` in `./daedalus` for this PR to work until upstream PR is merged as well.

darko-mijic:
@domenkozar: I did all of that before reporting the error.

domenkozar:
@darko-mijic could you try again and first pull `cardano-sl` repo?

darko-mijic:
@domenkozar: `cardano-sl` repo is in place. 

As I can see the issue is in `cardano-sl/scripts/build-daedalus-bridge.sh` which is being executed from the root of the `daedalus` repo using npm :`"build-bridge": "cardano-sl/scripts/build-daedalus-bridge.sh"` so the path here is wrong: 
`cd daedalus && npm install && npm run build:prod`
When I change it to 
`cd cardano-sl/daedalus && npm install && npm run build:prod` it works but i get brdige compilation error:

```
ERROR in ./src/Daedalus/ClientApi.purs
Module build failed: Error: compilation failed
    at ChildProcess.<anonymous> (/Users/darkomijic/iohk/daedalus/cardano-sl/daedalus/node_modules/purs-loader/lib/Psc.js:48:16)
    at emitTwo (events.js:106:13)
    at ChildProcess.emit (events.js:191:7)
    at maybeClose (internal/child_process.js:877:16)
    at Socket.<anonymous> (internal/child_process.js:334:11)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at Pipe._handle.close [as _onclose] (net.js:498:12)
 @ ./src/index.js 7:36-72
```

So we need to update https://github.com/input-output-hk/cardano-sl/blob/master/scripts/build-daedalus-bridge.sh or create a new dedicated script for this.

darko-mijic:
@domenkozar: Ignore my previous comment. I was not using the correct `cardano-sl` branch, I missed the fact that I need to pull it manually to review this.

This almost works now. I am getting JS error `Uncaught TypeError: Cannot read property 'TYPED_ARRAY_SUPPORT' of undefined` from `CardanoBridge.js` now. 

I have added this to the bridge build script in order to remove previously generated bridge before generating it since that was causing issues before: 
https://github.com/input-output-hk/cardano-sl/pull/24/commits/e7c5954d704dc302440e2f75de8689b9a3c7735d

domenkozar:
I'll get back to this after 15th Feb.

neongreen:
@domenkozar any estimates? It'd be really nice to have an easy way to build Daedalus

domenkozar:
@neongreen I'm focusing on OSX installer atm since that's the priority. All we need to figure out here is to be able to import the generated file as a file, not as npm module.

PR #102, opened 2017-02-02
[CSL-705] Fix Windows install path (x86/x64)

jmitchell:
Daedalus.exe is a 64-bit application and therefore must be installed to
`C:\Program Files` rather than `C:\Program Files (x86)`, as explained in NSIS'
docs: http://nsis.sourceforge.net/Docs/Chapter4.html#varconstant

The change may also fix the Windows Desktop link icon, but that must be
verified.

### DIFF
--- a/installers/WindowsInstaller.hs
+++ b/installers/WindowsInstaller.hs
@@ -13,7 +13,7 @@ writeNSIS = do
     _ <- constantStr "Version" (str version)
     name "Daedalus $Version"                  -- The name of the installer
     outFile "daedalus-win64-$Version-installer.exe"           -- Where to produce the installer
-    installDir "$PROGRAMFILES\\Daedalus"   -- The default installation directory
+    installDir "$PROGRAMFILES64\\Daedalus"   -- The default installation directory
     installDirRegKey HKLM "Software/Daedalus" "Install_Dir"
     requestExecutionLevel Highest     
 


jmitchell:
Note this change may break other NSIS-related paths,. The installer eventually produced by Appveyor must be verified before merging.

jmitchell:
I uninstalled Daedalus using Windows' Programs and Features utility and then ran the recently built installer from https://ci.appveyor.com/project/IOHK/daedalus/build/0.1.254/artifacts. The Windows desktop link's icon path now refers to the correct path.

There was an odd issue, though, and I suspect it has to do with Windows caching the previous default icon. Immediately after installing the desktop shortcut showed the default windows application icon, even though the icon's path was correct and daedalus.exe's icon was correct. Without changing anything, I rebooted and then saw the expected desktop icon. I don't expect that would happen on a machine where Daedalus was never installed (and if it does, it's most likely a Windows quirk out of our hands).

I'm not yet familiar with testing the Daedalus, so I can't say for sure this doesn't break anything else. The application opens and apparently gets stuck on the "Joining network..." view. The debug console periodically reports the following:

```
http://127.0.0.1:8090/api/get_wallets Failed to load resource: net::ERR_CONNECTION_REFUSED
bundle.js:101943Uncaught (in promise) Error: AJAX request failed: GET http://127.0.0.1:8090/api/get_wallets(…)
bundle.js:101943 GET http://127.0.0.1:8090/api/get_wallets net::ERR_CONNECTION_REFUSED(anonymous function) @ bundle.js:101943(anonymous function) @ bundle.js:96047(anonymous function) @ bundle.js:96047(anonymous function) @ bundle.js:96047(anonymous function) @ bundle.js:96047(anonymous function) @ bundle.js:96047(anonymous function) @ bundle.js:96047(anonymous function) @ bundle.js:98076(anonymous function) @ bundle.js:98076(anonymous function) @ bundle.js:76705n @ bundle.js:74421(anonymous function) @ bundle.js:74421e.(anonymous function) @ bundle.js:74421n @ bundle.js:28430(anonymous function) @ bundle.js:28430t @ bundle.js:16062(anonymous function) @ bundle.js:28430e @ bundle.js:76705(anonymous function) @ bundle.js:20304t @ bundle.js:16062value @ bundle.js:20304(anonymous function) @ bundle.js:84066B @ bundle.js:6903a @ bundle.js:6903
bundle.js:101943 GET http://127.0.0.1:8090/api/get_wallets net::ERR_CONNECTION_REFUSED
```

jmitchell:
I've rebuilt the current `develop` branch locally and confirmed the same `ERR_CONNECTION_REFUSED` errors happen with it, so this change is unrelated to that.

Otherwise, this is ready for others to review/merge (cc: @domenkozar).

domenkozar:
I've verified this works, backend doesn't start since we haven't entered the new dht keys for initial peers.

PR #109, opened 2017-02-14
[DAEF-12] Enable unicode installer in Windows

jmitchell:
Without this option enabled the installer and its output are unreliable, esp.
with non-ASCII usernames.


domenkozar:
@jmitchell nice work! 

I'd actually rather see for us to use your fork of nsis for now via stack than add more code. Any downsides to that?

domenkozar:
Heck, let's merge this to get it done. We'll scrape the code soon anyway.

PR #111, opened 2017-02-14
[CSL-737] Add few fixes to launcher params

volhovM:
cc @domenkozar

### DIFF
--- a/installers/WindowsInstaller.hs
+++ b/installers/WindowsInstaller.hs
@@ -13,7 +13,7 @@ shortcutParameters ipdht = L.intercalate " " $
     [ "--node \"%PROGRAMFILES%\\Daedalus\\cardano-node.exe\""
     , "--node-log-path", "\"%APPDATA%\\Daedalus\\Logs\\cardano-node.log\""
     , "--wallet \"%PROGRAMFILES%\\Daedalus\\Daedalus.exe\""
-    , "--updater \"" <> updaterPath <> "\""
+    , "--updater \"" <> installerPath <> "\""
     , "--node-log-config", "log-config-prod.yaml" -- should it be just here, not in programfiles/appdata?
     , "--node-log-path", "\"%APPDATA%\\Daedalus\\Logs\\nodeLogStdout.log\""
     , "--node-timeout 5"
--- a/installers/WindowsInstaller.hs
+++ b/installers/WindowsInstaller.hs
@@ -20,7 +20,7 @@ shortcutParameters ipdht = L.intercalate " " $
     , (" -n " ++ (L.intercalate " -n " nodeArgs))
     ]
   where
-    installerPath = "%APPDATA%\\Daedalus\\Installer.msi"
+    installerPath = "%APPDATA%\\Daedalus\\Installer.exe"
     nodeArgs = [
       "--listen", "0.0.0.0:12100",
       "--keyfile", "\"%APPDATA%\\Daedalus\\Secrets\\secret.key\"",
--- a/installers/WindowsInstaller.hs
+++ b/installers/WindowsInstaller.hs
@@ -1,32 +1,34 @@
 module WindowsInstaller where
 
-import           Data.Maybe (fromMaybe)
-import           Data.Monoid ((<>))
-import qualified Data.List as L
+import qualified Data.List          as L
+import           Data.Maybe         (fromMaybe)
+import           Data.Monoid        ((<>))
 import           Development.NSIS
-import           System.Directory (renameFile)
+import           System.Directory   (renameFile)
 import           System.Environment (lookupEnv)
-import           Turtle (echo, procs)
+import           Turtle             (echo, procs)
 
 shortcutParameters :: [String] -> String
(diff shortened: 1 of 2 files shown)


installers/WindowsInstaller.hs
```diff
@@ -1,32 +1,34 @@
...
-        "--logs-prefix", "\"%APPDATA%\\Daedalus\\Logs\"",
-        "--db-path", "\"%APPDATA%\\Daedalus\\DB-0.2\"",
-        "--wallet-db-path", "\"%APPDATA%\\Daedalus\\Wallet-0.2\"",
-        "--wallet"
-        ] <> ("--peer" : (L.intersperse "--peer" ipdht))
+    [ "--node \"%PROGRAMFILES%\\Daedalus\\cardano-node.exe\""
+    , "--node-log-path", "\"%APPDATA%\\Daedalus\\Logs\\cardano-node.log\""
+    , "--wallet \"%PROGRAMFILES%\\Daedalus\\Daedalus.exe\""
+    , "--updater \"" <> updaterPath <> "\""
+    , "--node-log-config", "log-config-prod.yaml" -- should it be just here, not in programfiles/appdata?
```
volhovM: @domenkozar where should it point to? I think we should put a full path here, not a relative (from what dir?).

domenkozar: It's fine here, current directory points to `C:\Program Files\Daedalus`.

installers/WindowsInstaller.hs
```diff
@@ -1,32 +1,35 @@
...
-        "--keyfile", "\"%APPDATA%\\Daedalus\\Secrets\\secret.key\"",
-        "--logs-prefix", "\"%APPDATA%\\Daedalus\\Logs\"",
-        "--db-path", "\"%APPDATA%\\Daedalus\\DB-0.2\"",
-        "--wallet-db-path", "\"%APPDATA%\\Daedalus\\Wallet-0.2\"",
-        "--wallet"
-        ] <> ("--peer" : (L.intersperse "--peer" ipdht))
+    [ "--node \"%PROGRAMFILES%\\Daedalus\\cardano-node.exe\""
+    , "--node-log-path", "\"%APPDATA%\\Daedalus\\Logs\\cardano-node.log\""
+    , "--wallet \"%PROGRAMFILES%\\Daedalus\\Daedalus.exe\""
+    , "--updater \"" <> updaterPath <> "\""
```
domenkozar: Does this compile? Where does `updaterPath` come from?

installers/WindowsInstaller.hs
```diff
@@ -1,32 +1,35 @@
...
-        "--db-path", "\"%APPDATA%\\Daedalus\\DB-0.2\"",
-        "--wallet-db-path", "\"%APPDATA%\\Daedalus\\Wallet-0.2\"",
-        "--wallet"
-        ] <> ("--peer" : (L.intersperse "--peer" ipdht))
+    [ "--node \"%PROGRAMFILES%\\Daedalus\\cardano-node.exe\""
+    , "--node-log-path", "\"%APPDATA%\\Daedalus\\Logs\\cardano-node.log\""
+    , "--wallet \"%PROGRAMFILES%\\Daedalus\\Daedalus.exe\""
+    , "--updater \"" <> installerPath <> "\""
+    , "--node-log-config", "log-config-prod.yaml" -- should it be just here, not in programfiles/appdata?
+    , "--node-log-path", "\"%APPDATA%\\Daedalus\\Logs\\nodeLogStdout.log\""
```
jmitchell: This should be removed.

volhovM: For what reason?

domenkozar: It duplicates an arg 4 lines above.

jmitchell:
@volhovM: There's pretty good evidence this change caused a regression, although I haven't yet found a root cause. Using the latest installers I get the following error when I double click the Daedalus desktop shortcut.

daef-12 cardano-launcher exe error2

I've confirmed Appveyor build 0.1.332 (a10383e0) doesn't exhibit this error, however build 0.1.341 (34816aea) does. Both those builds used artifacts from the same Daedalus build (0.1.2908), so I suspect something in these changes is the cause.

Initially I suspected non-ASCII characters in paths were causing it because we saw a similar error in DAEF-12 (user's with non-ASCII usernames have non-ASCII `%APPDATA%` paths; special care must be taken when printing unicode from Haskell in Windows).

volhovM:
I don't think you're precisely right. Usually when we're getting problems with unicode output on windows, the error is the following: `hPutChar: invalid argument (invalid character)`, but i've checked it once again and removed all unicode chars from string literals in csl (just in case). This particular problem might be caused by wrong resourse allocation/releasing, but i've never encountered it. Please share logs if you have any.

jmitchell:
We discovered the extraneous `--node-log-path` was causing the error.

There was a subsequent unicode error, but it's also been resolved. The new `--updater` parameter led cardano-launcher.exe could to sometimes print a path with non-ASCII characters. The offending line has been commented out with a reference to DAEF-12.

Things are looking good now. Thanks!

PR #117, opened 2017-02-15
[CSL-531] Use NSIS DSL fork to inject unsupported options

jmitchell:
Until the ndmitchell/nsis#11 PR (or similar) is merged and published in the NSIS DSL package, we can use [my fork](https://github.com/jmitchell/nsis/tree/feature/escape-hatch) with the `injectLiteral` and `injectGlobalLiteral` functions. They're used in this PR to set `Unicode true` (replacing my earlier ugly hack) and the appropriate `VIProductVersion`.


domenkozar:
Thanks!

PR #119, opened 2017-02-15
[CSL-531] Build uninstaller before installer for code signing

jmitchell:
A previous iteration of this resulted in an uninstall.exe which worked as
expected, both via Windows' Uninstall Programs menu and as a standalone
executable (found in %TEMP%).

Additionally this includes an attempted yet untested fix for running the
uninstaller installer (tempinstaller.exe) in headless mode for CI purposes. The
obvious approach with Turtle failed for some reason (doesn't seem to be passing
"/S" argument as expected).


jmitchell:
I verified the hack to run `tempinstaller.exe` in silent (aka headless) mode worked. I also successfully installed and uninstalled using a local build made from this branch.

PR #120, opened 2017-02-15
Feature/Sync state on the loading screen

darko-mijic:


darko-mijic:
I have added loading wallets stage to loading screen since it takes time to have `wallets.first` in the state. Please verify after launch @DominikGuzei.

PR #125, opened 2017-02-16
Removes "Request a specific amount" button from the wallet receive page

darko-mijic:
Everything is in the title ;)


DominikGuzei:
Works as intended! 👍

PR #128, opened 2017-02-16
Fix switching to wallets after wallet creation and wallet restoring

darko-mijic:
- Removes walletsCache from the WalletsStore
- Fixes switching to wallets after wallet is created and restored


DominikGuzei:
Works great now! 👍

PR #135, opened 2017-02-18
Chore/remove default wallet

darko-mijic:
This PR introduces support for not having default wallet:
- loading screen logic was changed to support having no wallets
- new blank "no wallets" page was introduced if loading wallet data is complete and there are no wallets
- a user is forced to create or restore a wallet, a user can cancel those action and in that case, wallet dialog is shown again until one wallet has been created/restored
- messages introduced by Nikos which have <strong> elements are now properly formatted


app/api/CardanoClientApi.js
```diff
@@ -40,6 +40,7 @@ export default class CardanoClientApi {
   async getWallets() {
     console.debug('CardanoClientApi::getWallets called');
     const response = await ClientApi.getWallets();
+    response.shift(); // TODO: Remove this before merging
```
darko-mijic: This needs to be removed before we merge, backend is still returning default wallet so this has been introduced only for testing purpuses.

PR #136, opened 2017-02-18
Back to loading screen when disconnected

darko-mijic:
This PR introduces redirection to loading screen when the connection is lost.
It also improves router usage. We had warnings when pushing to the same route we were already on. This PR fixes that.

### DIFF
--- a/app/stores/AppStore.js
+++ b/app/stores/AppStore.js
@@ -1,5 +1,5 @@
 // @flow
-import { observable } from 'mobx';
+import { observable, computed } from 'mobx';
 import Store from './lib/Store';
 
 export default class AppStore extends Store {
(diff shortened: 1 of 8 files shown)


app/stores/AppStore.js
```diff
@@ -11,8 +11,13 @@ export default class AppStore extends Store {
     this.actions.goToRoute.listen(this._updateRouteLocation);
   }
 
+  @computed get currentRoute() {
```
DominikGuzei: Not sure if that's the best idea because it's just an alias to the router store without any benefit?

DominikGuzei:
Works perfectly 👍

app/stores/AppStore.js
darko-mijic: This looked ugly to me: `return router.location ? router.location.pathname === screenRoute : false;`
But your point is valid. I leave the decision to you. I will change it back if you want.

PR #138, opened 2017-02-19
Feature/Wallet key import dialog

darko-mijic:
This PR introduces importing wallet from the file with the key. 

screen shot 2017-02-19 at 18 07 15
screen shot 2017-02-19 at 18 07 17

### DIFF
--- a/app/components/wallet/key-import/WalletKeyImportDialog.js
+++ b/app/components/wallet/key-import/WalletKeyImportDialog.js
@@ -1,0 +1,124 @@
+// @flow
+import React, { Component, PropTypes } from 'react';
+import { observer } from 'mobx-react';
+import classnames from 'classnames';
+import Dialog from 'react-toolbox/lib/dialog/Dialog';
+import DialogCloseButton from '../../widgets/DialogCloseButton';
+import ReactToolboxMobxForm from '../../../lib/ReactToolboxMobxForm';
+import { defineMessages, intlShape } from 'react-intl';
+import FileUploadWidget from '../../widgets/forms/FileUploadWidget';
+import LocalizableError from '../../../i18n/LocalizableError';
+import styles from './WalletKeyImportDialog.scss';
+
+const messages = defineMessages({
+  headline: {
+    id: 'wallet.key.import.dialog.headline',
+    defaultMessage: '!!!Key import',
+    description: 'headline for "Key import" dialog.'
+  },
+  keyFileLabel: {
+    id: 'wallet.key.import.dialog.keyFileLabel',
+    defaultMessage: '!!!Upload your key',
+    description: 'Label for the certificate file upload'
+  },
+  keyFileHint: {
+    id: 'wallet.key.import.dialog.keyFileHint',
+    defaultMessage: '!!!Drop file here or click to choose',
+    description: 'Hint for the key file upload'
+  },
+  submitLabel: {
+    id: 'wallet.key.import.dialog.submitLabel',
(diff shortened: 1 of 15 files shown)


darko-mijic:
TODO: update wording since importing from key can import multiple wallets
TODO: finish handling the case when some wallets are imported and duplicate ones are not and present that to the user nicely

app/components/wallet/key-import/WalletKeyImportDialog.js
```diff
@@ -0,0 +1,122 @@
...
+      styles.component,
+      'WalletKeyImportDialog',
+      isSubmitting ? styles.isSubmitting : null
+    ]);
+
+    const actions = [
+      {
+        label: intl.formatMessage(messages.submitLabel),
+        primary: true,
+        disabled: keyFile.value === null || !keyFile.value.path,
```
DominikGuzei: I guess `keyFile.value instanceof File` would be clearer here

app/components/wallet/key-import/WalletKeyImportDialog.js
```diff
@@ -0,0 +1,122 @@
...
+        active
+      >
+
+        <div className={styles.keyUpload}>
+          <FileUploadWidget
+            {...keyFile.bind()}
+            selectedFile={keyFile.value}
+            onFileSelected={(file) => {
+              keyFile.onChange(file);
+            }}
```
DominikGuzei: Can probably be simplified to `onFileSelected={keyFile.onChange}` or at least to `onFileSelected={(file) => keyFile.onChange(file)}

app/stores/WalletsStore.js
```diff
@@ -194,6 +220,14 @@ export default class WalletsStore extends Store {
...
-  }
+  };
+
+  _patchWalletRequestWithWallet = async (wallet) => {
+    await this.walletsRequest.patch(result => {
+      const foundWallet = _.find(result, { id: wallet.id });
+      if (foundWallet) result.pop(foundWallet);
+      result.push(wallet);
+    });
+  };
```
darko-mijic: @DominikGuzei: We had a race condition here, polling wallets sometimes got the new wallet before we patched the request. Please take a look at this and merge if everything is ok.

darko-mijic: And thanks for a pedantic and detailed review :)

DominikGuzei: Always a pleasure! Why are you popping out the found wallet here … what is the difference between the the wallet returned by the import and the wallet from the request?

darko-mijic: I would assume that wallet returned by the import/restore will be more up-to-date than the cached one so I have decided to pop the cached one.

DominikGuzei: But in which cases is there a cached one when you restore/import?

darko-mijic: When we import new wallet is returned along with the others from the `getWallets`. And we are polling that endpoint. Client API is returning imported wallet from `importKey` as a convenience so we can redirect to it.

DominikGuzei: Ok that's what i thought … so technically there is no reason for this extra patching logic (it's mostly confusing) since they are exactly the same results.

- if you patch before the results are updated -> fine, that's the reason we do patching because it might take some time until the wallet arrives in other ways.
- if the patch is immediately overwritten, also fine – that's what an optimistic patch is there for (to be replaced) at some later point by the "real" api data.

PR #142, opened 2017-02-20
add proper error handling to wallet send form

DominikGuzei:
This PR adds proper api error handling to the wallet send form, so that we display errors like `Not enough money to make this transaction.` in the UI instead of failing silently. Also switches to the simple `send` backend api which does not take `title`, `description` etc.


darko-mijic:
I am not able to test this with the latest `cardano-node`. I have restored a wallet with 0 money in it and I am able to send Ada from it. The backend is not returning "not enough money to send" error. I end up sending money from wallet with zero money in it to the same wallet and I and up with money in that wallet:

screen shot 2017-02-21 at 14 32 21

@akegalj: has something changed on the backend?

DominikGuzei:
@darko-mijic yeah this is pretty buggy on the backend side. Sometimes I could rig it correctly by restarting the backend often enough until the correct error message came.

@akegalj if you send more money than the wallet has it throws a really strange 500 internal server error. So that error handling is broken on the backend side.

darko-mijic:
OK, thanks @DominikGuzei. I suggest we keep this open until we can verify the functionality introduced here.

darko-mijic:
I have verified this and it works great. I will close this PR and create a new one against the master branch since our workflow has been updated.

PR #144, opened 2017-02-20
[CSL-807] Sign cardano-node.exe and cardano-launcher.exe

jmitchell:
WACK's UAC checks suggested signing may resolve this issue, so I'm starting by ensuring the installed EXEs are signed. The installer and uninstaller are already getting signed.


jagajaga:
LGTM

PR #145, opened 2017-02-20
[CSL-705] Add icon to installer

jmitchell:


jagajaga:
Why image quality is so poor?

jmitchell:
@jagajaga: I used Paint's naive resize tool on the 256x256.png. I'll try installing Gimp and use a better downscaling method.

jmitchell:
@jagajaga: This is a bit better, but as far as I can tell we're constrained to using a 57px high bitmap.

PR #152, opened 2017-02-27
Fix relative syncing percentage

darko-mijic:
- fixes NaN when node get's in sync after first local network difficulty message is received

### DIFF
--- a/app/stores/NetworkStatusStore.js
+++ b/app/stores/NetworkStatusStore.js
@@ -78,7 +78,6 @@ export default class NetworkStatusStore extends Store {
     const { app, wallets } = this.stores;
     if (this.isConnected && this.isSynced && wallets.hasLoadedWallets && app.currentRoute === '/') {
       this.isLoadingWallets = false;
-      this._localDifficultyStartedWith = null;
       if (wallets.first) {
         this.actions.goToRoute({ route: wallets.getWalletRoute(wallets.first.id) });
       } else {
--- a/app/stores/NetworkStatusStore.js
+++ b/app/stores/NetworkStatusStore.js
@@ -1,6 +1,7 @@
 // @flow
 import { observable, action, computed } from 'mobx';
 import Store from './lib/Store';
+import Request from './lib/Request';
 
 export default class NetworkStatusStore extends Store {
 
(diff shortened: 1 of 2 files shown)


darko-mijic:
Please review this carefully @DominikGuzei and feel free to push improvements. I did this very tired.

georgeee:
Why there is no issue is in commits? Why did not you create one in YT?

app/stores/NetworkStatusStore.js
```diff
@@ -25,9 +26,17 @@ export default class NetworkStatusStore extends Store {
   }
 
   @computed get syncPercentage(): number {
-    if (this.networkDifficulty > 0 && this._localDifficultyStartedWith != null) {
+    if (this.networkDifficulty > 0 && this._localDifficultyStartedWith !== null) {
```
DominikGuzei: It was on purpose to check against `!= null` because that includes the check against `undefined`, otherwise `undefined` would be a valid value here!

app/stores/NetworkStatusStore.js
```diff
@@ -25,9 +26,17 @@ export default class NetworkStatusStore extends Store {
...
       const relativeNetwork = this.networkDifficulty - this._localDifficultyStartedWith;
+      // In case node is in sync after first local difficulty messages
+      // local and network difficulty will be the same (0)
+      console.log('networkDifficulty', this.networkDifficulty);
+      console.log('localDifficulty', this.localDifficulty);
+      console.log('relativeLocal', relativeLocal);
+      console.log('relativeNetwork', relativeNetwork);
+
+      if (relativeLocal === relativeNetwork) return 100;
       return relativeLocal / relativeNetwork * 100;
```
DominikGuzei: Im missing a check against `relativeNetwork === 0` here? (division by zero)

app/stores/NetworkStatusStore.js
darko-mijic: I know. `undefined` value should not be possible here. If it is I would handle that explicitly.

app/stores/NetworkStatusStore.js
darko-mijic: If my calculations are correct then division by zero should happen only when `(relativeLocal === relativeNetwork === 0)`. We can add this to improve code readability.

DominikGuzei: Ah yeah, because network difficulty can never be less than local. Either add a comment about that or the check 😉

DominikGuzei:
Tested and works correctly! 👍

PR #158, opened 2017-03-03
Chore/introduce skinnable input component

DominikGuzei:
This PR is the first step to use [react-polymorph](https://github.com/CodeAdventure/react-polymorph) components which are easier to customize than those from react-toolbox. As a first example I created a special `CurrencyInputSkin` for the wallet send form, that decorates the standard input with the currency name in the input field:

screenshot 2017-03-03 um 23 47 27

**NOTE:** Since `react-polymorph` is not yet published on npm, you have to clone (+ `npm run build` in the downloaded folder) and link it into this repo locally to try this PR.


DominikGuzei:
This is old stuff … so i will close it 😉

PR #159, opened 2017-03-04
[DAEF-42] Redesigned and improved ada redemption feature

darko-mijic:
This PR introduces redesigned Ada redemption feature with small improvements:
- instructions for the user
- uploaded certificate can be removed 
- if the certificate is uploaded user can not enter mnemonic manually

Manually entering redemption code:
screen shot 2017-03-04 at 18 45 16

Valid and parsable PDF uploaded:
screen shot 2017-03-04 at 18 45 39

Encrypted PDF uploaded:
screen shot 2017-03-04 at 18 45 32

Invalid PDF uploaded:
screen shot 2017-03-04 at 18 47 29


darko-mijic:
This is ready for review @DominikGuzei. I was hoping to have updated endpoint for Ada redemption before we review and merge this but that update will have be done in a separate PR.

app/components/widgets/forms/AdaCertificateUploadWidget.js
```diff
@@ -0,0 +1,90 @@
...
+    } = this.props;
+    const certificateIcon = isCertificateEncrypted ? certificateLockedIcon :
+      isCertificateInvalid ? certificateInvalidIcon : certificateNormalIcon;
+    return (
+      <div>
+        <div className={styles.label}>{label}</div>
+        <div className={styles.uploadBox}>
+          {
+            (() => {
+              if (isCertificateSelected) {
```
darko-mijic: Do you find this more readable than ternary operator @DominikGuzei? IIFE will definitely be more readable for complex if statements.

DominikGuzei: Hmm … i find it a bit verbose for just rendering two different options. I would always reduce "complex if statements" to a new variable that sums all the conditions up 😉

darko-mijic: Yes, I agree. I will remove this experiment ;)

DominikGuzei:
@darko-mijic im getting this error (only in the console, nothing is shown to user) when i try to submit a manually entered redeem code (that is wrong):
screenshot 2017-03-06 um 13 38 42

DominikGuzei:
I just had the case that the `redemption code could not be parsed from the certificate` error was shown but when i tried it again it worked fine -> so maybe we should update that error message to include "please try again".

darko-mijic:
Good catch @DominikGuzei. I will introduce validation for the redemption code.

darko-mijic:
OK, thanks. I have created an issue to expose redemption code validation on the Client API so we can use same validation in the future. For now we are checking the length so "Seed is not 32-byte long" error is not happening anymore.

PR #160, opened 2017-03-07
[DAEF-77] Fixes transaction ordering and displays transaction date

darko-mijic:
This PR fixes transaction ordering and introduces transaction time in the transaction list. 

screen shot 2017-03-07 at 07 03 57


DominikGuzei:
Thanks @darko-mijic 👍

PR #162, opened 2017-03-07
WIP - [CSLA-62] Use $INSTDIR for Windows shortcut.

jmitchell:
Should fix installations to non-default paths.

This doesn't yet make the shortcut target a batch file instead. Manual testing of that approach raised some other issues. Investigating that while this minimal fix builds.

### DIFF
--- a/installers/Installer.hs
+++ b/installers/Installer.hs
@@ -9,7 +9,7 @@ import qualified MacInstaller
 
 main :: IO ()
 main = do
-  echo $ "Generating installer for " <> T.pack os <> "-" <> T.pack arch
+  echo . unsafeTextToLine . pack $ "Generating installer for " <> os <> "-" <> arch
   case os of
     "linux" -> fail "No installer yet"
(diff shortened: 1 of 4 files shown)


jmitchell:
This initial approach didn't work. `$INSTDIR` appears verbatim in the shortcut link itself.

jmitchell:
The generated nsi script escapes `$INSTDIR`, e.g. `$\"$\"$$INSTDIR\cardano-node.exe$\"$\"`

domenkozar:
Yeah I remember that's why I hardcoded it :)

domenkozar:
@jmitchell can we write `.bat` file using nsis variables?

jmitchell:
@domenkozar: Perhaps, although at the moment I'm not focused on using `.bat`. Running `cardano-launcher.exe` directly from cmd or batch files tend to result in an `hPutChar: invalid argument (Bad file descriptor)`. I need to diagnose/fix that first.

PR #167, opened 2017-03-15
Fix/daef 89 fix sidebar toggle tests

DominikGuzei:
**NOTE:** you have to run backend at commit `c88ead10caee086eb667a6130937e2f460bf5870` to test this!

This PR fixes the sidebar toggle acceptance tests and adjusts them to the new UX behavior. (Sidebar is never hidden, only sub menus can be toggled). It fixes / improves some parts of the stores / actions logic on the way.

**One important change:** i re-introduced `daedalus.reset()` before each test case and added caching for the `networkStatus` store, to avoid wasting time waiting on the difficulty messages from the backend.


darko-mijic:
Dev mode is working on testnet 0.3, reviewing this PR now.

PR #179, opened 2017-03-23
Fix/daef 100 no transaction message styling

nikolaglumac:
This PR introduces VerticalFlexContainer as a fix for 'No transactions message' vertical positioning issue on Transactions page.

transactions-page


darko-mijic:
I like this change. We are removing inline styles from container components. Container components should not be concerned by styling (how things look). That is the reason why we avoided creating associated stylesheets for container components.

The only thing I am not so sure about is the name of the presentational component `VerticalFlexContainer`. We should use domain language here. Flex is implementational detail. I would also like to avoid word Container in the name since it is misleading. I suggest something like `VerticalContent`. On the other hand, things are much clearer with the original name `VerticalFlexContainer` and this is utility component, so there is no domain specific language to apply here. 
I need to give this some thought. Suggestions are welcome. :)

nikolaglumac:
This is an eternal debate :)
I'm pro current `VerticalFlexContainer` naming for two reasons:
- `VerticalFlex` in my mind has important advantage in respect to just `Vertical` as it declares that this wrapper will fill available height of it's parent and not just expand based on it's contents (but this view might be result of my CSS-mindset),
- `Container` is somehow more appropriate than `Content` as this component only wraps it's contents - so it is Container > Content :)

darko-mijic:
You obviously gave this some thought. I agree with all of the arguments. Consumers of this component are developers and components solves a technical issue, so the name is right. And we would never use the word Container in the name of the component that serves the role of the container component. I am still glad we had this conversation. :)

PR #182, opened 2017-03-23
Feature/daef 101 copy address to clipboard

tomothespian:
This PR adds introduces copy-to-clipboard icon notification message for copy wallet address feature on Receive page.

screen shot 2017-03-23 at 18 09 24


DominikGuzei:
Reviewing this now

app/stores/WalletsStore.js
```diff
@@ -262,4 +266,19 @@ export default class WalletsStore extends Store {
...
+    this.isWalletAddressCopyNotificationVisible = false;
+  }
+
+  _onShowWalletAddressCopyNotification = action(() => {
+    if (this._hideWalletAddressCopyNotificationTimeout) {
+      clearTimeout(this._hideWalletAddressCopyNotificationTimeout);
+    }
+    this._hideWalletAddressCopyNotificationTimeout = setTimeout(
+      this._hideWalletAddressCopyNotification,
+      10000
```
DominikGuzei: Please move this magic constant for hiding delay to a config value. I see that we don't have anything like that yet … so let's create `app/config.js` and add 
```js
export default {
  wallets: {
    ADDRESS_COPY_NOTIFICATION_DURATION: 10000,
  }
};
```

app/components/widgets/NotificationMessage.js
```diff
@@ -0,0 +1,35 @@
...
+    ]);
+
+    return (
+      <div className={notificationMessageStyles}>
+
+        {icon && <img className={styles.icon} src={icon} role="presentation" />}
+
+        {(
+          // eslint-disable-next-line react/no-danger
+          <span className={styles.message} dangerouslySetInnerHTML={{ __html: message }} />
```
tomothespian: @DominikGuzei yep - this is much better! Fixed in my latest commit ;)

app/stores/WalletsStore.js
tomothespian: @DominikGuzei app/config.js added in my latest commit!

PR #185, opened 2017-03-25
[CSLA-62] Fix Daedalus shortcut link in Windows

jmitchell:
I've tested the changes with a local build and verified that

1. daedalus.bat installs to the correct path
2. the shortcut targets the batch file instead of cardano-launcher.exe + args
3. running the shortcut works
4. the uninstaller removes daedalus.bat and the shortcut

Before merging I'd like to confirm the version built on appveyor works as well.

PR #189, opened 2017-03-27
[CSLA-62] Avoid long target string in Windows shortcut

jmitchell:
Replaces PR #185.

### DIFF
--- a/installers/WindowsInstaller.hs
+++ b/installers/WindowsInstaller.hs
@@ -12,19 +12,28 @@ import           Turtle.Line          (unsafeTextToLine)
 
 import           Launcher
 
-shortcutParameters :: String
-shortcutParameters = launcherArgs $ Launcher
-  { nodePath = "%PROGRAMFILES%\\Daedalus\\cardano-node.exe"
-  , nodeLogPath = "%APPDATA%\\Daedalus\\Logs\\cardano-node.log"
-  , walletPath = "%PROGRAMFILES%\\Daedalus\\Daedalus.exe"
-  , installerPath = "%APPDATA%\\Daedalus\\Installer.exe"
-  , runtimePath = "%APPDATA%\\Daedalus\\"
-  }
+launcherScript :: String
+launcherScript = unlines
+  [ "@echo off"
+  , ""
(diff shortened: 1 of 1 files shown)


installers/WindowsInstaller.hs
```diff
@@ -11,21 +11,30 @@ import           Turtle             (echo, proc, procs)
...
-  , installerPath = "%APPDATA%\\Daedalus\\Installer.exe"
-  , runtimePath = "%APPDATA%\\Daedalus\\"
-  , reportServerUrl = "http://35.156.164.19:8080"
-  , systemStart = Nothing
-  }
+launcherScript :: String
+launcherScript = unlines
+  [ "@echo off"
+  , ""
+  , "powershell.exe -Command \"(Get-ItemProperty HKLM:\\SOFTWARE\\Wow6432Node\\Daedalus).Install_Dir\" > install_dir.txt"
```
georgeee: Are you sure PowerShell is present in all WIndows 7+?

jmitchell: Great point. I'll find a more portable solution.

jmitchell:
@georgeee: I verified this installs, runs, and uninstalls properly in Win Server 2012 r2, both to default and non-default directories. Do you mind doing some additional testing on other Windows versions?

fersel:
@jmitchell I tested it and though this works fine using the shortcut I don't think the current approach is good enough. Two main things I'm concerned about:
1) putting .bat file in %appdata%. It should stay in the installation folder
2) not writing down full paths in the .bat file. That makes us depend on the shortcut (which might be deleted or whatever), i.e. if you try to launch via .bat directly that *won't* work.
(these problems are somewhat connected)

Also:
* each version installation will override the prev version shortcut (you can't change that during the installation process), that's inconvenient.
* can we have different folders in %appdata% for different app versions as well?

I can describe this more detailed if needed, but I'd prefer to do that in the ticket.

jmitchell:
@fersel, I've responded in the ticket. Thank you for testing it and giving your feedback.

domenkozar:
You should now make a PR against master, since some parts changed a bit. We will be rolling out 0.4 soon so that train is gone :)

PR #190, opened 2017-03-28
Korean translation

hotblade:
I found that the locale file for Korean have not beet translated yet.
There are about a few hundred ADA holders in Korea.

So I translated it's contents into Korean and I hope it helps.
While doing it, I referred to Japanese translation(ja-JP.json) to understand correct context.


nikolaglumac:
Thank you for your contribution! Unfortunately, at this moment in time, we can not accept this translation as it would be hard to maintain it without internal translation/audit process.

PR #191, opened 2017-03-28
Feature/daef 107 integrate language selection with the backend

tomothespian:
This PR introduces language selection backend integration.

### DIFF
--- a/app/stores/AppStore.js
+++ b/app/stores/AppStore.js
@@ -1,6 +1,7 @@
 // @flow
 import { observable, action, computed } from 'mobx';
 import Store from './lib/Store';
+import Request from './lib/Request';
 import CachedRequest from './lib/CachedRequest';
 import globalMessages from '../i18n/global-messages';
 
@@ -15,7 +16,8 @@ export default class AppStore extends Store {
     { value: 'hr-HR', label: globalMessages.languageCroatian },
   ];
 
-  @observable profileLocaleRequest = new CachedRequest(this.api, 'setUserLocale');
+  @observable getProfileLocaleRequest = new CachedRequest(this.api, 'getUserLocale');
+  @observable setProfileLocaleRequest = new Request(this.api, 'setUserLocale');
 
   setup() {
     this.actions.router.goToRoute.listen(this._updateRouteLocation);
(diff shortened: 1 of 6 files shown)


app/stores/AppStore.js
```diff
@@ -32,23 +34,30 @@ export default class AppStore extends Store {
...
+
+  @computed get hasLoadedCurrentLocale(): bool {
+    return this.getProfileLocaleRequest.wasExecuted;
   }
 
   @computed get isCurrentLocaleSet(): bool {
-    return this.profileLocaleRequest.wasExecuted;
+    if (this.getProfileLocaleRequest.result == null) return false;
+    return (this.getProfileLocaleRequest.wasExecuted && this.getProfileLocaleRequest.result) ||
+           (this.setProfileLocaleRequest.wasExecuted && this.setProfileLocaleRequest.result);
```
DominikGuzei: I see why you are doing it like that but i think it is cleaner if you invalidate and execute the `getProfileLocaleRequest` anytime the `setProfileLocaleRequest` was executed like this (in `_updateLocale` method):

`this.setProfileLocaleRequest.execute(locale).then(() => this.getProfileLocaleRequest.invalidate().execute())`

This way everything stays declarative and you don't need the complicated check (you can just ask for `this.getProfileLocaleRequest.result` to see if `isCurrentLocaleSet`)

app/stores/AppStore.js
```diff
@@ -32,23 +34,30 @@ export default class AppStore extends Store {
 
   @computed get currentLocale(): string {
     if (!this.isCurrentLocaleSet) {
+      this.getProfileLocaleRequest.execute();
       return 'en-US';
     }
-    return this.profileLocaleRequest.result;
+    return this.getProfileLocaleRequest.result || this.setProfileLocaleRequest.result;
```
DominikGuzei: Here it's cleaner to only return the `this.getProfileLocaleRequest.result`

app/stores/AppStore.js
nikolaglumac: @tomothespian and I discussed this yesterday and were thinking to do it the way you suggested BUT we wanted to avoid additional call to getProfileLocaleRequest  since setProfileLocaleRequest is returning saved locale already ;)
But I agree that it will be much nicer the way you suggested!

DominikGuzei: True, you're right - there is an additional improvement you can make:

```
this.setProfileLocaleRequest.execute(locale).then(() => this.getProfileLocaleRequest.invalidate().patch(() => locale)
```
This patches the result of the `getProfileLocaleRequest` and marks it as dirty (so it is actually re-fetched the next time `execute` is called. But it avoids an immediate additional call 😉

darko-mijic:
Reviewing this now

tomothespian:
@DominikGuzei I have updated eslintrc file and everything works now :) Thanks :)

app/stores/AppStore.js
```diff
@@ -51,18 +52,21 @@ export default class AppStore extends Store {
...
+  _redirectToLoadingScreenWhenDisconnected = () => {
+    if (!this.stores.networkStatus.isConnected) {
+      this.actions.router.goToRoute({ route: '/' });
+    }
+  };
+
+  @action _updateLocale = async ({ locale }: { locale: string }) => {
+    await this.setProfileLocaleRequest.execute(locale);
+    this.getProfileLocaleRequest.invalidate().patch(() => locale);
   };
```
darko-mijic: @tomothespian: async await syntax makes this cleaner so I have used it here.

app/stores/AppStore.js
```diff
@@ -51,18 +52,21 @@ export default class AppStore extends Store {
...
 
-  @action _updateLocale = ({ locale }: { locale: string }) => {
-    this.setProfileLocaleRequest.execute(locale).then(() => (
-      this.getProfileLocaleRequest.invalidate().patch(() => locale)
-    ))
-    .catch(action((error) => {
-      this.error = error;
-    }));
+  _redirectToLoadingScreenWhenDisconnected = () => {
+    if (!this.stores.networkStatus.isConnected) {
```
darko-mijic: I have introduced another reaction @tomothespian to cover the case when we are not client is not connected to node or connection is lost after language selection screen is displayed.

PR #194, opened 2017-03-28
Feature/daef 109 new wallet setting units ada or lovelaces

nikolaglumac:
This PR introduces wallet settings page with unit (ada or lovelace) selection.

settings-page


nikolaglumac:
@darko-mijic as agreed InlineEditingDropdown component exceeds the scope of this task so I have reverted to using standard Dropdown component and we will finish development of InlineEditingDropdown in a separate task.

DominikGuzei:
@nikolaglumac sorry for the lack of solid requirements, it seems like we won't need different units in the app for longer time, so i will close this and remove that feature from the other PR

PR #195, opened 2017-03-28
[daef-57] add wallet deletion feature

DominikGuzei:
This PR adds the wallet deletion feature to the settings page:

screenshot 2017-03-28 um 18 36 23

screenshot 2017-03-28 um 18 36 31

screenshot 2017-03-28 um 18 36 46

screenshot 2017-03-28 um 18 36 54

(note: i made the checkbox red after taking the screenshots)

## Additionally this PR adds several re-usable improvements:
- Introduced `UiDialogsStore` to manage active dialog visibility and data
- The data handling of dialogs is greatly simplified with this change (no need to store properties on other stores)
- Used a dedicated container component for the delete wallet dialog -> this means the other containers remain light-weight and we don't have to pass around so much data (please follow that in future PRs and let's refactor existing dialogs to this pattern.
- Fixed the styling / proportions of dialogs in our app

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -22,6 +22,7 @@
     "promise/no-native": 0,
     "react/jsx-no-bind": "off",
     "react/jsx-filename-extension": ["error", { "extensions": [".js", ".jsx"] }],
+    "react/jsx-closing-bracket-location": 1,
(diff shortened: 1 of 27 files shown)


nikolaglumac:
Reviewing this now

PR #200, opened 2017-03-29
Feature/daef 124 wallet settings with changing language

tomothespian:
This PR introduces general wallet setting page with language selection dropdown.
general settings page


app/components/settings/categories/GeneralSettings.js
```diff
@@ -0,0 +1,85 @@
...
+  },
+});
+
+@observer
+export default class GeneralSettings extends Component {
+
+  static propTypes = {
+    languages: MobxPropTypes.arrayOrObservableArrayOf(PropTypes.shape({
+      value: PropTypes.string.isRequired,
+      label: PropTypes.object.isRequired,
```
DominikGuzei: It might be a good idea to introduce a prop type for the expected translations schema here.
You can put it into `app/propTypes.js`:

```js 
export const translationMessageParams = PropTypes.shape({
  id: PropTypes.string.isRequired,
  defaultMessage: PropTypes.string.isRequired,
  description: PropTypes.string,
});
```

app/components/settings/categories/GeneralSettings.scss
```diff
@@ -0,0 +1,25 @@
...
+      line-height: 1.21;
+    }
+    p {
+      color: #5e6066;
+      font-family: $font-regular;
+      font-size: 16px;
+      line-height: 22px;
+      margin-top: 1em;
+    }
+  }
```
DominikGuzei: Why are the `h1` and `p` styles used here? I guess those are copied over from the old settings screens? At the moment we don't have any text and it's unclear if we will need something like that.

app/containers/settings/categories/GeneralSettingsPage.js
```diff
@@ -0,0 +1,44 @@
...
+
+@inject('stores', 'actions') @observer
+export default class GeneralSettingsPage extends Component {
+
+  static propTypes = {
+    actions: PropTypes.shape({
+      profile: PropTypes.shape({
+        updateLocale: PropTypes.func.isRequired,
+      }),
+    }),
```
DominikGuzei: Please add `isRequired` to everywhere, so that potential problems are caught when things are wired wrongly

app/containers/settings/categories/GeneralSettingsPage.js
```diff
@@ -0,0 +1,44 @@
...
+      app: PropTypes.shape({
+        setProfileLocaleRequest: PropTypes.instanceOf(Request).isRequired,
+        LANGUAGE_OPTIONS: PropTypes.array.isRequired,
+        currentLocale: PropTypes.string.isRequired,
+      }).isRequired,
+    }).isRequired,
+  };
+
+  onSelectLanguage = (values: { locale: string }) => {
+    console.debug('SELECT LAGUAGE SUBMIT: ', values);
```
DominikGuzei: Please remove this debug message because it's not very relevant

tomothespian:
@DominikGuzei all issues are fixed in my latest commit :)

PR #206, opened 2017-03-30
Feature/daef 129 Testnet label

tomothespian:
This PR introduces testnet label on the right of walet top bar.
screen shot 2017-03-30 at 17 04 07


DominikGuzei:
Reviewing this now

PR #219, opened 2017-04-04
[DAEF-138] convert lovelace amounts from backend into ada on frontend

DominikGuzei:
This PR fixes the current bug that we treat the lovelace amounts from backend like ADA in the frontend. It introduces the [bignumber.js](https://github.com/MikeMcl/bignumber.js/) package to work with arbitrary numbers and formatting.

screenshot 2017-04-04 um 14 23 22


darko-mijic:
Reviewing this now

PR #226, opened 2017-04-06
Merge Cardano sl 0.4 rc back into master

DominikGuzei:


domenkozar:
This one should be rebased on top of `cardano-sl-0.4-rc` and we can start testing it again :)

PR #239, opened 2017-04-13
Feature/daef 155 entering decimal points in send money form

tomothespian:
This PR introduces Ada input field and ada-to-lovelaces conversion logic.

ada-input-1
ada-input-2

### DIFF
--- a/app/i18n/locales/en-US.json
+++ b/app/i18n/locales/en-US.json
@@ -121,7 +121,7 @@
   "wallet.send.form.description.hint": "You can type a message here, if you want.",
   "wallet.send.form.description.label": "Description",
   "wallet.send.form.errors.invalidAddress": "Please enter a valid address.",
-  "wallet.send.form.errors.invalidAmount": "Please enter a valid amount (positive integer).",
+  "wallet.send.form.errors.invalidAmount": "Please enter a valid amount.",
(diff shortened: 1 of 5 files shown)


app/i18n/locales/en-US.json
```diff
@@ -117,7 +117,7 @@
   "wallet.restore.dialog.wallet.name.input.label": "Wallet name",
   "wallet.send.form.amount.equalsAda": "equals {amount} ADA",
   "wallet.send.form.amount.hint": "Amount in Lovelaces",
-  "wallet.send.form.amount.label": "Amount in Lovelaces",
+  "wallet.send.form.amount.label": "Amount",
   "wallet.send.form.description.hint": "You can type a message here, if you want.",
```
darko-mijic: Let's use `"wallet.redeem.dialog.adaAmountLabel": "Ada amount"` here since we have it translated to Japanese. Not the same translation key, but same copy: "Ada amount". So we can create Japanese translation without translators.

nikolaglumac: @darko-mijic as agreed during our discussion today, this is not needed as Japanese translation actually contains translation for "Amount" word and not "Amount in Lovelaces": https://github.com/input-output-hk/daedalus/blob/cardano-sl-0.4-rc/app/i18n/locales/ja-JP.json#L120

PR #251, opened 2017-04-19
[DAEF-170] Fix wallet name hardcoded in japanese translation issue

darko-mijic:
Wallet name was hardcoded to "Shopping" in Japanese translation. This PR fixes that.

screen shot 2017-04-19 at 12 51 13


darko-mijic:
This is ready now @nikolaglumac

PR #252, opened 2017-04-19
Fix/daef 165 hamburger button and wallets name are present on the Ada redemption and Settings screens.

tomothespian:
This PR Introduces a fix for displaying hamburger button and wallet name in header only on wallet pages.
wallets page header
ada redemption page header
settings page header


darko-mijic:
Reviewing this now

PR #259, opened 2017-04-20
Feature/daef 167 improve sidebar toggling

tomothespian:
This PR introduces sidebar toggling on window resize and by click on hamburger button.


app/stores/SidebarStore.js
```diff
@@ -12,24 +12,19 @@ export default class SidebarStore extends Store {
...
 
-  INITIAL_HIDE_SUB_MENU_DELAY = 4000;
-  ACTION_HIDE_SUB_MENU_DELAY = 1000;
+  DEFAULT_WINDOW_WIDTH = 1150;
 
   @observable activeSidebarCategory: string = this.CATEGORIES.WALLETS;
-  @observable isShowingSubMenus: boolean = true;
-
-  _hideSubMenuTimeout = null;
+  @observable isShowingSubMenus: bool = false;
```
DominikGuzei: `boolean` is the correct Flow type 😉

app/stores/SidebarStore.js
```diff
@@ -12,24 +12,19 @@ export default class SidebarStore extends Store {
...
     const actions = this.actions.sidebar;
     actions.toggleSubMenus.listen(this._toggleSubMenus);
     actions.activateSidebarCategory.listen(this._onActivateSidebarCategory);
     actions.walletSelected.listen(this._onWalletSelected);
-    this.actions.networkStatus.isSyncedAndReady.listen(() => {
-      this._hideSubMenusAfterDelay(this.INITIAL_HIDE_SUB_MENU_DELAY);
-    });
     this.registerReactions([
       this._syncSidebarRouteWithRouter,
+      this._setInitialSubMenusState,
```
DominikGuzei: You should not register this as reaction since `_setInitialSubMenusState` should only be run once and does not depend on any state from what i see?

DominikGuzei:
Works great and is much, much better than what we had before! 👍 
I just added some comments for small fixes to make!

app/stores/SidebarStore.js
tomothespian: You are absolutely right! Fixed in my latest commit ;)

app/stores/SidebarStore.js
tomothespian: Fixed! ;)

PR #266, opened 2017-04-21
[CSL-1041] Enable --web when running the node (needed for utxo api)

neongreen:
Don't merge, this PR is only needed to build an installer.


jmitchell:
Closing in favor of duplicate #271.

PR #264, opened 2017-04-20
[DEVOPS-10] Windows build automation

deepfire:
- move most contents of the `appveyor.yml` build script into `scripts/windows-build-fresh-daedalus.bat`, adding some sanity checks on the way
- scripts/windows-build-fresh-daedalus.bat, for a shortcuttable easy builds on Windows

PR #269, opened 2017-04-21
[DEVOPS-64] dont sign macos installer for PRs

domenkozar:
If users don't have write access, CI will now fail since it won't be able to get the installer certificate.

Let's just sign for branch builds.

### DIFF
--- a/installers/MacInstaller.hs
+++ b/installers/MacInstaller.hs
@@ -72,13 +72,19 @@ main = do
        ]
   run "productbuild" productargs
 
-  -- Sign the installer with a special macOS dance
-  run "security" ["create-keychain", "-p", "travis", "macos-build.keychain"]
-  run "security" ["default-keychain", "-s", "macos-build.keychain"]
-  shells "security import macos.p12 -P $CERT_PASS -k macos-build.keychain -T `which productsign`" mempty
-  run "security" ["set-key-partition-list", "-S", "apple-tool:,apple:", "-s", "-k", "travis", "macos-build.keychain"]
-  run "security" ["unlock-keychain", "-p", "travis", "macos-build.keychain"]
-  shells ("productsign --sign \"Developer ID Installer: Nikolaos Bentenitis\" --keychain macos-build.keychain dist/temp2.pkg " <> T.pack pkg) mempty
+  isPullRequest <- fromMaybe "false" <$> lookupEnv "TRAVIS_PULL_REQUEST"
+  putStrLn isPullRequest
+  if isPullRequest == "false" then do
+    -- Sign the installer with a special macOS dance
+    run "security" ["create-keychain", "-p", "travis", "macos-build.keychain"]
+    run "security" ["default-keychain", "-s", "macos-build.keychain"]
+    shells "security import macos.p12 -P $CERT_PASS -k macos-build.keychain -T `which productsign`" mempty
+    run "security" ["set-key-partition-list", "-S", "apple-tool:,apple:", "-s", "-k", "travis", "macos-build.keychain"]
+    run "security" ["unlock-keychain", "-p", "travis", "macos-build.keychain"]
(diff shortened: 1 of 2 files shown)


installers/MacInstaller.hs
```diff
@@ -72,13 +72,18 @@ main = do
...
 
-  -- Sign the installer with a special macOS dance
-  run "security" ["create-keychain", "-p", "travis", "macos-build.keychain"]
-  run "security" ["default-keychain", "-s", "macos-build.keychain"]
-  shells "security import macos.p12 -P $CERT_PASS -k macos-build.keychain -T `which productsign`" mempty
-  run "security" ["set-key-partition-list", "-S", "apple-tool:,apple:", "-s", "-k", "travis", "macos-build.keychain"]
-  run "security" ["unlock-keychain", "-p", "travis", "macos-build.keychain"]
-  shells ("productsign --sign \"Developer ID Installer: Nikolaos Bentenitis\" --keychain macos-build.keychain dist/temp2.pkg " <> T.pack pkg) mempty
+  isPullRequest <- fromMaybe "false" <$> lookupEnv "TRAVIS_PULL_REQUEST"
+  if isPullRequest == "false" then do
```
deepfire: I'd love to know the value of the `TRAVIS_PULL_REQUEST` as this test was failiing..

installers/MacInstaller.hs
```diff
@@ -72,13 +72,19 @@ main = do
...
   run "productbuild" productargs
 
-  -- Sign the installer with a special macOS dance
-  run "security" ["create-keychain", "-p", "travis", "macos-build.keychain"]
-  run "security" ["default-keychain", "-s", "macos-build.keychain"]
-  shells "security import macos.p12 -P $CERT_PASS -k macos-build.keychain -T `which productsign`" mempty
-  run "security" ["set-key-partition-list", "-S", "apple-tool:,apple:", "-s", "-k", "travis", "macos-build.keychain"]
-  run "security" ["unlock-keychain", "-p", "travis", "macos-build.keychain"]
-  shells ("productsign --sign \"Developer ID Installer: Nikolaos Bentenitis\" --keychain macos-build.keychain dist/temp2.pkg " <> T.pack pkg) mempty
+  isPullRequest <- fromMaybe "false" <$> lookupEnv "TRAVIS_PULL_REQUEST"
```
jmitchell: We should only sign when `$TRAVIS_PULL_REQUEST` is defined and is `"false"`. If for some strange reason `$TRAVIS_PULL_REQUEST` was ever undefined this code would attempt to sign anyway.

Please change the default case to something other than "false".

installers/MacInstaller.hs
```diff
@@ -72,13 +72,19 @@ main = do
...
+  putStrLn isPullRequest
```
jmitchell: `putStrLn $ "Pull request: " <> isPullRequest` or remove entirely.

installers/MacInstaller.hs
domenkozar: Good point.

deepfire:
https://travis-ci.org/input-output-hk/daedalus/jobs/224451193 -- with https://github.com/input-output-hk/daedalus/pull/264/commits/3a4d48bc5dc93d146181c5d52ce8a20f515ae976 added on top of your commit.

And the result is still odd -- somehow the thing manages to fail.

deepfire:
Ah!

That's probably due to:
`The command "if [[ "$TRAVIS_PULL_REQUEST" == "false" ]]; then nix-shell -p awscli --run "travis_retry aws s3 cp --region eu-central-1 s3://iohk-private/macos.p12 macos.p12"; done" exited with 1.`

jmitchell:
@deepfire, yes, I've pushed a likely fix.

domenkozar:
@jmitchell @deepfire it passes CI now.

PR #270, opened 2017-04-21
Fix/daef 174 split pending confirmation amounts to incoming and outgoing

tomothespian:
This PR introduces pending confirmation amounts split into incoming and outgoing confirmations.
incoming outgoing pending confirmation


nikolaglumac:
Thanks @DominikGuzei - thought you might like that :)

PR #274, opened 2017-04-24
Feature/daef 168 optionally setting a password during wallet creation

nikolaglumac:
This PR introduces optionally setting a password during wallet creation.

wallet-password-1
wallet-password-2

## TODO:
- [x] Run translation manager


app/actions/wallets-actions.js
```diff
@@ -4,7 +4,7 @@ import Action from './lib/Action';
 // ======= WALLET ACTIONS =======
 
 export default class WalletsActions {
-  createWallet: Action<{ name: string, currency: string }> = new Action();
+  createWallet: Action<{ name: string, currency: string, password: string }> = new Action();
```
DominikGuzei: I guess you should make the password field optional like this `password?: string` (means the property can be missing) or like this: `password: ?string` (means it can either be a string or null)

nikolaglumac: I was thinking about that but since we default the password to empty string here: https://github.com/input-output-hk/daedalus/blob/feature/daef-168-optionally-setting-a-password-during-wallet-creation/app/components/wallet/WalletCreateDialog.js#L130 it should never happen that it has null value but if you want I can change it to `?string`

DominikGuzei: Yeah but that's only an implementation detail inside the form … I think you should not submit an empty string password at all (we should only allow valid data to be sent). So i would prefer it to be `null` in case of no password is set

nikolaglumac:
@DominikGuzei I've updated it as you requested. Now the logic is held within the form and if no password is set `null` value is submitted.

app/api/index.js
```diff
@@ -22,6 +22,7 @@ export type CreateWalletRequest = {
   name: string,
   currency: string,
   mnemonic: string,
+  password: string,
```
DominikGuzei: This must also be `?string`

nikolaglumac: Good catch! Fixed ;)

DominikGuzei:
@nikolaglumac what about the open todos you mentioned in the description?

nikolaglumac:
@DominikGuzei todos are now checked off - we are ready for merging :)

DominikGuzei:
@nikolaglumac just merge as soon as CI allows it, im off for lunch

PR #275, opened 2017-04-24
Fix/daef 176 redemption key is reset after unsuccessful redemption

tomothespian:
This PR introduces a fix which prevents redemption key reset after unsuccessful redemption.


app/stores/AdaRedemptionStore.js
```diff
@@ -30,6 +30,7 @@ export default class AdaRedemptionStore extends Store {
   @observable amountRedeemed: number = 0;
   @observable showAdaRedemptionSuccessMessage: boolean = false;
   @observable redeemAdaRequest: Request<Wallet> = new Request(this.api.redeemAda);
+  @observable redeemAdaSubmitted: boolean = false;
```
DominikGuzei: I would rename this to `adaSuccessfullyRedeemed` because it's not only about the "submit" state (which you could ask for via `redeemAdaRequest.wasExecuted` but about success state.

nikolaglumac:
@DominikGuzei we do still reset those values after successful submit. The scope of this task was just to keep the values in case error pops out. But now that you have mentioned this I agree we should probably re-think whole ada-redemption workflow...

app/stores/AdaRedemptionStore.js
nikolaglumac: @DominikGuzei I have refactored the code so that this variable is not needed at all. Please review again ;)

PR #277, opened 2017-04-24
[DO NOT MERGE] Csl 1071/addr conversion

gromakovsky:
It's only needed to produce an installer.

### DIFF
--- a/installers/Launcher.hs
+++ b/installers/Launcher.hs
@@ -35,6 +35,7 @@ launcherArgs launcher = unwords $
         "--peers-file", "ip-dht-mappings",
         "--system-start", "1492428658",
         "--wallet",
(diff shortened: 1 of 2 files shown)


domenkozar:
How is cardano-addr-convert invoked?

gromakovsky:
`cardano-addr-convert VENDING_ADDRESS` or just `cardano-addr-convert` and then it will be interactive mode.

domenkozar:
oh so through CLI?

gromakovsky:
Yes, it's a simple command line utility and the whole point is that we don't want Jeremy to build it locally. Well, now I am thinking about it and probably it is not proper approach, I could just build it locally and send it to him… 🤦‍♂️

PR #278, opened 2017-04-25
[DO-NOT-MERGE][DEVOPS-33] Daedalus/win64: TLS keys/certs made available for FE<->BE channel securing

deepfire:
This does 1/2 of the work for DEVOPS-33 -- the windows part of TLS CA/server key/cert generation.

Since this is heavily reliant on the Daedalus-on-Windows part of DEVOPS-10, this is currently on-top of the as-yet unmerged PR https://github.com/input-output-hk/daedalus/pull/264.

However, most of the non-DEVOPS-10 parts (basically, everything under `installers/`) is meaningfully available for review,
and so -- cc @akegalj, @domenkozar, @jmitchell.

For details, please see: https://issues.serokell.io/issue/DEVOPS-33#comment=96-3065

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,10 @@ main.js.map
 
 # Editors
 .idea
+*~
+\#*
+.\#*
+.~*
 
 # Generated translation files
 translations/messages
--- a/README.md
+++ b/README.md
@@ -1,6 +1,18 @@
 # daedalus
 Daedalus - cryptocurrency wallet
 
+## Quick all-in-one installer build script (Windows-only, for the time)
+
+0. Dependencies: `Node.js`, `7zip`, `git`
+1. Obtain https://github.com/input-output-hk/daedalus/blob/master/scripts/windows-build-fresh-daedalus.bat
+2. Run it:
+   ```cmd
+   C:\windows-build-fresh-daedalus.bat [BRANCH] [GITHUB-USER]
+   ```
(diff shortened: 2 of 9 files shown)


deepfire:
One known deficiency, albeit unexpected, is that invocation of the `installers/build-certificates-win64.bat` script by the means of the `Development.NSIS.exec` directive opens a `conhost.exe` window, whereas it is supposed not to -- see https://github.com/input-output-hk/daedalus/pull/278/files#diff-936ebc51c2ddde384bd5cf4fde429808R142.

This isn't something we can work around of, without involving the upstream, I would guess.

deepfire:
The implementation of `out` seems to suggest that `Development.NSIS.exec` is implemented in terms of `ExecShell` and not `Exec`: https://hackage.haskell.org/package/nsis-0.3.1/docs/src/Development-NSIS-Show.html#out

jmitchell:
@deepfire, I haven't looked into the details yet, but if what you want to do is possible in the NSIS scripting language there's likely a way to do it from the Haskell NSIS DSL using the `inject*` functions.

deepfire:
Due to CI-specific issues this was reissued as https://github.com/input-output-hk/daedalus/pull/286

deepfire:
This has been superceded by https://github.com/input-output-hk/daedalus/pull/293.

Thanks to @akegalj for spotting this : -)

PR #289, opened 2017-04-28
Fix/daef 118 refactor all dialog / notification (isOpen) flags to dedicated, generic manager

tomothespian:
This PR introduces a fix which presents refactored dialog/notification flags as dedicated, generic managers.


DominikGuzei:
reviewing this now …

PR #290, opened 2017-05-01
[CSL-1054] Enable --web when running the node (needed for utxo api)

gromakovsky:
`--web` flag enables some web endpoints which are mostly needed for debugging. It shouldn't add overhead or vulnerabilities.


georgeee:
Never considered it a good idea. Closing.
This was need probably only for operation of utxo dump, which is rare and could be performed by explicitly modifying CLI flags (e.g. @fersel knows how to do it)

PR #295, opened 2017-05-03
Feature/daef 198 change password dialog ux improvements

nikolaglumac:
This PR introduces change wallet password dialog UX improvements in terms of clear distinction between wallet create password, update password and remove password actions.

result


darko-mijic:
Reviewing this now...

PR #303, opened 2017-05-12
Fix/daef 225 hide pending amount if it is zero

tomothespian:
This PR introduces fix which hide pending amount if it is zero

### DIFF
--- a/app/components/wallet/summary/WalletSummary.js
+++ b/app/components/wallet/summary/WalletSummary.js
@@ -59,14 +59,14 @@ export default class WalletSummary extends Component {
             {amount}
             <img src={adaSymbolBig} role="presentation" />
           </div>
-          {pendingAmount.incoming.toFormat() > 0 &&
+          {pendingAmount.incoming.greaterThan(0) &&
             <div className={styles.pendingConfirmation}>
               {`${intl.formatMessage(messages.pendingIncomingConfirmationLabel)}`}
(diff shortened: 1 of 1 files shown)


app/components/wallet/summary/WalletSummary.js
```diff
@@ -59,16 +59,20 @@ export default class WalletSummary extends Component {
...
-            {`${intl.formatMessage(messages.pendingIncomingConfirmationLabel)}`}
-            : {pendingAmount.incoming.toFormat(DECIMAL_PLACES_IN_ADA)}
-            <img src={adaSymbolSmallest} role="presentation" />
-          </div>
-          <div className={styles.pendingConfirmation}>
-            {`${intl.formatMessage(messages.pendingOutgoingConfirmationLabel)}`}
-            : {pendingAmount.outgoing.toFormat(DECIMAL_PLACES_IN_ADA)}
-            <img src={adaSymbolSmallest} role="presentation" />
-          </div>
+          {pendingAmount.incoming.toFormat() > 0 &&
```
DominikGuzei: That won't work correctly (or is not wise) because [toFormat](http://mikemcl.github.io/bignumber.js/#toFor) returns a string. I will update it to `greaterThan(0)` which is the appropriate method in this case 😉

tomothespian: Ahaa, you're right. Good catch. Thx for fix ;)

PR #306, opened 2017-05-16
Chore/daef 224 use markdown for terms of use

tomothespian:
This PR introduces markdown for terms of use.


app/stores/AppStore.js
```diff
@@ -50,6 +50,11 @@ export default class AppStore extends Store {
     return 'en-US'; // default
   }
 
+  @computed get termsOfUse(): string {
+    const localizedTermsOfUse = require(`../i18n/locales/terms-of-use/${this.currentLocale}.md`); // eslint-disable-line
```
DominikGuzei: What was the resolution for this problem? Is this (still) a problem or does it work on all OS?

darko-mijic: This is OK. My suggestion was false.

PR #313, opened 2017-05-18
feature/daef 231 rename wallet

nikolaglumac:
This PR introduces wallet-rename feature on "Wallet Settings" screen and inline-editing-input field component.

wallet-rename


nikolaglumac:
This PR is now ready to be reviewed. Since it requires `old` (non-HD-wallet) backend and I know @darko-mijic is running the latest one I suggest @DominikGuzei does the review :)
Thanks!

PR #314, opened 2017-05-18
[DEVOPS-118] Windows installer: escalate to admin exec level

jmitchell:
Installer currently fails to run for non-admin users without explicitly running it as an admin. To avoid the error we should preemptively request admin permission.

Manually verified using a local build on Windows Server 2012 r2.

### DIFF
--- a/installers/WindowsInstaller.hs
+++ b/installers/WindowsInstaller.hs
@@ -107,7 +107,7 @@ writeInstallerNSIS fullVersion = do
     injectGlobalLiteral $ "VIAddVersionKey \"ProductVersion\" " <> fullVersion
     injectGlobalLiteral "Unicode true"
     installDir "$PROGRAMFILES64\\Daedalus"   -- The default installation directory
-    requestExecutionLevel Highest
+    requestExecutionLevel Admin
     injectGlobalLiteral "!addplugindir \"nsis_plugins\\liteFirewall\\bin\""
 
     page Directory                   -- Pick where to install


domenkozar:
@fersel can you test this installer with the update system?

installers/WindowsInstaller.hs
```diff
@@ -107,7 +107,7 @@ writeInstallerNSIS fullVersion = do
     injectGlobalLiteral $ "VIAddVersionKey \"ProductVersion\" " <> fullVersion
     injectGlobalLiteral "Unicode true"
     installDir "$PROGRAMFILES64\\Daedalus"   -- The default installation directory
-    requestExecutionLevel Highest
+    requestExecutionLevel Admin
```
deepfire: Context for this change:

- http://stackoverflow.com/a/12652091

deepfire:
Quoting http://stackoverflow.com/a/12652091:

> `highestAvailable` will elevate if the current user is an administrator. Which is consistent with what you have observed. When a standard user runs the process, no UAC dialog is shown and the process runs with the standard token. When an admin user executes, the UAC consent dialog is shown and the process will then run elevated.
>
> If your program requires admin rights to function then you need to use `requireAdministrator`. When a standard user starts such a process, the over-the-shoulder UAC dialog is shown. That gives the user an opportunity to ask an admin to supply their credentials.

This makes sense.

installers/WindowsInstaller.hs
jmitchell: Additional context from [NSIS docs](http://nsis.sourceforge.net/Docs/Chapter4.html#arequestexecutionlevel)

PR #316, opened 2017-05-24
[DEVOPS-102] Run cardano-launcher at highestAvailable level

jmitchell:
Necessary for getting an update Installer.exe to elevate to admin
execution level when run by an admin user.

DEVOPS-118 tried to address this by setting requestExecutionLevel
to admin, but it doesn't take take effect when executed by a
non-elevated exe (in this case cardano-launcher.exe).

This approach has been verified on a Windows 10 Pro VM.

### DIFF
--- a/installers/WindowsInstaller.hs
+++ b/installers/WindowsInstaller.hs
@@ -170,7 +170,7 @@ main = do
   signUninstaller
 
   echo "Adding permissions manifest to cardano-launcher.exe"
-  procs "mt.exe" ["-manifest", "cardano-launcher.exe.manifest", "-outputresource:cardano-launcher.exe;#1"] mempty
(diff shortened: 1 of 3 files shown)


installers/WindowsInstaller.hs
```diff
@@ -116,8 +116,6 @@ writeInstallerNSIS fullVersion = do
     _ <- section "" [Required] $ do
         setOutPath "$INSTDIR"        -- Where to install files in this section
         writeRegStr HKLM "Software/Daedalus" "Install_Dir" "$INSTDIR" -- Used by launcher batch script
-        createDirectory "$APPDATA\\Daedalus\\DB-0.2"
```
domenkozar: Why were these removed?

jmitchell: It's a minor unrelated bug Ilya and I noticed where we're unnecessarily creating directories for old versions. I addressed it in its [own commit](https://github.com/input-output-hk/daedalus/pull/316/commits/372adc9a044f9c90be8a453e9648432f59b93c56).

PR #322, opened 2017-05-29
[DEVOPS-102] Merge fix into master

jmitchell:
Changes are already in `cardano-sl-0.4`. See PR #316.


deepfire:
Summary of changes:

- drop install-time creation of stale directories
- supply a manifest file that asks WIndows to elevate permissions for the installer executable
- use `mt.exe` to attach the manifest to executable

Makes total sense.

PR #323, opened 2017-05-30
Fix builds on branches that have `/` in their name

jmitchell:
Fixes the URL for downloading daedalus bridge assets from S3 when
branch name has a / in it. Already tested in branch
nh2/csl-1004-node-catchup-improvements.

### DIFF
--- a/scripts/build-installer-unix.sh
+++ b/scripts/build-installer-unix.sh
@@ -59,7 +59,7 @@ upload_s3=
 test_install=
 
 daedalus_version="$1"; argnz "product version"; shift
-cardano_branch="$1";   argnz "Cardano SL branch to build Daedalus with"; shift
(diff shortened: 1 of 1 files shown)


scripts/build-installer-unix.sh
```diff
@@ -59,7 +59,7 @@ upload_s3=
 test_install=
 
 daedalus_version="$1"; argnz "product version"; shift
-cardano_branch="$1";   argnz "Cardano SL branch to build Daedalus with"; shift
+cardano_branch="$(printf '%s' "$1" | tr '/' '-')"; argnz "Cardano SL branch to build Daedalus with"; shift
```
deepfire: By the way, why not `echo "$1"` instead of `printf '%s' "$1"`?

jmitchell: `echo` adds a trailing `\n` by default. `echo -n "$1"` probably would have been fine too.

deepfire: Oh, right : -)

PR #325, opened 2017-06-02
Feature/daef 252 exporting paper wallets

tomothespian:
This PR introduces paper wallets export.

## TODO
- [x] Remove link to paper wallet export dialog from wallet settings as the logic part is not yet implemented

export paper wallets
check printer copy
mnemonic
verify mnemonic
paper certificate


DominikGuzei:
Reviewing this now …

DominikGuzei:
✅ This PR can be merged as soon as CI is green!

DominikGuzei:
@nikolaglumac and @tomothespian are the ones who built that feature, so let's ask them 😉

adinapoli-iohk:
Of course! Fire away, guys! 😀

nikolaglumac:
@adinapoli-iohk I don't think we have some hard demands in our minds.
Perhaps the question is if it is better to have one single endpoint which will cover all 3 possible cases shown in the UI (full-export, read-only-export and paper-wallet-export) where these 3 would be in form of a call param sent to this endpoint OR if it is better to have 3 different endpoints. But the problem is that we currently have only UI for the paper-wallet-export.
What do you think @adinapoli-iohk?

adinapoli-iohk:
> Perhaps the question is if it is better to have one single endpoint which will cover all 3 possible cases shown in the UI (full-export, read-only-export and paper-wallet-export) where these 3 would be in form of a call param sent to this endpoint OR if it is better to have 3 different endpoints. 

Uhm, good question! I guess that in order to answer this I would need to understand the overall feature more broadly. For example, if we expect those 3 endpoints to be largely similar and to produce the same output (from a RESTful response point of view) then yes, I think it makes sense to bundle them up where the `exportType` is just another parameter of the request.

Conversely, if we think the outputs will be different, it looks like we might need three different endpoints.

Generally speaking, I would say that for now we can start with a single endpoint covering _only_ the paper wallet case (example: `/api/wallets/{walletId}/paper/export` or similar) and for the upcoming revision of the middleware we come back to our design decisions. Would that cut it?

nikolaglumac:
@adinapoli-iohk since I am not sure if or how different the outputs will be (we need an input from @darko-mijic for this) I agree that we proceed with single paper-wallet export endpoint.
I would only suggest that `export` and `paper` params switch places so that it reads `/api/wallets/{walletId}/export/paper`.

PR #327, opened 2017-06-08
Fix/daef 232 issues with ada amount input

DominikGuzei:
This PR finishes the integration of the new `NumericInput` component in react-polymorph.
On the way we discovered a bunch of issues which resulted in the following included improvements:

- Simplified wallets request handling and improved performance by tracing down unnecessary mobx re-calculations and dependencies
- Reduced re-renderings caused by network difficulty updates that have been de-referenced on layout level and thus caused everything to rerender unnecessarily (and not just the sync state widget).

## Tasks left todo:

- [x] fix bug in `NumericInput` component: When you enter `0.000010` char-by-char then you end up with `0.000100` instead (the first zero after the dot is swallowed)
- [x] check if acceptance tests are running through after the fix (and fix if necessary)
- [x] review and merge the [react-polymorph bug fix](https://github.com/CodeAdventure/react-polymorph/pull/7)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -51,7 +51,9 @@
     "spaced-comment": 1,
     "quotes": 1,
     "import/imports-first": 1,
-    "no-multi-spaces": 1
+    "no-multi-spaces": 1,
+    "no-multiple-empty-lines": 1,
+    "react/jsx-indent": 1
   },
   "plugins": [
     "flowtype-errors",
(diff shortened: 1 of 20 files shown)


DominikGuzei:
@nikolaglumac this is ready for review … please not that it only works with the fixes from the `react-polymorph` PR which is also ready for review: https://github.com/CodeAdventure/react-polymorph/pull/7

PR #329, opened 2017-06-13
Feature/daef 261 spending password on import wallet from key dialog

tomothespian:
This PR introduces a feature for spending password on import wallet from key import
screen shot 2017-06-13 at 10 43 58


nikolaglumac:
@tomothespian @darko-mijic @DominikGuzei it is decided that this PR should not be merged at the moment due to issues with backend logic where within importKey endpoint spending password is not meant to be created for newly imported wallet but for authorization of wallet key which has built-in password.

PR #346, opened 2017-06-27
[DAEF-325] Disable polling for wallet data while syncing

darko-mijic:
This PR disables polling for wallet data while the wallet node is syncing. Polling while syncing is slowing dow the node.

To do:
- [x] Verify that tests are green
- [x] Crate a card for changing the logic for `isSynced`, with 15-second slots sync is jumping from 99% to 100% constantly since new blocks are received very often.

### DIFF
--- a/app/stores/WalletsStore.js
+++ b/app/stores/WalletsStore.js
@@ -61,7 +61,7 @@ export default class WalletsStore extends Store {
       this._toggleAddWalletDialogOnWalletsLoaded,
     ]);
     if (environment.CARDANO_API) {
-      setInterval(this.refreshWalletsData, this.WALLET_REFRESH_INTERVAL);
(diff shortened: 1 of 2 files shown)


app/stores/WalletsStore.js
```diff
@@ -180,7 +180,7 @@ export default class WalletsStore extends Store {
   isValidPrivateKey = () => { return true; }; // eslint-disable-line
 
   @action refreshWalletsData = async () => {
-    if (this.stores.networkStatus.isConnected) {
+    if (this.stores.networkStatus.isConnected && this.stores.networkStatus.isSynced) {
```
darko-mijic: I guess we only need `isSynced` here. Or maybe we should check the sync percentage to be more than 99%? With 15-second slots syncing is very often less than 100% so the syncing spinner is spinning.

PR #349, opened 2017-06-28
Fix/daef 324 acceptance test for double money received bug

nikolaglumac:
This PR extends acceptance test for "Sending money" feature by checking receiver's wallet balance after transaction is complete and by doing so exposes current API double-amount bug.


nikolaglumac:
@darko-mijic @DominikGuzei @tomothespian until API is fixed we should see two failing acceptance tests like shown in the following screenshot:
screen shot 2017-06-28 at 09 33 35

DominikGuzei:
Reviewing this now …

PR #352, opened 2017-06-28
Feature/daef 321 Receive page update

nikolaglumac:
This PR introduces updated "Receive page" design.

receive-page


darko-mijic:
This is exactly what we specified. Good work. But we should try to do something about the layout of the top part. It looks bad with the screen size from the screenshot.

darko-mijic:
"Your wallet address" is wrong since this is the address (one of the addresses) of the account. Maybe that can be moved as a title above everything else on the page. CC @a-rukin

nikolaglumac:
@darko-mijic top part is now updated according to our discussion. Here's how it looks:

screen shot 2017-06-29 at 09 48 50
screen shot 2017-06-29 at 09 48 47
screen shot 2017-06-29 at 09 44 00

PR #357, opened 2017-06-29
Fix/daef 329 improve tests for generating new addresses

tomothespian:
This PR introduces acceptance tests fix for generating new addresses


nikolaglumac:
@darko-mijic I have fixed the problem with refresh of lastGeneratedAddress and now the tests are passing in 90%. In other 10% when it fails it is to the fact API returns wallet's addresses in alphabetical order instead of ordering them by creation/generation date...

nikolaglumac:
NOTE: Acceptance tests will pass 100% after this issue is fixed: https://issues.serokell.io/issue/CSM-340

PR #360, opened 2017-07-04
Feature/daef 306 Opt-in mode for sending logs to the remote server for the mainnet version of Daedalus

tomothespian:
This PR introduces opt-in mode for sending logs to remote server UI, frontend logic and acceptance tests. UI includes opt-in screen during application start as well as support settings screen.

## TODOs
- [x] Implement UI components and logic to save user's send-logs preference
- [x] Use saved user's send-logs preference in actual logging process
- [x] Inject missing Japanese translations

send logs on support settings screen
send logs form


nikolaglumac:
Great work @tomothespian! I have only added few minor flow-type and styling fixes.
UI part works as expected (both application-start opt-in and settings screen one).

@darko-mijic @DominikGuzei please also review Tomislav's work as this is important part of the application. In the meantime I will add "WIP" tag on this PR as we still need to implement the backend part (connect user's opt-in setting with actual logging process) in the scope of this task.

DominikGuzei:
Reviewing this now …

app/stores/AppStore.js
```diff
@@ -96,21 +109,23 @@ export default class AppStore extends Store {
     this.getTermsOfUseAcceptanceRequest.execute();
   };
 
+  _getSendLogsChoice = async () => {
+    const sendLogs = await this.getSendLogsChoiceRequest.execute().promise;
+    ipcRenderer.send('send-logs-choice', sendLogs);
```
DominikGuzei: Hmm … why are you `send`ing something to the main process in a getter function? This is a bit confusing for me.

app/stores/AppStore.js
```diff
@@ -96,21 +109,23 @@ export default class AppStore extends Store {
...
+  };
+
+  _setSendLogsChoice = async ({ sendLogs }: { sendLogs: boolean }) => {
+    await this.setSendLogsChoiceRequest.execute(sendLogs);
+    await this._getSendLogsChoice();
```
DominikGuzei: And here you use the getter to send the change to the main process? 😦

features/send-money-to-receiver.feature
```diff
@@ -3,6 +3,7 @@ Feature: Send Money to Receiver
   Background:
     Given I have selected English language
     And I have accepted "Terms of use"
+    And I agree to send logs to remote server
```
DominikGuzei: I guess the duplication of the background stories here is a good indicator that we need to refactor this into a unified step that does all necessary background work to get to the meat of other tests.

Added a card for this: https://issues.serokell.io/issue/DAEF-365

nikolaglumac: We should definitely unify these 3 steps into a single one 👍

app/stores/AppStore.js
nikolaglumac: It is because _getSendLogsChoice is called during AppStore setup and a moment when we actually read current user's send-logs-choice value that we need to send to the main process. I'm open to suggestions :)

app/stores/AppStore.js
nikolaglumac: Ah I was trying to optimize this part. Have reverted it now to what I initially had in mind ;)

app/stores/AppStore.js
DominikGuzei: So then i would move this into its own method called `sendLogsChoiceToMainProcess` and call it separately in the `setup` method after `_getSendLogsChoice`. One very important rule for me is that getters don't have side-effects, as this leads to big confusion and bad design.

DominikGuzei:
@nikolaglumac this looks great now! Will merge as soon as CI is green 👍

DominikGuzei:
@nikolaglumac @darko-mijic the only thing left to clarify is if this can be merged in to master without causing issues!

nikolaglumac:
@darko-mijic @DominikGuzei @tomothespian it has been decided that we will merge this branch directly into `release/0.7.0-rc` (https://github.com/input-output-hk/daedalus/pull/390) and thus this PR will be closed without merging.

PR #365, opened 2017-07-05
Fix/daef 352 Modify acceptance tests to work with production environment setup

nikolaglumac:
This PR replaces default-wallet.key used in acceptance tests with one that works in production environment and increases acceptance tests before steps timeout.


nikolaglumac:
@darko-mijic @DominikGuzei I had to increase both `BeforeFeatures` and `Before` timeouts in order for the acceptance tests to work in production mode. Please review my updates.

nikolaglumac:
All tests should pass except for `User Sends Money from wallet with spending password to Receiver` where following error is thrown:

screen shot 2017-07-05 at 16 24 09

Please note that when I run this test isolated it passes normally...

nikolaglumac:
Also I have recorded few occasions where "Generating wallet address" acceptance test fails BUT I know for sure this is a false-positive due to race conditions between updating wallet's lastGeneratedAddress and all wallet addresses.

DominikGuzei:
testing this now …

DominikGuzei:
@nikolaglumac but will these changes solve the problem of `function timed out after 5000 milliseconds`? it seems to me that these settings are not affecting that. I will investigate and increase these limits too

nikolaglumac:
@DominikGuzei I doubt that increasing individual step's timeout would help. For me only send-money scenarios fail and even though the failure is recorded as "function timed out after 5000 milliseconds" it times-out due to "Not enough money to send" message.

DominikGuzei:
@nikolaglumac still, we should keep the cucumber timeout always higher than the webdriver timeouts … otherwise this make no sense 😉 the cucumber timeout should be the last resort and not a pre-emptive cut-off of the tests. Cucumber also produces non-informative errors "that step just timed out" and webdriver would tell you "this element was not visible after 5 seconds"

DominikGuzei:
I can confirm that error messages are much better now:

`Error: element (".Transaction_title") still not visible after 10000ms`

instead of:

`function timed out after 10000 milliseconds`

DominikGuzei:
@nikolaglumac will the production keys actually work in the dev environment?

nikolaglumac:
@DominikGuzei I doubt they will work in dev mode - we have replaced them in order to be able to run the tests in production/staging mode.
If you want to run the tests in dev mode you will need to generate new key.

DominikGuzei:
@nikolaglumac ok i guess at the moment it's not super important as we don't really need dev mode … it's much faster though, so at some point it could be that we need it again.

nikolaglumac:
@DominikGuzei agreed! Perhaps we could have both keys and some custom flag to define which one we want to use for certain test-run ;)

PR #370, opened 2017-07-06
Fix/daef 354 Show more specific error messages on Change password dialog

tomothespian:
This PR introduces a fix which presents more specific error message on Change password dialog.
change password dialog - more specific error message


nikolaglumac:
@darko-mijic can you please take care of the merging? Not sure if it is enough to just merge this fix into release/0.7.0-rc... Thanks!

PR #388, opened 2017-07-10
Fix/daef 363 Implement moment.js internationalization

nikolaglumac:
This PR introduces moment.js internationalization which enables presentation of dates in user's locale.


nikolaglumac:
@DominikGuzei hope it is better now :)
https://github.com/input-output-hk/daedalus/pull/388/commits/fe1da496abab5cad38ed8e72be87af46a2d1b507

DominikGuzei:
@nikolaglumac how did you test the groupings of the transactions? Without mocks i don't have any easy way to simulate transactions from different dates …

nikolaglumac:
@DominikGuzei I was testing it by sending multiple transactions:

screen shot 2017-07-12 at 11 39 18
screen shot 2017-07-12 at 11 39 13

DominikGuzei:
Ah ok … but that's not possible if you have a clean slate, we currently have no way (that i know of) to "add" transactions that are in the past, do we?

nikolaglumac:
You are wright. The only way I can think of is to fake `Transaction` date here: 
https://github.com/input-output-hk/daedalus/blob/feature/daef-306-opt-in-mode-for-sending-logs-to-the-remote-server-for-the-mainnet-version-of-Daedalus/app/api/CardanoClientApi.js#L666
...but then again this would require you to already have at least 2 existing transactions...

DominikGuzei:
But i will add a storbook story with fake data to display a transaction list now … this is the fastest/easiest way to make this testable manually

PR #393, opened 2017-07-20
Fix/daef 381 Detect and fix all brittle acceptance tests

nikolaglumac:
This PR fixes brittle acceptance test steps (wallet latest transaction amount check step).


nikolaglumac:
@DominikGuzei I've found only one acceptance test step definition that I felt it needs fixing.

PR #397, opened 2017-07-21
Fix/daef 383 Fix all design implementation issues

nikolaglumac:
This PR fixes all design implementation issues:

### Wallet Summary Screen
- [x] total amount and ada symbol different font-width, seems that amount should be Ultralight
- [x] lets finally update icons for transactions ;)
- [x] transaction should expand on "whole line" click, not only title
- [x] in expanded view of transaction bottom padding is too small (need to be 20px)

### Wallet Send Screen
- [x] fields are active
- [x] border radiuses
- [x] amount field: type -> clear (cmd+a and backspace) -> type new number -> cursor jumps 1 character before sometimes
- [x] amount field: type -> select all (cmd+a) -> type again -> first character becomes last (may be same as cursor jump)
- [ ] ~~amount field: is it possible to have unused part faded? (see example https://zpl.io/Zfi1EH)~~ `[REJECTED]`

### Wallet Receive Screen
- [x] generated addresses: "copy address" should be aligned 1 column (right in this case)

### Wallet Settings Screen
- [x] "Transaction assurance security level" dropdown is always in active state
- [ ] "your changes saved" should be written near field it connected to, not the top one `[SEPARATE TASK: https://issues.serokell.io/issue/DAEF-333]`
- [x] "password" field, "create" doesn't have hover state (should become more visible on hover)
- [x] input fields doesn't seem to have 2px border radius
- [ ] ~~password dialog: note should go before~~ `[REJECTED]`
- [x] password dialog: error message moves layout a bit
- [x] password dialog: error message doesn't appear if initial password field was changed (repeat password already filled correctly)

### ADA Redemption Screen
- [x] same thing with active dropdown
- [x] border radiuses
- [x] dropdown arrow looks wrong way (each tab)
- [x] distance between redemption key field and certificate "square" is wrong (need to be 30px)

### Settings General Screen
- [x] active dropdown

### Settings Terms-of-use Screen
- [ ] ~~caps text looks bad... may be bold instead?~~ `[REJECTED]`

### Wallet Create Recovery-phrase-dialog
- [ ] ~~can't remove word from field (second click removes word from string, even not last)~~ `[REJECTED]`

### Dialogs
- [x] big shadow, need to be like here: https://zpl.io/1sipYh

### General
- [x] I personally missing "cmd + H" hotkey `[SEPARATE TASK: https://issues.serokell.io/issue/DAEF-385]` PR: https://github.com/input-output-hk/daedalus/pull/404


nikolaglumac:
@a-rukin most of the design-implementation issues have been fixed. Some are to be discussed - I suggest we do it on Monday via Hangout.

@darko-mijic @DominikGuzei cc

nikolaglumac:
@a-rukin I would like to replace the transaction icons but I can't match the existing ones with the ones in the designs. Currently we have 4 different types of icons we use and I can find only 2 of those in the designs. Can you please take a look which ones are missing and provide those? Thanks!
[Icons.zip](https://github.com/input-output-hk/daedalus/files/1169385/Icons.zip)

@darko-mijic cc

nikolaglumac:
@a-rukin please disregard my last comment. We only need the two of the icons that you have provided in the design. Sorry for the false alarm :)

a-rukin:
@nikolaglumac okay ))

nikolaglumac:
@a-rukin most of the issues are now fixed.
After discussing open tasks with the frontend team some were rejected and for two of them we have created separate YouTrack tasks since they are more complex.

@DominikGuzei @darko-mijic please review the fixes so we can merge this PR as soon as possible ;)

PS. here's how new transactions icons look like now @a-rukin:
screen shot 2017-07-24 at 13 11 00
screen shot 2017-07-24 at 13 11 05

PR #400, opened 2017-07-24
Feature/daef 372 Setup environment variable for testnet/mainnet mode

nikolaglumac:
This PR introduces new environment variable `NETWORK` which is to be used to distinguish testnet/mainnet/staging/development network modes.
Several features behave differently based on the value of this variable:
- In Mainnet mode we hide TopBar red Testnet label
- In Mainnet mode we don't show "Ada redemption disclaimer" screen
- Different terms of use are displayed for Mainnet and all other modes

To use this new variable start Daedalus with the following command:
```
$ NETWORK=mainnet npm run dev
```
Available options for `NETWORK` are `mainnet`, `testnet`, `staging` and `development` while default value is `development`.


app/environment.js
```diff
@@ -3,12 +3,14 @@ const environment = Object.assign({
   DEVELOPMENT: 'development',
   TEST: 'test',
   PRODUCTION: 'production',
+  NETWORK: process.env.NETWORK || 'development', // options: mainnet | testnet | staging | development
```
DominikGuzei: Defaults for environment params are all set in the [base webpack config](https://github.com/input-output-hk/daedalus/blob/master/webpack/webpack.config.base.js#L44) atm. I don't fully remember the reason but let's stick to that for consistency.

darko-mijic:
We should also update readme in the scope of this PR @nikolaglumac.

app/environment.js
nikolaglumac: Consider it done!

nikolaglumac: @DominikGuzei done!

domenkozar:
`staging` shouldn't exist as it should be the same as mainnet/testnet. Also staging can be either of the two.

nikolaglumac:
@domenkozar we currently don't have anything special for `staging`. Actually all three of the following: `staging`, 'testnet' and `development` behave the same. Only `mainnet` results in certain changes such as hiding the Testnet label, not showing Ada disclaimer and presenting a different Terms of use. We will probably remove the `staging` option in the future...

domenkozar:
I think that would be wise. Our part of the work is tracked in DEVOPS-226.

nikolaglumac:
@domenkozar I have removed mentioned `staging` option from the README file: https://github.com/input-output-hk/daedalus/pull/410

@darko-mijic cc

PR #402, opened 2017-07-24
Chore/daef 266 improve webpack build performance

DominikGuzei:
This PR introduces two webpack optimizations to improve build performance:

1. [HappyPack](https://github.com/amireh/happypack) is used to compile .js files with multiple threads in parallel
2. Following this [article on webpack performance](http://engineering.invisionapp.com/post/optimizing-webpack/) i introduced the `DLLPlugin` which pre-compiles most of our vendor dependencies into a separate bundle, that is used as reference during compilation. This DLL bundle only has to be re-compiled when any dependency listed changes version (which is much rarer than our code changes 😉 )

These two improvements result in build times ~15 seconds on my machine.

**Todos:**
- [x] Increase performance of es-lint or remove/replace it as it drags down compilation time significantly (+30 seconds)
- [x] Extensively test that everything works correctly with the DLL setup.
- [x] Fix translation files loading error


nikolaglumac:
@DominikGuzei reviewing and testing this one now.

PR #405, opened 2017-07-25
[DEVOPS-223] AppVeyor: stack 1.5.0 upgrade

jmitchell:
In stack v1.4.0 the stack executable was in the root of the downloaded
zip, but now in stack v1.5.0 it's under the stack.exe directory. The
link we're using now resolves to stack v1.5.0.


domenkozar:
Pushed to cardano-sl-0.5-staging

jmitchell:
Hope it works!

jmitchell:
This latest iteration works. However, the change in `stack.exe` location inside the zip was unintended and will eventually be changed back: https://twitter.com/snoyberg/status/889896922546397184

Given we don't have a timeline for that, my suggestion in the short-term is to apply this change once CI passes and clean it up when the upstream fix is live.

PR #428, opened 2017-08-22
Fix/daef 408 ENOENT error

nikolaglumac:
This PR fixes ENOENT error caused by log file rotation by replacing log-file tailing logic with direct logging.


darko-mijic:
Reviewing this now.

nikolaglumac:
Great! :)

PR #430, opened 2017-08-22
Feature/daef 418 About page, available from the system menu

tomothespian:
This PR introduces about page window from system menu.
about window


nikolaglumac:
@DominikGuzei: @tomothespian and I have tried to find a way to speed it up and failed :(
The only possible way to speed it up would be to pre-generate about page in pure html. That way it would be display instantaneously BUT this approach has 2 pitfalls: we need theming and translations support.
In the end I don't think our About page is much slower (or at all slower) than the one from Slack ;)

DominikGuzei:
Ok thanks for the report 👍 you're right, this seems to be a problem with electron / rendering speed of the web app. It's not critical since the about page is not viewed regularly 😉

DominikGuzei:
Merging when CI is done

darko-mijic:
Let's not merge it until we update the list of people.

PR #436, opened 2017-08-24
Feature/daef 330 Send error logs and crash reports to the logging server

nikolaglumac:
This PR implements sending of crash reports to the remote server.
For this feature we rely on Electron's `crashReporter.setUploadToServer`
method https://electron.atom.io/docs/api/crash-reporter/#crashreportersetuploadtoserveruploadtoserver-linux-macos
Wether the application will or won't be sending the crash-reports to the remote server depends on user's opt-in status.

### Testing
To test if crash reporting works simply start Daedalus, open JS console and trigger: `process.crash()` after which new crash report should be generated locally AND sent to the remote server.

screen shot 2017-08-24 at 14 11 56


nikolaglumac:
Closing this PR since this feature is canceled - CrashReporter was completely removed in the scope of: https://github.com/input-output-hk/daedalus/pull/682

PR #445, opened 2017-08-28
[DAEF-448] load bundled app when running test suite to speedup page l…

DominikGuzei:
This PR improves the performance when running the whole test suite with `npm run test` by using the app in production mode instead of dev with webpack hot-reload server (which is slow).

This means for running all tests you also don't have to `npm run hot-server` in a second terminal anymore. But you have to `npm run build` once before running the tests! (which creates the `dist/bundle.js`)

However if you are developing test cases you should use `npm run test-watch` in combination with `npm run hot-server` and tags to get the same environment as we had before.


electron/main.development.js
```diff
@@ -141,35 +141,42 @@ app.on('ready', async () => {
...
-    notify(
-      ca,
-      function handleNotifyMessage(...args) {
-        if (messageCallback) {
-          try {
-            messageCallback(...args);
-          } catch (e) {
-            // The callback might have been released on page refresh etc.
-            messageCallback = null;
+    let isNotiyRunning = false;
```
darko-mijic: The value assigned to this variable is not really changing as far as I can see. It is defined as `let` so I suppose the intention was to change it, right?

DominikGuzei: Great catch, that's a bug - will fix now

DominikGuzei:
@darko-mijic the latest changes give a huge speedup to connecting logic! This PR is now ready for merge from my side.

PR #443, opened 2017-08-26
[DEVOPS-278] [STAGING] Daedalus fixes for Windows

deepfire:
1. "fix" DEVOPS-290 -- we need proper NSIS in the build environment: https://issues.serokell.io/issue/DEVOPS-290
2. update Daedalus launch flags
3. since the `dns` library on Windows is broken, provide a topology file with _relay_ IP addresses hard-coded -- this binds the installer to `staging`
4. adapt `build-certiicates-win64.bat` to the changes that happened since it was introduced -- most importantly, make it use the `%APPDATA%` folder for intermediate/generated output.
5. fix the launch script constructor ending with a `^` -- this was causing a Haskell exception in `cardano-launcher`
6. redirect `report-server` output to `report-server.aws.iohkdev.io`
7. the Daedalus log directory no longer exists prior to `build-certificates-win64.bat` execution, so it has to be created

This still ends up blocked by https://issues.serokell.io/issue/DAEF-437

PR #446, opened 2017-08-28
Daedalus fixes for master

deepfire:
This is https://github.com/input-output-hk/daedalus/pull/443 re-assembled, so it's mergeable into Daedalus `master`.

This PR, obviously won't function with `staging` without the corresponding addition of `daedalus-fixes-for-staging`.

To recap the summary:

0. some unused files were dropped (Darko)
1. a fix for OS X (Domen)
2. fix DEVOPS-290 -- we need proper NSIS in the build environment: https://issues.serokell.io/issue/DEVOPS-290
3. update Daedalus launch flags
4. since the `dns` library on Windows is broken, provide a topology file with _relay_ IP addresses hard-coded -- while this technically binds the installer to `staging`, the new Cardano can't be used on Windows otherwise anyway.
5. adapt `build-certiicates-win64.bat` to the changes that happened since it was introduced -- most importantly, make it use the `%APPDATA%` folder for intermediate/generated output.
6. fix the launch script constructor ending with a `^` -- this was causing a Haskell exception in `cardano-launcher`
7. the Daedalus log directory no longer exists prior to `build-certificates-win64.bat` execution, so it has to be created
8. the Electron's app is fixed to look for `tls/ca/ca.cert` in the proper location.


domenkozar:
This is stil missing macOS fixes for TLS

deepfire:
@domenkozar, addressed your suggestion.

electron/main.development.js
```diff
@@ -137,7 +137,7 @@ app.on('ready', async () => {
 
   try {
 
-    const ca = readCA(path.join(__dirname, '../tls/ca.crt'));
+    const ca = readCA(path.join(__dirname, '../../../tls/ca/ca.crt'));
```
darko-mijic: This change is breaking TLS configuration in development. With this configuration TLS configuration file needs to be outside of the project directory. Something to discuss tomorrow @domenkozar and @deepfire.

akegalj: Maybe we should put it in AppData as suggested by @domenkozar , where we have DB and logs

PR #457, opened 2017-08-31
Feature/daef 446 Do not block the UI while wallet is being restored/imported

nikolaglumac:
This PR introduces a change in the wallet restore and import workflows.
Workflow now looks like this:
- user opens wallet import/restore dialog
- fills in necessary data and submits the dialog
- after a delay of 500ms we check if the request has failed or if it still active
- in case of an error dialog is left open and error message is presented to the user
- in case the import/restore is still in progress we hide the dialog (which unblocks the UI) and show status-notification (e.g. `Wallet import in progress...`)
- when import/restore finishes status-notification is hidden again...

I believe this is as much as we can do without backend support (aka. adding `ready` flag on wallet object). Basically the wallet can be used from the moment it is appears in the UI with a few pitfalls:
- wallet addresses used state is unknown (all addresses appear as non-used)
- wallet transactions history is empty (no transactions shown)
- not sure if wallet send/receive actions would work (needs testing)

import

It is possible to have both restore and import processes running in parallel now (but it is **NOT** possible to have more than one active import or restore process in the same time)
screen shot 2017-08-31 at 13 29 17

## TODO
- [x] Inject missing Japanese translation for newly added status messages:
```
"wallet.statusMessages.activeImport": "Wallet import in progress",
"wallet.statusMessages.activeRestore": "Wallet restore in progress",
```
- [x] Keep import/restore dialog open if user is importing/restoring first wallet (there are no wallets)
- [x] Test with version of the backend that is not returning wallets which are being imported/restored:
https://issues.serokell.io/issue/CSM-435
https://github.com/input-output-hk/cardano-sl/pull/1483
- [x] Remove `do not merge` label before the release which will include required backend change from above, possibly 0.6.3 (0.6.2 will not have it included)
- [x] Make sure acceptance tests for import/restore work after this change is implemented


a-rukin:
Hi! Can we expand padding around separator line in "system-notification"? It's too close to text compare to main padding of those block

nikolaglumac:
@a-rukin like this?

screen shot 2017-08-31 at 15 04 25

a-rukin:
@nikolaglumac yeah, awesome!

nikolaglumac:
@a-rukin done ;)

akegalj:
@darko-mijic @nikolaglumac 
Looking good. Note that without `ready` functionality implemented in the backend, user might do some action with the wallet (ie, try to send money to the wallet hash that is still being importend or possibly some other action) and trigger some race condition in the backend.

Current wallet code was not written with this kind of concurrency in mind, so it is necessary for backend to add `ready` flag before this is being merged as we will most probably encounter some backend bugs in some special edge cases.

darko-mijic:
Reviewing this now.

darko-mijic:
We have translations:
```
“wallet.statusMessages.activeImport”: “ウォレットをインポート中“,
“wallet.statusMessages.activeRestore”: “ウォレットを復元中”
```

DominikGuzei:
@darko-mijic what's the status of this PR? @akegalj brought up some concerns but it's not flagged as `do not merge` nor `wip`

darko-mijic:
This PR will be merged after the RC launch @DominikGuzei. We don't want to introduce any risks. I believe Ante was referring to another branch in `cardano-sl`.

nikolaglumac:
@darko-mijic Japanese translations injected!

app/stores/WalletsStore.js
```diff
@@ -25,6 +25,8 @@ export default class WalletsStore extends Store {
   WALLET_REFRESH_INTERVAL = 5000;
 
   @observable active: ?Wallet = null;
+  @observable isImportActive = false;
```
darko-mijic: Flow type declaration is missing.

darko-mijic:
@nikolaglumac I have updated todo list, we should keep dialogs open if a user has no wallets. @DominikGuzei please review this PR too. This will most probably be included in the 0.8.2/0.6.3 release.

nikolaglumac:
@darko-mijic this PR is still WIP. Please wait with the reviews...

app/stores/WalletsStore.js
nikolaglumac: @darko-mijic fixed ;)

nikolaglumac:
@darko-mijic @DominikGuzei this feature is now done and is ready for review and testing.
Please notice that when running acceptance tests import/restore takes quite a while on staging so in order to avoid timeouts simply update DEAFULT_TIMEOUT to smth like `60000`: https://github.com/input-output-hk/daedalus/blob/master/features/support/electron.js#L7

akegalj:
Seems like we do have functionality for this here https://issues.serokell.io/issue/CSM-435 and it was merged to 1.0 https://github.com/input-output-hk/cardano-sl/pull/1483

cc @flyingleafe @Martoon-00

nikolaglumac:
The Api for this task has been implemented and tested against Cardano-sl master branch.
We have confirmed that it works as expected and therefore this task can be finalized.

app/api/CardanoClientApi.js
```diff
@@ -307,7 +307,7 @@ export default class CardanoClientApi {
   }
 
   async createAddress(request: CreateAddressRequest) {
-    Logger.debug('CardanoClientApi::createAddress called: ' + stringifyData(request));
+    Logger.debug('CardanoClientApi::createAddress called');
```
DominikGuzei: Why are we not logging the request here anymore?

nikolaglumac: Because createAddress can contain the passphrase! ;)

DominikGuzei: makes sense 😉 merging now!

PR #458, opened 2017-09-01
Fix/daef 395 Suppress electron window from forcing focus during acceptance testing

nikolaglumac:
This PR introduces a fix which prevents automatic application window focusing during acceptance testing.
Please note that the window still sets itself on top of other windows which can not be prevented BUT the good thing is that focus remains intact. Easy way to get around this is to simply drag the application window out of sight.


darko-mijic:
@domenkozar we have a failing build here.

PR #465, opened 2017-09-05
Fix/daef 466 Limit and validate Wallet name max length

nikolaglumac:
This PR introduces maximum length validation (and limiting) of wallet name attribute.
Wallet name maximum length is set to 40 characters.
In the scope of this PR sidebar styling was updated - until now wallet name shown sidebar items was truncated via char count (which has different results based on width of individual characters) and now we use more precise CSS method (text-oveflow: ellipsis).

## TODO
- [x] Inject update Japanese translation for invalid wallet name error message:
```
"global.errors.invalidWalletName": "Requires at least 3 and at most 40 letters."
```

screen shot 2017-09-05 at 13 43 25

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ Changelog
 - Improved QR code colors on wallet send screen for dark themes ([PR 448](https://github.com/input-output-hk/daedalus/pull/448))
 - Fixed failing acceptance-tests ([PR 454](https://github.com/input-output-hk/daedalus/pull/454))
 - Prevent application window auto focusing while running acceptance-tests ([PR 458](https://github.com/input-output-hk/daedalus/pull/458))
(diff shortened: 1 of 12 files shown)


DominikGuzei:
Will review this now

DominikGuzei:
Hiroto just sent me the translation for the error message:
`ウォレット名は3文字以上40文字以内である必要があります`

The direct translation to English would be `Wallet name requires at least 3 and at most 40 letters` which also seems to be more precise, so you could also update that 😉

nikolaglumac:
@DominikGuzei both English and Japanese translations are now updated.
Syncing label overlap will not happen in the mainnet and creating solution which doesn't overlap is to complex so it is decided we keep things as-is.

DominikGuzei:
Travis build fails due to broken mac install script … does not seem to be related to the frontend code changes in this PR.

PR #472, opened 2017-09-08
Fix/daef 475 Maximised window should not have max width and max height

nikolaglumac:
This PR removes window maximum width and height in full-screen mode which enables the application to expand to actual screen dimensions.
Still, outside full-screen mode standard maximum width/height dimensions are respected.

screen shot 2017-09-08 at 12 30 37


darko-mijic:
@domenkozar: We have a failing CI here.

domenkozar:
macOS situation has been fixed from travis-ci point of view - so restarting should help now.

PR #475, opened 2017-09-11
[DEVOPS-317] Windows installer: build cert synchronously

jmitchell:
It's possible for the Windows installer to "finish" before the build
certificates script has finished. This change may fix it, although it
hasn't yet been tested.

Before merging test the resulting CI build manually.

PR #478, opened 2017-09-13
[DEVOPS-317] Windows installer: build cert synchronously (#475)

jmitchell:
It's possible for the Windows installer to "finish" before the build
certificates script has finished.


jmitchell:
Change was already approved in #475, but it needs to be merged into `master`.

PR #483, opened 2017-09-17
[DO-NOT-MERGE]  Staging installers for `cardano-sl-1.0`

deepfire:
- [ ] get genesis and configuration from cardano-sl bridge


domenkozar:
@deepfire we also need to set: `NETWORK=mainnet` for travis/appveyor to change the UI.

.travis.yml
```diff
@@ -8,6 +8,7 @@ matrix:
 env:
   global:
     - VERSION=1.0-rc
+    - NETWORK=mainnet
```
deepfire: @domenkozar, does that look good?

darko-mijic: Looks good! @deepfire @domenkozar

.travis.yml
```diff
@@ -7,8 +7,9 @@ matrix:
     osx_image: xcode8.3
 env:
   global:
-    - VERSION=0.6
-    - CARDANO_SL_BRANCH=cardano-sl-0.6
+    - VERSION=1.0-rc
```
domenkozar: Should also bump the first line in `appveyor.yml`

installers/MacInstaller.hs
```diff
@@ -34,6 +34,8 @@ main = do
   copyFile "cardano-launcher" (dir <> "/cardano-launcher")
   copyFile "cardano-node" (dir <> "/cardano-node")
   copyFile "wallet-topology.yaml" (dir <> "/wallet-topology.yaml")
+  copyFile "configuration.yaml" (dir <> "/configuration.yaml")
```
domenkozar: This file should be copied from cardano-sl bridge, otherwise we'll have to keep it in sync in daedalus was well, that's 3 places including iohk-ops.

deepfire: @domenkozar, acknowledged

.travis.yml
domenkozar: fixed

domenkozar:
I've force pushed only changes for staging on top of current release branch.

domenkozar:
Outdated.

PR #485, opened 2017-09-18
[CSM-458] Query `postponeUpdate` endpoint when postponing the update

flyingleafe:
These changes should make Daedalus query `postponeUpdate` endpoint on postponing the update. The endpoint is introduced in https://github.com/input-output-hk/cardano-sl/pull/1586. This is needed to avoid problem with postponing the update described in https://issues.serokell.io/issue/CSM-458?query=%23CSM%20-Resolved

### DIFF
--- a/app/stores/NodeUpdateStore.js
+++ b/app/stores/NodeUpdateStore.js
@@ -3,7 +3,11 @@ import { observable, action } from 'mobx';
 import Store from './lib/Store';
 import Request from './lib/LocalizedRequest';
 import environment from '../environment';
-import type { NextUpdateResponse, PostponeUpdateResponse, ApplyUpdateResponse } from '../api';
+import type {
+  NextUpdateResponse,
+  PostponeUpdateResponse,
+  ApplyUpdateResponse,
+} from '../api';
 
 export default class NodeUpdateStore extends Store {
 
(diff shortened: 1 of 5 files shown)


nikolaglumac:
Reviewing this one now...

app/stores/NodeUpdateStore.js
```diff
@@ -40,6 +41,7 @@ export default class NodeUpdateStore extends Store {
   };
 
   @action _postponeNodeUpdate = () => {
+    this.postponeUpdateRequest.execute();
```
DominikGuzei: We should probably add some basic error handling here! Like
```javascript
@action _postponeNodeUpdate = async () => {
  try {
    await this.postponeUpdateRequest.execute();
    this.isUpdatePostponed = true;
  } catch(error) {
    this.isUpdatePostponed = false;
  }

nikolaglumac: Agreed @DominikGuzei - the only part that worries me is the fact that we should add "submitting" state on postpone button AND handle error state in the UI as if this call fails we will not hide the update notification so the user should be noticed that the error happened...

DominikGuzei: Yeah definitely, the whole error handling logic needs to be fleshed out for the UI

PR #486, opened 2017-09-19
Feature/daef 298 Japanese terms of use for the mainnet

tomothespian:
This PR introduces Japanese terms of use for the mainnet.


nikolaglumac:
@darko-mijic please also give this update a view - I want to make sure nothing is missed ;)

PR #487, opened 2017-09-19
Fix/daef 488 Fix JS error on create-dialog's last step checkbox

DominikGuzei:
This PR bumps the `react-polymorph` version to 0.3.5 which includes a fix for `label` prop type on checkboxes, which in turn solves the error we have atm. when assigning non-strings (e.g html formatted text) as labels.


nikolaglumac:
@DominikGuzei don't we need a changelog item for this fix?

DominikGuzei:
Closing this PR in favor of new target against `release/0.8.2`

PR #491, opened 2017-09-20
[DAEF-420] show transaction status based on condition

DominikGuzei:
This PR adds support for the `ctCondition` property that comes with every transaction from the backend and provides information about its status in the blockchain. It can have the following values:

- `CPtxApplying` transaction application is in progress
- `CPtxInBlocks` transaction is detected to be in blockchain
- `CPtxWontApply`  transaction can't be applied (and it will not with up to high assurance, i.e. unless severe rollback occurs, UPD: this is not true now, this status is permanent now)
- `CPtxNotTracked`  transaction was made not by us / before last wallet restore - this state never changes

Which are simplified by our frontend into:
`export type TransactionState = 'pending' | 'failed' | 'ok';`

screenshot 2017-09-21 um 16 07 10

screenshot 2017-09-21 um 16 07 26

## TODOS:

- [x] Implement final design specs for all themes
- [x ] Try with working (real) backend / transactions (i built this with storybook now because i have no keys for wallets to work with atm)
- [x] Inject missing Japanese translations:
```
  wallet.transaction.state.failed: !!!Transaction failed
  wallet.transaction.state.pending: !!!Transaction pending
```


darko-mijic:
CC @a-rukin

app/domain/WalletTransaction.js
```diff
@@ -41,4 +50,12 @@ export default class WalletTransaction {
     return assuranceLevels.HIGH;
   }
 
+  getState(): TransactionState {
+    switch(this.condition) {
+      case 'CPtxApplying': return 'pending';
```
Martoon-00: Well actually, if you ever have a short doc over this, meaning of `pending` would hardly have one-sentence explanation, because incoming transaction with `CPtxNotTracked` state could be not yet in blockchain as well. But yep, for now it's the closest approximation.

a-rukin:
for failed transactions we have separate icon, also updated those animated statuses for "pending" and "retrying"
https://zpl.io/2ynM1yV

a-rukin:
if we can't track retrying transactions, then ignore them in zeplin

app/domain/WalletTransaction.js
DominikGuzei: Ok thanks for the explanation @Martoon-00 if you agree that this is a good-enough abstraction for end-users then i will continue like that?!

Martoon-00: Sure :wink:

DominikGuzei:
@a-rukin @darko-mijic i implemented the latest design specs now (with pending animation bar), please review again!

@a-rukin please notice that the pending bar does not look 100% the same as in the designs, please consider if this is good enough because it would be additional work to change it … if you like this version too then let's go with it 😉

nikolaglumac:
This PR contains fixed styling for confirmation-dialog on dark-theme:
screen shot 2017-09-22 at 17 51 40

PR #507, opened 2017-09-29
Remove maximum window size limitations

adlk:
Limiting setMaximumSize does not really serve a purpose. It is actually more annoying to not being able to maximize the window on a 13" retina display.

### DIFF
--- a/electron/main.development.js
+++ b/electron/main.development.js
@@ -185,7 +185,6 @@ app.on('ready', async () => {
   });
 
   mainWindow.setMinimumSize(900, 600);
-  mainWindow.setMaximumSize(1500, 2500);
 
   // Initialize our ipc api methods that can be called by the render processes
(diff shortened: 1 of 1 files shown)


nikolaglumac:
Not sure when this was planned but it was not done in correct manner...
We still have max-window-size limitations: https://github.com/input-output-hk/daedalus/blob/master/electron/main.development.js#L239-L246
A task will be created in Youtrack to solve this.

@darko-mijic @darko-mijic cc

DominikGuzei:
Thanks @nikolaglumac, didn't see that extra call … the magic numbers should also be turned into constants.

nikolaglumac:
@DominikGuzei since the plan is to remove these numbers we will not have anything to turn into constants :)

DominikGuzei:
Just sayin … 😆

PR #511, opened 2017-10-05
Chore/refactor to multi currency setup

DominikGuzei:
This PR is a huge refactoring to our discussed multi-currency setup.

- [x] One api namespace per currency like `api.etc.getSyncProgress` and `api.ada.getSyncProgress`
- [x] All apis are available at the same time (that's why you can start the app with `API=etc npm run dev` now and it only uses etc stuff for the loading page but ada for the rest. This will change as soon as we start to swap out containers to build etc specific versions)
- [x] All currency specific stores are namespaced under their currency (e.g: `stores.ada.wallets`
- [x] We can reuse / share logic on all layers:
   - api: flow types and common errors can be shared in `api/common.js`
   - stores: The first level of stores is currency agnostic and can use `environment.API` to access currency specific universal apis like `getSyncProgress` (see example in the `NetworkStatusStore`)
   - components: loaded as implemented by nikola (components in sub-folders override generic ones)

I think with this setup we can start working on the etc client for real 😉


DominikGuzei:
@darko-mijic acceptance tests are fixed now (although sending money does not work 100% correctly with the backend at commit `516641351cdf510ddbc894ebce1793381cbf4489`) - from my side this PR is now ready for review and merging 👍

nikolaglumac:
@DominikGuzei actions namespacing is now also done and all acceptance-tests are green. We are ready for the merge 🎉

DominikGuzei:
Great work @nikolaglumac 👍  merging now 🎉

PR #513, opened 2017-10-10
[DEVOPS-384] Single source of truth for genesis/configuration.yaml

domenkozar:
Now pick whatever cardano-sl ships with.


domenkozar:
Depends on https://github.com/input-output-hk/cardano-sl/pull/1744

domenkozar:
There is noone around to review, I'm merging this to unblock @fersel

deepfire:
So, the key change is at https://github.com/input-output-hk/daedalus/pull/513/files#diff-636efe9ceea923a2b78adb0bdcf1982fR109, as I understand it -- these files now come from the Cardano archive.

Looks good!

PR #517, opened 2017-10-13
[CSL-1747] Adapt to newer launcher features

neongreen:
**Note: this PR shouldn't be merged until https://github.com/input-output-hk/cardano-sl/pull/1840 is merged.**

This PR does two things:
* it instructs the launcher to write its logs into `Logs/pub`
* it increases the node killing timeout to 30s (from 5s) because now the launcher uses SIGKILL and sigkilling the node after it's been given only 5×2 seconds to terminate is too harsh


deepfire:
Looks good to me.

PR #526, opened 2017-10-26
Fix/daef 520 Improve messages for update notifications

darko-mijic:
PR improves messages for update notifications, updates changelog for 0.8.3 and bumps application version in the about dialog.

screen shot 2017-10-26 at 09 17 02

screen shot 2017-10-26 at 09 39 54

screen shot 2017-10-26 at 09 42 18


darko-mijic:
@fersel Did you get a chance to look at this?

fersel:
Not yet, sorry. Though it looks better now, in the initial issue I wanted another fact to be stressed out in the update message: _if you don't choose anything, update will still be installed on the next start._ Currently user may think that if s/he opts to not choose anything the update will never be installed, so the purpose is to clear up this case. Thus I suggested a message "Note: the update will be installed automatically on the next Daedalus launch" - but I see that it might be not too good either.

darko-mijic:
OK, good point @fersel. We will update this.

darko-mijic:
This is my proposition for the final solution @fersel:
 
```
"cardano.node.update.notification.message": "Would you like to install the update? If you choose to postpone, or if you take no action at all, the update will be installed automatically on the next Daedalus launch.",
```

darko-mijic:
@fersel we will merge this into the release branch.

PR #531, opened 2017-11-06
Feature/daef 515 instructions for setting up NTP on Daedalus website

tomothespian:
This PR introduces instructions for setting up NTP on Daedalus website.

**TODOs:**
- [x] Temporarily hide "See problem solutions" button as there is no web link ready
- [x] Implement acceptance-test for this new feature
- [x] Add 2 seconds polling for isSystemTimeCorrect Api call
- [x] Integrate Api call and remove hardcoded "isSystemTimeCorrect" value from NetworkStatusStore
- [x] Inject missing Japanese translations:
```
"systemTime.error.overlayButtonLabel": "!!!See problem solutions",
"systemTime.error.overlayText": "!!!ATTENTION: Time of your machine is different from global time. You are 2 hours 12 minutes 54 seconds behind. You need to fix issue, because you are gonna be unable to sync system!",
"systemTime.error.overlayTitle": "!!!Daedalus Sync Error",
```

systemtimeerroroverlay


nikolaglumac:
UI part looks great @tomothespian!
We need to wait for Api integration in order to merge this PR...

nikolaglumac:
@darko-mijic we need a URL to point the "See problem solutions" button to. Thanks for providing it!

nikolaglumac:
@darko-mijic @DominikGuzei @tomothespian in the scope of this PR the Api call for checking user's local time sync is implemented as a "slow" poller which is triggered on the application start and in 24h cycles after that. I'm afraid this is not enough. There are two different situations where this would break:
1. If during initial application start time check there is time difference detected and user is presented with the NTP notification, even if the user fixes the local time Daedalus will remain blocked with the notification (as it is not aware of this change - will take another 24h or a manual application restart for the change to be noted)
2. If during normal application usage user on purpose changes the local time Daedalus will not be aware of this and that would probably result in many errors.
...both of this cases point to a need of polling local-time-check call more often - probably every few seconds or so.
We should discuss this.

nikolaglumac:
@darko-mijic @DominikGuzei this PR is ready for review ;)

DominikGuzei:
@nikolaglumac what about the merge conflicts?

nikolaglumac:
@DominikGuzei master has etc-client stuff now. We will have to fix the conflicts obviously...

tomothespian:
@DominikGuzei @nikolaglumac latest master merged.

app/stores/NetworkStatusStore.js
```diff
@@ -199,6 +231,19 @@ export default class NetworkStatusStore extends Store {
     }
   };
 
+  _redirectToSyncingWhenLocalTimeDifferent = () => {
+    if (
+      this.localTimeDifference > this.ALLOWED_TIME_DIFFERENCE &&
+      !this.isSynced &&
+      this.stores.app.currentRoute !== ROUTES.PROFILE.LANGUAGE_SELECTION &&
+      this.stores.app.currentRoute !== ROUTES.PROFILE.TERMS_OF_USE &&
+      this.stores.app.currentRoute !== ROUTES.PROFILE.SEND_LOGS
```
DominikGuzei: I think we should move these "if we are not on one of the setup pages" checks into a dedicated, reusable function. We have that already in a few places as far as i remember

app/api/ada/index.js
```diff
@@ -675,6 +678,18 @@ export default class AdaApi {
       throw new GenericApiError();
     }
   }
+
+  async getLocalTimeDifference(): Promise<GetLocalTimeDifferenceResponse> {
+    Logger.debug('AdaApi::getWallets called');
```
DominikGuzei: Copy & paste sin! 😆

app/api/ada/index.js
```diff
@@ -675,6 +678,18 @@ export default class AdaApi {
...
+    try {
+      const response: AdaLocalTimeDifference = await getAdaLocalTimeDifference({ ca });
+      Logger.debug('AdaApi::getLocalTimeDifference success: ' + stringifyData(response));
+      return response;
+    } catch (error) {
+      Logger.error('AdaApi::getWallets error: ' + stringifyError(error));
```
DominikGuzei: Here too 😉

app/api/ada/mocks/patchAdaApi.js
```diff
@@ -36,4 +38,13 @@ export default (api: AdaApi) => {
     }
     return { amount: new BigNumber(1000) };
   };
+
+  api.getLocalTimeDifference = async () => {
+    const response = Promise.resolve(LOCAL_TIME_DIFFERENCE);
+    return response;
```
DominikGuzei: This function can be shortened to a one-liner

app/components/loading/SystemTimeErrorOverlay.js
```diff
@@ -0,0 +1,87 @@
...
+        <SvgInline svg={attentionIcon} className={styles.icon} />
+
+        <h1>{intl.formatMessage(messages.overlayTitle)}</h1>
+
+        <p>{intl.formatMessage(messages.overlayText, { behindTime })}</p>
+
+        {/* <Button
+          label={intl.formatMessage(messages.buttonLabel)}
+          skin={<SimpleButtonSkin />}
+        /> */}
```
DominikGuzei: Why is this button commented out?

app/api/ada/index.js
tomothespian: aaah, my bad hehe 😆

app/components/loading/SystemTimeErrorOverlay.js
nikolaglumac: This was requested by @darko-mijic...

PR #533, opened 2017-11-06
Chore/daef 525 Update to latest mobx-react-form

nikolaglumac:
This PR introduces latest available update of MobX React Form plugin.
Biggest change was the way `file` type fields are handled as from now you have to use `set` method in place of `onChange` in order to update field's value (e.g. on file drop or selection).

### Note
As a part of this PR a fix for `isValidAddress` endpoint regression issue has been implemented: https://issues.serokell.io/issue/CSM-557 (endpoint started to throw an error if invalid address has been passed to it - where in fact it should return success with boolean value `false`)


DominikGuzei:
Reviewing this now

app/components/wallet/file-import/WalletFileImportDialog.js
```diff
@@ -215,7 +215,9 @@ export default class WalletFileImportDialog extends Component {
           <FileUploadWidget
             {...walletFile.bind()}
             selectedFile={walletFile.value}
-            onFileSelected={walletFile.onChange}
+            onFileSelected={(file) => {
+              walletFile.set(file);
+            }}
```
DominikGuzei: Why not keep it simple as before? `onFileSelected={walletFile.set}` 😉

nikolaglumac: Tried that but it does not work... Believe it or not ;)

nikolaglumac: Tried that but it does not work... Believe it or not ;)

DominikGuzei: Ah ok … so probably the method is not bound then 😉  then put a comment there so we know for future reference 👍

nikolaglumac: @DominikGuzei comment added - hope it is fine now ;)

PR #538, opened 2017-11-09
Italian translation

purplesoft-io:


nikolaglumac:
Thank you for your contribution! Unfortunately, at this moment in time, we can not accept this translation as it would be hard to maintain it without internal translation/audit process.

PR #542, opened 2017-11-15
[DAEF-542] handle special etc tx errors

DominikGuzei:
This PR adds handling of special errors that can be thrown by the mantis backend when sending ETC.

```
personal_sendTransaction:
    LogicError("No key found for the given address")
    LogicError("account is locked or unknown")
```

# WAITING FOR from Mantis team:

- [ ] Explanation of what these error messages actually mean for the end user / Daedalus (so we can write up a proper error message
- [x] How can we trigger/test these errors with Daedalus?
- [ ] my mantis/daedalus setup is broken so i could not properly test these

### DIFF
--- a/app/api/etc/errors.js
+++ b/app/api/etc/errors.js
@@ -1,0 +1,33 @@
+import { defineMessages } from 'react-intl';
+import LocalizableError from '../../i18n/LocalizableError';
+
+const messages = defineMessages({
+  noKeyForGivenAddress: {
+    id: 'api.errors.etc.NoKeyForGivenAddress',
+    defaultMessage: '!!!No key found for given address.',
+    description: 'Error message that is displayed when no key was found for a given address.'
+  },
+  accountLockedOrUnknown: {
+    id: 'api.errors.etc.AccountLockedOrUnknown',
+    defaultMessage: '!!!Account is locked or unknown',
+    description: 'Error message that is displayed when account is locked or unknown.'
+  },
+});
+
+export class NoKeyForGivenAddress extends LocalizableError {
+  constructor() {
+    super({
+      id: messages.noKeyForGivenAddress.id,
+      defaultMessage: messages.noKeyForGivenAddress.defaultMessage,
+    });
+  }
+}
+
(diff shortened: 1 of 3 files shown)


DominikGuzei:
This PR is closed because the error messages are not relevant since they should never happen in Daedalus … for reference here is the discussion between Lukasz and me:

```
Lukasz Gasior [11:36 AM] 
I think "account is locked or unknown" is returned if you try to send a tx with a sender address that is not managed by this node (if you always send tx along with a passphrase in a single request you can ignore the "is locked" part)


[11:38] 
"key not found" is returned when there is no keyfile for given address on disk


[11:39] 
to test the first one, just remove keyfile from disk (or try to send a tx with random sender address) (edited)


[11:41] 
the latter happens if you try to send a tx (without a passphrase in a request) from an unlocked account


[11:42] 
you either need to unlock it first or provide the password


Dominik Guzei
[12:05 PM] 
ok so basically these errors are not relevant for us … because Daedalus users are not supposed to be able to send from unlocked/not existing accounts/after key file has been deleted (or not managed by node) (edited)


Lukasz Gasior [12:08 PM] 
yes, it seems these errors shouldn't really happen
```

nikolaglumac:
@DominikGuzei what about situation where user tries to send amount greater than the balance of his wallet? Do we handle this situation?

DominikGuzei:
Good catch @nikolaglumac i didn't see any mantis error that corresponds to that - will have to ask if there is anything we can use

nikolaglumac:
We should see how to handle that situation PLUS possible errors on calculate-transaction-fees ;)

DominikGuzei:
@nikolaglumac i didn't see any relevant errors for tx fees endpoint either 😉 so we just have the general case of "something went wrong" … all other cases are only relevant if you call the endpoint in a wrong way which we should never be doing.

DominikGuzei:
@nikolaglumac mantis guys say that there are no error messages for these cases (no enough money / tx fees) and they don't think that this should be checked / thrown.

PR #547, opened 2017-11-22
[DAEF-551] rewrite notification logic to fix acceptance tests

DominikGuzei:
This PR fixes the broken acceptance tests for the restore and import features.
The original problem was that the notifications were not shown in dev mode because the backend is "too fast" and we waited 500ms before showing the notification.

This is now reversed: we always show the notification + at least for 500ms (which also makes sense in terms of UX, although that scenario only happens in dev mode).


DominikGuzei:
Good catch @nikolaglumac, the bug is fixed now!

DominikGuzei:
@nikolaglumac all bugs are now resolved and i added a scenario for the wallet already imported error case 😉 

BTW: this is also an example of how i would like to refactor our acceptance test code to be more reusable (move steps code into dedicated functions), and thus more specific (we can have many more step definitions because they only map to function calls instead of implementing the logic themselves).

nikolaglumac:
@DominikGuzei reviewing this now...

PR #558, opened 2017-11-27
Chore/daef 553 get test reset working for etc

DominikGuzei:
This PR introduces a new `api.etc.testReset` method which deletes all wallets … and refactors the `createWallet` and `restoreWallet` endpoints to use the new `importWallet` that works for private keys.


DominikGuzei:
Great @nikolaglumac, will merge this now!

PR #589, opened 2017-12-15
Fix/ddw 62 Disable transactions sorting in Daedalus

nikolaglumac:
This PR removes sorting of transactions from Daedalus side and keep Api response original transactions order.


nikolaglumac:
@DominikGuzei please do the review once again. I have now updated `getTransactions` Api endpoint handling. Mantis `daedalus_getAccountTransactions` endpoint is updated to return a single array of all transactions (sent+received) and each of the transactions in this response will have an additional `isOutgoing` flag which we use to determine if the transactions is a sent or received.

### Note:
Mantis branch with this Api update is yet to be created (so you will not be able to really test if it works until we get the correct branch from ETC team)

nikolaglumac:
This update should be tested against Mantis running on `feature/txOrder` branch.

FYI: @DominikGuzei @darko-mijic

DominikGuzei:
Great work 👍 tested against `feature/txOrder` mantis branch and ready to be merged as soon as we have CI passing!

PR #590, opened 2017-12-15
Fix/ddw 63 ETC transaction amount validation

nikolaglumac:
This PR introduces a more precise ETC transaction amount validation.
Until now we only made sure that the transaction amount is a positive number, while the updated validator checks if the value is greater than zero and less or equal to current wallet balance.

### Note:
Updated validation does not take the transactions fees into account - therefore it is still possible to create a transaction which would require more funds than available in the wallet.

screen shot 2017-12-15 at 17 34 53


DominikGuzei:
Reviewing this now …

nikolaglumac:
@DominikGuzei we need one more review after my latest update.
This update adds default transaction fees (HARDCODED TO: 0.00042 ETC) into transaction amount validator.

screen shot 2017-12-18 at 12 16 07
screen shot 2017-12-18 at 12 15 11

DominikGuzei:
Great work @nikolaglumac, merging now 👍

PR #594, opened 2017-12-17
[CSL-1918] Pass launcher args via config file

Pastafarianist:
Summary of changes:

1. Use config file (`launcher-config.yaml`) to pass parameters to `cardano-launcher` instead of command-line arguments. This entails rewriting `Launcher.hs` and modifying `WindowsInstaller.hs` and `MacInstaller.hs`. The relevant changes to `cardano-launcher` are in https://github.com/input-output-hk/cardano-sl/pull/2076.
2. Add dependency on `universum` and enable `NoImplicitPrelude` by default.
3. Fix indentation to 4 spaces, use explicit imports & exports, apply `stylish-haskell` everywhere. (Since there are a lot of whitespace changes, the diff is better viewed with the `w=1` GET parameter, [like so](https://github.com/input-output-hk/daedalus/pull/594/files?w=1).)

**HELP WITH TESTING IS WANTED,** both on Mac OS and Windows. Until we make sure that this works under both OSs, this must not be merged. This should be a drop-in replacement for the previous version, so testing procedures should not change.

Also help with fixing Nix build (see Travis logs) would be welcome.

A branch which switches the CSL branch used to build Daedalus (and this allows to build correct installers) can be found [here](https://github.com/input-output-hk/daedalus/compare/thatguy/csl1918-pass-launcher-args-via-config-file...thatguy/csl1918-prepare-installers?expand=1).


installers/Installer.hs
```diff
@@ -1,18 +1,20 @@
-import           Data.Text as T
-import           Data.Monoid ((<>))
-import           System.Info
-import           Turtle
-import           Turtle.Line          (unsafeTextToLine)
+{-# LANGUAGE NoImplicitPrelude #-}
```
domenkozar: Isn't this the default extension now?

domenkozar:
Depends on https://github.com/input-output-hk/cardano-sl/pull/2166 to unblock appveyor

installers/Installer.hs
Pastafarianist: I had doubts about removing these pragmas because I am not familiar with how `nix-build` treats `cabal` files (whether it knows about them at all), and also because `RewriteLibs.hs` was clearly intended to be run as a standalone script. I've removed `NoImplicitPrelude` from all files except for `RewriteLibs.hs` now.

Pastafarianist:
The error is still there even though I restarted the build yesterday after https://github.com/input-output-hk/cardano-sl/pull/2166 was merged. @domenkozar, is there anything else that needs to be done for passing CI aside from restarting it?

domenkozar:
@Pastafarianist I think my latest commit should fix it.

domenkozar:
@fersel ready to test

domenkozar:
@fersel @Pastafarianist what's the status of this?

Pastafarianist:
@domenkozar Windows build works, awaiting news about Mac from @fersel.

Pastafarianist:
Myself and @fersel have not found any issues with the latest version, so we think the PR is ready to be merged as soon as https://github.com/input-output-hk/cardano-sl/pull/2076 is merged.

domenkozar:
@Pastafarianist I have one question left: does now cardano-launcher change it's working directory? That was the purpose of the .bat script

Pastafarianist:
@domenkozar I'm looking at https://github.com/input-output-hk/daedalus/pull/594/files#diff-936ebc51c2ddde384bd5cf4fde429808L17 and I don't see any calls to `cd`. Am I misunderstanding something?

domenkozar:
https://github.com/input-output-hk/daedalus/pull/594/files#diff-936ebc51c2ddde384bd5cf4fde429808L20 this sets CWD for windows

Pastafarianist:
I see, the `/D` flag changes the current directory. No, there are no calls to `chdir` in `cardano-launcher`.

domenkozar:
In that case we have to add that in or keep the .bat file (I'd prefer the former).

Pastafarianist:
After some discussion in Slack we came to the conclusion that changing the current directory is not needed.

PR #606, opened 2017-12-18
Chore/ddw 65 update terms of use for ETC version

tomothespian:
This PR introduces update for ETC version of terms of use.


nikolaglumac:
I have added special text for Terms-of-use acceptance page checkbox in case user has selected Japanese language and we have Terms-of-use only in English. In this case the label will show Japanese version of `I understand that the terms of use are only available in English and agree to the terms of use`. This text is used for ETC Terms-of-use acceptance page under Japanese locale while normal version `I agree with terms of use` is used for Cardano.

FYI @darko-mijic @DominikGuzei @tomothespian

DominikGuzei:
Reviewing this now …

PR #639, opened 2018-01-03
purge-translations now works on Windows

SebastienGllmt:
rm -rf doesn't work on Windows and so

> npm run dev

would cause an error.

I changed it to use rimraf instead which is a cross-compatible package. See documentation [here](https://www.npmjs.com/package/rimraf)


DominikGuzei:
Thanks for the notice @SebastienGllmt, im including your fix in #692

PR #667, opened 2018-01-15
[DEVOPS-622/github #664]:  deprecate non-sha256 MD support in the cert generator

### LINKED ISSUE #664: [build-certificates-win64.bat] Fix OPENSSL_MD replacement
After DAYS of research of the https://github.com/input-output-hk/daedalus/issues/608 issue on Windows 7, getting no answers from the developers, and reading many people's suggestions over the net that this issue is not caused by the Daedalus wallet and should be fixed by the user, here I present to you:

[build-certificates.log]
OPENSSL_MD is an unsupported message digest type

This happens because the replacement in step 2 doesn't work as intended.

Workaround:

Manually change all occurrences of OPENSSL_MD to sha256 in ca.conf, client.conf and server.conf, and rerun the script.

Now launch Daedalus, and surprise - the endless "connecting to network" screen is now gone, and you can actually start using your wallet.

There are so many users affected by this issue, and it must be fixed ASAP. I have no idea why it has persisted up until now.

deepfire:
This fixes #664 by removing the logic that used to choose the OpenSSL MD on OS X -- this sometimes breaks on Windows in an obscure way, apparently due to Powershell incompatibilities.

This also essentially breaks OS X Yosemite (circa 2014 and earlier) -- but those are arguably insecure to use: the last update, 10.10.5 is dated 13 Aug 2015.

Tested on Windows 7.

Sister mainnet hot-fix is https://github.com/input-output-hk/daedalus/pull/666

### DIFF
--- a/installers/WindowsInstaller.hs
+++ b/installers/WindowsInstaller.hs
@@ -1,7 +1,8 @@
 module WindowsInstaller where
 
-import           Control.Monad      (unless)
+import           Control.Monad      (unless, forM_)
 import qualified Data.List          as L
+import qualified Data.List.Split    as L
 import           Data.Maybe         (fromJust, fromMaybe)
 import           Data.Monoid        ((<>))
 import           Data.Text          (pack, split, unpack)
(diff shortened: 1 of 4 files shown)


domenkozar:
According to https://en.wikipedia.org/wiki/OS_X_Yosemite - last update was 5 months ago. We can't just drop macOS support for such a small reason.

domenkozar:
I think better approach is to replace these when we create the installer, as it is static for windows at runtime (just always use sha256).

domenkozar:
Also, this lacks an issue id.

deepfire:
@domenkozar, redone as per your suggestion, and retested on Windows.

deepfire:
Revalidated.

PR #670, opened 2018-01-19
[DO-NOT-MERGE] staging-1.1 installers

deepfire:
This PR is to facilitate maintenance of the installer builds.


domenkozar:
This needs to be rebased and bridge removed.

PR #678, opened 2018-01-25
Fix/ddw 76 Time of your machine is different from global time. You are 0 seconds behind

tomothespian:
This PR fixes a bug where "You are 0 seconds behind" message is shown to the user when the time on the user's machine is slightly off. Threshold has been changed to 15 seconds to fix this issue. Polling interval has been reduced from 2000 ms to 500 ms.


darko-mijic:
CC @domenkozar

app/stores/NetworkStatusStore.js
```diff
@@ -185,11 +185,13 @@ export default class NetworkStatusStore extends Store {
...
-      const response = await this.localTimeDifferenceRequest.execute().promise;
-      runInAction('update time difference', () => (this.localTimeDifference = response));
-    } catch (error) {
-      runInAction('update time difference', () => (this.localTimeDifference = this.ALLOWED_TIME_DIFFERENCE + 1));
+    if (this.isConnected) {
+      try {
+        const response = await this.localTimeDifferenceRequest.execute().promise;
+        runInAction('update time difference', () => (this.localTimeDifference = response));
+      } catch (error) {
+        runInAction('update time difference', () => (this.localTimeDifference = this.ALLOWED_TIME_DIFFERENCE + 1));
```
DominikGuzei: Does this mean that we show the "out of time sync" error screen anytime the request fails?

app/stores/NetworkStatusStore.js
```diff
@@ -27,14 +27,14 @@ export default class NetworkStatusStore extends Store {
   _startTime = Date.now();
   _startupStage = STARTUP_STAGES.CONNECTING;
 
-  ALLOWED_TIME_DIFFERENCE = 0;
+  ALLOWED_TIME_DIFFERENCE = 15 * 1000000; // 15 seconds
```
DominikGuzei: What unit is this difference measured in? Normally we are dealing with milliseconds, so 15 seconds would be `15 * 1000` right?

nikolaglumac: In this case we are dealing with `microseconds`. You can easily check by changing your machine time in the settings.

app/stores/NetworkStatusStore.js
nikolaglumac: Yes this is true. That is what we agreed in the beginning.
Do you think we should change this? 

@darko-mijic cc

DominikGuzei: If we agreed to this behavior all is good 😉 just wanted to ask to be sure

nikolaglumac: I am open to any suggestions - do you think we should change this logic?

DominikGuzei: Hmm … im not sure, i guess this would only happen if the node is down? In that case we should be falling back to the "connecting" stage anyway, or am i mistaken there?

nikolaglumac: @DominikGuzei you are right - in case we get disconnected user will be redirected to connecting screen. Still after discussion with @darko-mijic we decided to reset timeDifference to zero in case of error response from localTimeDifferenceRequest.
@tomothespian please fix this ☝️

tomothespian: done!

nikolaglumac:
We can merge this PR as soon as @DominikGuzei approves it (and CI passes) ;)

PR #680, opened 2018-01-25
Chore/ddw 80 Update for the about screen

nikolaglumac:
This PR introduces updated About screen content.
The content now includes Daedalus `platform` and `build` information as well as the Api `name` and `version` details.

screen shot 2018-01-25 at 12 45 48

### TODO:
- [x] Inject missing Japanese translations


nikolaglumac:
@DominikGuzei @darko-mijic this PR can be reviewed - I have put the WIP tag because of the missing Japanese translations. Will add them in as soon as I get them ;)

app/components/static/About.js
```diff
@@ -90,6 +95,24 @@ export default class About extends Component<any> {
 
     const apiIcon = environment.isAdaApi() ? cardanoIcon : mantisIcon;
 
+    let platform;
+    switch (environment.platform) {
+      case 'darwin':
+        platform = 'macOS';
+        break;
+      case 'win32':
+        platform = 'Windows';
```
DominikGuzei: Are we sure that the names of the OSes are universally used in all languages (without translation)?

nikolaglumac: No we are not :)
I have requested the translations and if the Japanese translation is different then I will turn these hardcoded values into translation-keys.
The same is for the Cardano/Mantis names...

nikolaglumac: @DominikGuzei I've received translations and it turns out macOS/Windows/Linux is the same in Japanese as in English so I will leave it like is.

nikolaglumac:
@DominikGuzei @darko-mijic please review this one again - I have added missing Japanese translations.

DominikGuzei:
Great work 🎉 ready to merge when CI is green

PR #688, opened 2018-01-26
Fix/ddw 72 Update copy for the sync error screen 

tomothespian:
This PR updates copy of the sync error screen.

incorrect system time overlay


DominikGuzei:
@tomothespian please fix the conflict with the default messages … my `manage-translations` setup still doesn't work properly … i think we should do a hangout soonish where we go over my setup to fix it 😉

PR #692, opened 2018-01-26
Chore/ddw 47 improve build system

DominikGuzei:
This PR introduces a simplified Gulp + webpack setup that works reliable and without any hacks. It also removes any dev servers and hot-code-reloading since that didn't really work with our codebase anyway. We can still invest some time later on to improve the speed of the auto-reloading on code changes but i think it's acceptable for the moment.

# Todos:
- [x] `API=etc` works correctly too
- [x] Acceptance tests are all green (apart from the two sending transaction failures due to the pending tx issue -> is that already fixed in some backend branch?)
- [x] Added console logging for failed acceptance tests to make it easier to debug main/render process problems
- [x] Linting 
- [x] npm run flow-test
- [x] Fix broken `npm run manage-translations`
- [ ] Fix broken `npm run package`
- [x] Add runtime `NODE_ENV` which sets `enviroment.current` to `test` during tests run


DominikGuzei:
@nikolaglumac im merging this now into `development-webpack-fork` since it's definitely better as what i have there 😉 and it's easier to have only one "fork" branch 😂

PR #701, opened 2018-01-31
[DEVOPS-646] staging installers - DONT MERGE

domenkozar:


.travis.yml
```diff
@@ -9,7 +9,7 @@ env:
   global:
     - VERSION=1.1
     - NETWORK=mainnet
-    - CARDANO_SL_BRANCH=release/1.1.0
+    - CARDANO_SL_BRANCH=1.1.0-staging
```
domenkozar: Why do we need a separate branch for staging on cardano side?

deepfire: @domenkozar, because there is no sure-fire way to determine if the current combination of versions will work -- so I was hesitant to make a PR to `cardano-sl`.  If the branch produces an installer that actually works -- then I think we're safe to make that PR.

At least that's my idea.

domenkozar: Ah, sounds good! I would prefer another daedalus PR also for such experiments, since this is considered a staging de-facto usable installer.

deepfire: But we don't have usable staging installers yet -- _and_ this branch isn't supposed to be merged (unlike the Cardano one).  What purpose the extra PR would serve?

domenkozar: Once I turn on mirroring they will be functional, regardless of this change :)

darko-mijic:
@domenkozar One of the Travis builds fails. How can we get Mac installer?
cc @nikolaglumac

domenkozar:
Click on the buildkite CI -> click `daedalus-x86_64-darwin` -> artifacts tab

.travis.yml
domenkozar: @gromakovsky will correct this

installers/launcher-config-windows.yaml
```diff
@@ -33,11 +33,11 @@ updateArchive: null
 updateWindowsRunner: ! '%APPDATA%\Daedalus\Installer.bat'
 
 nodeTimeoutSec: 30
-reportServer: http://report-server.cardano-mainnet.iohk.io:8080
+reportServer: http://report-server.awstest.iohkdev:8080
```
nikolaglumac: @domenkozar @cleverca22 I believe that you have a missing `.io` part here. The `reportServer` url should read: `http://report-server.awstest.iohkdev.io:8080`?

domenkozar: yes :(

nikolaglumac: OK. But still - even with the "correct" url: http://report-server.awstest.iohkdev.io:8080 I get error 500 `Internal server error` :((

nikolaglumac: @domenkozar @cleverca22 I see this issue is still not fix and it blocks us from testing of the send-bug-report feature via the installers :(
Can you please help?
screen shot 2018-02-22 at 11 05 45

cc @darko-mijic @DominikGuzei

domenkozar: @nikolaglumac Fixed - sorry I missed this :(

deepfire:
Rebased, to elucidate the changeset:
```
deepfire@andromedae:~/daedalus$ git diff 70b2b8eb35f3b833e25d99acb3bbeb16e1879fa0 4e9e8cdf398fc176a2508d08156181a0b2d8cffa
```

deepfire:
Dropped a couple of co-cancelling commits:
```
deepfire@andromedae:~/daedalus$ git diff 77deac49c31108636bbc070dd312b350ad7da638 4e9e8cdf398fc176a2508d08156181a0b2d8cffa
deepfire@andromedae:~/daedalus$
```

rvl:
Rebased on ba037d30491803500cca05829e6dce7533b1b021 (latest [release/0.9.0](https://github.com/input-output-hk/daedalus/tree/release/0.9.0)).

PR #710, opened 2018-02-04
KR translate

KimUJeon:
I'm done the translation of file.
Can I translate terms of use too?


SebastienGllmt:
Looks like your PR is not setup correctly. Please modify it so that your branch and PR only contain the changes you've made.

PR #719, opened 2018-02-10
Fix/DDW-93 Poll local time difference every 1 hour when connected

darko-mijic:
This PR changes the logic for polling for local time difference:
- polling is started when connected
- polling is stopped when disconnected
- the polling interval is one hour

Changing the polling interval to one hour is a temporary measure needed since `/api/settings/time/difference` endpoint is consulting the NTP server every time and frequent polling is spamming. Later on, this endpoint will be fixed so the response from the NTP server will be cached. After that, we can poll more frequently.

### DIFF
--- a/app/stores/NetworkStatusStore.js
+++ b/app/stores/NetworkStatusStore.js
@@ -14,7 +14,7 @@ let cachedDifficulties = null;
 // Maximum number of out-of-sync blocks above which we consider to be out-of-sync
 const OUT_OF_SYNC_BLOCKS_LIMIT = 6;
 const SYNC_PROGRESS_INTERVAL = 2000;
-const TIME_DIFF_POLL_INTERVAL = 500;
+const TIME_DIFF_POLL_INTERVAL = 60 * 60 * 1000; // 60 minutes
 const ALLOWED_NETWORK_DIFFICULTY_STALL = 2 * 60 * 1000; // 2 minutes
 
 const STARTUP_STAGES = {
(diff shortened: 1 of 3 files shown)


DominikGuzei:
@darko-mijic so you would just add this and tell users in the FAQ to restart the application when they see this error?

darko-mijic:
Yes @DominikGuzei. Later or when the endpoint is fixed we will poll more frequently and a user will not be stuck on the screen with the error.

app/stores/NetworkStatusStore.js
```diff
@@ -277,4 +286,13 @@ export default class NetworkStatusStore extends Store {
     return Date.now() - this._startTime;
   }
 
+  _setupTimeDifferencePolling = () => {
```
darko-mijic: Not sure about the name for this reaction...

DominikGuzei: eg: `_pollTimeDifferenceWhenConnected`?

darko-mijic: That was one of my choices. It is also stopping the polling when disconnected.

darko-mijic: But that is more in line with our naming convention for reactions. Changing it.

darko-mijic:
Since this is a fix for the release I have targeted a release branch. That is per our workflow. As soon as the release is out of the door we must merge back to development and to master. 
Am I following the procedure correctly?

DominikGuzei:
There are two options and we should research the better one:
1. Merge release branch both into `master` and `develop` separately
2. Merge release branch only into `master` and then `master` back into `develop`

nikolaglumac:
@DominikGuzei I like the second option 🎉 
```
Merge release branch only into master and then master back into develop
```

nikolaglumac:
@darko-mijic @DominikGuzei tested against latest `release/1.1.0` cardano-sl branch and confirmed that it is working 🎉  - so I would say that https://github.com/input-output-hk/cardano-sl/pull/2486 did the trick 👍

PR #735, opened 2018-02-20
Fix/ddw 119 Fix wallet restore reactions configuration

nikolaglumac:
This PR improves async restore/import dialogs closing logic by enforcing the following:
```
Once restore/import is under way we need to either:
A) show the 'Add wallet' dialog (in case we don't have any wallets) or
B) just close the active dialog and unblock the UI
```
...which prevents the following situation:
screen shot 2018-02-19 at 22 54 38


DominikGuzei:
screenshot 2018-02-20 um 11 23 05

PR #738, opened 2018-02-21
[DDW-124] dont show validation error after clearing spending password field

DominikGuzei:
This PR fixes a minor UI issue on the receive screen when generating new address for a wallet that has a spending password set. After submit the password field was cleared and remained in an error state … this is now fixed by simplifying the setup and removing the validation debounce (which doesn't have any positive UX effect in this case anyway)


nikolaglumac:
@DominikGuzei merge at will ;)

PR #744, opened 2018-02-23
generate a daedalus installer with nix-bundle

cleverca22:
this generates an installer script contained inside a nix-bundle, which will then untar daedalus to a path under $HOME, and configure starting it with nix-user-chroot
it includes adding a daedalus binary to `~/bin/` which most distros auto-add to $PATH and a .desktop file at `~/.local/share/applications`, which at least xfce and ubuntu-gnome will add to the launcher menus


arianvp:
`~/bin` is not supported by any distro's afaik.   `~/.local/bin` is very common, however.

default.nix
```diff
@@ -1,16 +1,85 @@
-with (import (fetchTarball https://github.com/NixOS/nixpkgs/archive/fb235c98d839ae37a639695ad088d19ef8382608.tar.gz) {});
-# NOTE: when bumping nixpkgs, also update nixpkgs-src.json and .travis.yml
+{ cluster ? "mainnet" }:
+let
+  pkgs = import (import ./fetchNixpkgs.nix (builtins.fromJSON (builtins.readFile ./nixpkgs-src.json))) { config = {}; overlays = []; };
+  packages = self: {
+    inherit pkgs;
+    master_config = {
```
domenkozar: All of these should probably be top-level function args

default.nix
```diff
@@ -1,16 +1,85 @@
...
+      daedalus_darwin_url = "https://update-cardano-mainnet.iohk.io/Daedalus-installer-1.1.0.408.pkg";
+      cardano_rev = "eef095";
+      daedalus_hash = "1b0iy6c5nlsa2n2ylcc2kfm8ykkmdp6ncnx4lwwvc6f0662cf175";
```
domenkozar: This should default to parsing .git?

default.nix
```diff
@@ -1,16 +1,85 @@
...
+      cardano_hash = "151qjqswrcscr2afsc6am4figw09hyxr80nd3dv3c35dvp2xx4rp";
```
domenkozar: Depends on DEVOPS-673

default.nix
```diff
@@ -1,16 +1,85 @@
...
+              mv -vi $out/lib/librocksdb.a $static/lib/
+            '';
+          });
+        };
+      };
+    };
+    daedalus = self.callPackage ./linux.nix {};
+    rawapp = self.callPackage ./rawapp.nix {};
+    nix-bundle = import (pkgs.fetchFromGitHub {
+      owner = "input-output-hk";
```
domenkozar: Can we switch to upstream now?

default.nix
```diff
@@ -1,16 +1,85 @@
...
+      icon = "INSERT_ICON_PATH_HERE";
+    };
+    iconPath = ./installers/icons/1024x1024.png;
+    namespaceHelper = pkgs.writeScriptBin "namespaceHelper" ''
+      #!/usr/bin/env bash
 
-stdenv.mkDerivation {
-  name = "daedalus";
+      set -ex
+      cd ~/daedalus-install/
```
domenkozar: This could be a top-level configurable field and default to something like `~/.daedalus` - what do you think?

default.nix
```diff
@@ -1,16 +1,85 @@
...
+      mkdir -p etc
```
domenkozar: This should be inside `~/.daedalus`

fetch-nixpkgs.nix
```diff
@@ -1,23 +1,8 @@
...
-      $coreutils/mkdir $out
-      cd $out
-      $gzip -d < $src | $tar -x --strip-components=1
-    '')
-  ];
-  coreutils = builtins.storePath nixcfg.coreutils;
-  tar = builtins.storePath nixcfg.tar;
-  gzip = builtins.storePath nixcfg.gzip;
-}
+  # temporary hack until scripts/nix-shell.sh ceases to use -p
```
domenkozar: Can you elaborate what's going on there?

linux.nix
```diff
@@ -0,0 +1,124 @@
...
+  };
+  perClusterConfig = {
+    mainnet = {
+      key = "mainnet_wallet_macos64";
+    };
+    staging = {
+      key = "mainnet_dryrun_wallet_macos64";
+    };
+  };
+  launcherConfig = writeText "launcher-config.json" (builtins.toJSON {
```
domenkozar: Indeed, depends on DEVOPS-690 which is almost ready (at least the gist of it - configuration dhall-to-yaml)

nix-installer.nix
```diff
@@ -0,0 +1,172 @@
...
+    rmrf $UNPACK
+    ls -ltrh $UNPACK2
+    NIX_REMOTE=local?root=$UNPACK2 nix-store --load-db < $UNPACK2/nix-path-registration
+    NIX_REMOTE=local?root=$UNPACK2 nix-store --verify --check-contents
+    nix copy --no-check-sigs --from local?root=$UNPACK2 $(readlink $UNPACK2/firstGeneration)
+    export NIX_PROFILE=$DIR/nix/var/nix/profiles/profile
+    nix-env --set $(readlink $UNPACK2/firstGeneration)
+    rmrf $UNPACK2
+
+    post-install || true
```
domenkozar: Shouldn't this also launcher Daedalus as well? I'd imagine it would start up after installation.

nix-installer.nix
```diff
@@ -0,0 +1,172 @@
+{ installationSlug ? "nix-install", installedPackages, postInstall ? null }:
```
domenkozar: (nitpick): these files should go somewhere inside `installers` like `installer/nix`

nix-installer.nix
```diff
@@ -0,0 +1,172 @@
...
+let
+  pkgs = import (import ./fetchNixpkgs.nix (builtins.fromJSON (builtins.readFile ./nixpkgs-src.json))) { config = {}; overlays = []; };
+  nix-bundle = import (pkgs.fetchFromGitHub {
```
domenkozar: This is imported at two places

nix-installer.nix
```diff
@@ -0,0 +1,172 @@
...
+    repo = "nix-bundle";
+    rev = "630e89d1d16083";
+    sha256 = "1s9vzlsfxd2ym8jzv2p64j6jlwr9cmir45mb12yzzjr4dc91xk8x";
+  }) { nixpkgs = pkgs; };
+  installerBundle = nix-bundle.nix-bootstrap {
+    target = "${installer}";
+    run = "/bin/installer";
+    nixUserChrootFlags = "-c -m /home:/home -p HOME";
+  };
+  nixSrc = pkgs.fetchFromGitHub {
```
domenkozar: This can be removed after nixpkgs bump to include Nix 2.0?

release.nix
```diff
@@ -0,0 +1,13 @@
+with import ./. {};
```
domenkozar: This will have to match up DEVOPS-690 to be built for multiple envs (staging, mainnet)

shell.nix
```diff
@@ -0,0 +1,16 @@
+with import (import ./fetchNixpkgs.nix (builtins.fromJSON (builtins.readFile ./nixpkgs-src.json))) {};
+# NOTE: when bumping nixpkgs, also update nixpkgs-src.json and .travis.yml
```
domenkozar: This comment can now go away :)

default.nix
cleverca22: the cardano rev controls which cardano it fetches, so there is no `.git` before that
and the daedalus_darwin_url is to fetch the pre-npm'd javascript, mayflower will be replacing that

fetch-nixpkgs.nix
cleverca22: `scripts/nix-shell.sh` needs to be able to `nix-build` the nixpkgs to fetch it and populate `NIX_PATH`
but the nix 2.0 `builtins.fetchTarball` doesnt return a derivation, just a string with a storepath, and `nix-build` doesnt like that

so i need a dummy derivation to make `nix-build` happy

once `nix-shell.sh` is redone to not use `-p`, then we dont need this anymore

nix-installer.nix
cleverca22: this is to handle a bug that is in the released version of 2.0, so we have to wait for that fix to hit the nixpkgs daedalus is pinned to

cleverca22:
replaced by #797

PR #746, opened 2018-02-26
[DEVOPS-643] Fix nix-shell dependencies

rvl:
This adds the full list of build dependencies required for building to the nix-shell.

There is also a change to make the `buildkite-agent` program available when running on nix-darwin.

I'm not sure I like the `nix-shell.sh` script. Perhaps if all build agents are configured with nix available and the pipelines specify `NIX_PATH`, then it wouldn't be required?

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -4,7 +4,12 @@ with (import (fetchTarball https://github.com/NixOS/nixpkgs/archive/fb235c98d839
 stdenv.mkDerivation {
   name = "daedalus";
 
-  buildInputs = [git python27 curl electron nodejs-6_x nodePackages.node-gyp nodePackages.node-pre-gyp ];
+  buildInputs = [
+    nix bash binutils coreutils curl
+    git python27 curl electron nodejs-6_x
+    nodePackages.node-gyp nodePackages.node-pre-gyp
+    gnumake
+  ];
 
   src = null;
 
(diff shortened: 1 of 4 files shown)


domenkozar:
@rvl I don't like it too much as well :) buildkit starts with fresh env, so we need to come up with a way to setup Nix env in pure way.

scripts/nix-shell.sh
```diff
@@ -20,4 +20,6 @@ export NIX_REMOTE=daemon
 export NIX_PATH="nixpkgs=$(${NIX_BUILD} fetch-nixpkgs.nix -o nixpkgs)"
 export NIX_BUILD_SHELL
 
-${NIX_SHELL} -p nix bash binutils coreutils curl "$@"
+exec ${NIX_SHELL} default.nix \
```
domenkozar: Nice! :)

default.nix
```diff
@@ -4,7 +4,12 @@ with (import (fetchTarball https://github.com/NixOS/nixpkgs/archive/fb235c98d839
...
 
-  buildInputs = [git python27 curl electron nodejs-6_x nodePackages.node-gyp nodePackages.node-pre-gyp ];
+  buildInputs = [
+    nix bash binutils coreutils curl
+    git python27 curl electron nodejs-6_x
+    nodePackages.node-gyp nodePackages.node-pre-gyp
+    gnumake
+  ];
 
   src = null;
```
cleverca22: one small nit-pick, this file only ever works with `nix-shell`, so shouldnt it be called `shell.nix`?
ive renamed it in one of my PR's

domenkozar: +1 - we can migrate that as part of linux installer :)

rvl: I agree with both comments.

PR #749, opened 2018-02-26
WIP [DEVOPS-690] Dhall-based multi-cluster configuration

deepfire:


domenkozar:
One possible improvement would be to make `environment` specific configuration a function of `OS` specific configuration, that way we can deduplicate them.

domenkozar:
Boolean conditionals won't work well as soon as you need to add another cluster (testnet). We need to be easily to copy an existing environment and slightly modify it. With current implementation we'll have to nest booleans.

config/mainnet.dhall
```diff
@@ -0,0 +1,5 @@
+{ keyType      = ""
```
domenkozar: This file should accept OS settings as input and then you can just do:

`configurationKey = mainnet_wallet_${os.updateSystemName}"

config/mainnet.dhall
```diff
@@ -0,0 +1,5 @@
...
+, reportServer = "http://report-server.cardano-mainnet.iohk.io:8080"
+, updateServer = "http://update.cardano-mainnet.iohk.io"
```
domenkozar: This url is now changed in existing launcher files

config/launcher-macos64.dhall
```diff
@@ -0,0 +1,40 @@
+\(perCluster: { keyType : Text, relays : Text, reportServer : Text, updateServer : Text })
+->
+{ key      = "mainnet_${perCluster.keyType}wallet_macos64"
+, pass     =
+  { reportServer        = perCluster.reportServer
+  , nodePath            = "./cardano-node"
+  , nodeDbPath          = "$HOME/Library/Application Support/Daedalus/DB-1.0"
+  , nodeLogPath         = "$HOME/Library/Application Support/Daedalus/Logs/cardano-node.log"
+  , nodeArgs =
```
domenkozar: `nodeArgs` can be in final `launcher.dhall` and then just OS/environment specific settings can be substituted.

domenkozar: This is mainly because right now, someone might add nodeArgs here and forget to update the other file.

os/environment specific settings can be a simple record that is at the very end substituted in `launcher.dhall`.

scripts/build-single-installer-win64.bat
```diff
@@ -0,0 +1,62 @@
...
+    @echo ### Generating configs:  %CLUSTER%
+    @echo ###
+    @echo ###  LAUNCHER=%LAUNCHER:"=%
+    @echo ###  TOPOLOGY=%TOPOLOGY:"=%
+    @echo ###
+    @echo ##############################################################################
+    del /f ..\installers\launcher-config.yaml 2>nul
+    del /f ..\installers\wallet-topology.yaml 2>nul
+
+    echo %LAUNCHER:"=% | call dhall-to-yaml > ..\installers\launcher-config.yaml
```
domenkozar: Why is this not part of haskell code? It could be there united for both platforms, not having to maintain .bat files

config/mainnet.dhall
deepfire: But we already have an OS-specific file, why would we want another one?

config/mainnet.dhall
deepfire: Yes, this PR's base has aged, need to adjust..

scripts/build-single-installer-win64.bat
deepfire: It's a good step, I agree.

config/mainnet.dhall
deepfire: Rebased on top of current `release/0.9.0`

scripts/build-installer-unix.sh
```diff
@@ -148,37 +148,46 @@ cd installers
...
-        echo "Installer was not made."
-        mkdir -p dist/${APP_NAME}
-    fi
-
-    if test -n "${test_install}"
-    then echo "$0:  --test-install passed, will test the installer for installability";
-         case ${OS_NAME} in
-                 darwin ) sudo installer -dumplog -verbose -target / -pkg "dist/Daedalus-installer-${DAEDALUS_VERSION}.pkg";;
-                 linux )  echo "WARNING: installation testing not implemented on Linux" >&2;; esac; fi
+    case ${OS_NAME} in
```
domenkozar: `make-installer` already conditionalizes per OS

deepfire: @domenkozar, so you suggest moving installer testing into the installer generator?

config/check.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+set -e
+
+CLUSTERS="mainnet staging"
```
domenkozar: This duplicates what clusters are configured at multiple places, creating no benefit for having a separate CI check if dhall is evaluated as part of CI anyway.

installers/MacInstaller.hs
```diff
@@ -55,24 +60,40 @@ travisJobIdFromEnv :: IO (Maybe String)
...
+      , clusterName = clusterName
       }
 
 main :: IO ()
 main = do
   hSetBuffering stdout NoBuffering
 
   cfg <- installerConfigFromEnv
+
+  let capitalize :: String -> String
```
domenkozar: You can just write `fmap toUpper` instead of defining `capitalize`

scripts/build-installer-unix.sh
domenkozar: Sure - why not it really just works for darwin.

installers/MacInstaller.hs
```diff
@@ -55,24 +60,40 @@ travisJobIdFromEnv :: IO (Maybe String)
...
+      capitalize [] = []
+      capitalize (x:xs) = [Data.Char.toUpper x] <> xs
+      cluster'    = fromMaybe (error "Unrecognised cluster name in DAEDALUS_CLUSTER: should be one of:  mainnet staging") $ read $ capitalize $ clusterName cfg
```
domenkozar: Why not just `read clusterName :: Cluster`

installers/Config.hs
```diff
@@ -0,0 +1,90 @@
...
+
+  Dhall.detailed $ do
+    let inText = dhallTopExpr (Data.Text.Lazy.pack configRoot) rConfig rOS rCluster
+    print inText
+    expr <- case Dhall.Parser.exprFromText (Directed "(stdin)" 0 0 0 0) inText of
+              Left  err  -> Control.Exception.throwIO err
+              Right expr -> return expr
+
+    expr' <- Dhall.Import.load expr
+    case Dhall.TypeCheck.typeOf expr' of
```
domenkozar: Pretty sure this is not needed? `dhallToJSON` will definitely type check

config/check.sh
deepfire: @domenkozar, this is for quick local checks, during development

installers/MacInstaller.hs
deepfire: Because `read` fails obscurely, and because the input strings are lowercase.

installers/Config.hs
deepfire: @domenkozar, it doesn't: https://github.com/dhall-lang/dhall-json/blob/master/src/Dhall/JSON.hs#L159

installers/MacInstaller.hs
domenkozar: Not sure I follow - you're using read anyway? I'm surprised this compiles since `fromMaybe :: a -> Maybe a -> a` and you're passing `read` output to `Maybe a`

config/check.sh
domenkozar: It shouldn't introduce duplicated list again - we've made this mistake too many times.

deepfire:
@domenkozar, moved to `installers/dhall`

installers/MacInstaller.hs
deepfire: it's been fixed to `readMaybe` in the new commits -- and refactored into a common location : -)

deepfire:
I also need to resolve conflicts with the base..

deepfire:
Ok, the installers appear to build OK.  Next steps are rebasing and re-testing.

domenkozar:
This should target `develop` branch

deepfire:
@domenkozar,
1. rebased on develop
2. macOS installer testing now part of `make-installer`

scripts/build-installer-unix.sh
deepfire: Done

installers/Types.hs
```diff
@@ -0,0 +1,66 @@
...
+  deriving (Bounded, Enum, Eq, Read, Show)
+
+data Request
+  = Request
+  { rOS      :: OS
+  , rCluster :: Cluster
+  , rConfig  :: Config
+  } deriving (Eq, Show)
+
+class (Bounded a, Eq a) => Flag a where
```
domenkozar: Let's get rid of this magic and rather provide a simple function if we need explicit `CI -> Bool`.

installers/Config.hs
```diff
@@ -0,0 +1,134 @@
...
+dhallTopExpr path Launcher os cluster = path <> "/launcher.dhall ( "<>path<>"/" <> lshow cluster <> ".dhall "<>path<>"/" <> lshow os <> ".dhall ) "<>path<>"/" <> lshow os <> ".dhall"
+dhallTopExpr path Topology os cluster = path <> "/topology.dhall ( "<>path<>"/" <> lshow cluster <> ".dhall "<>path<>"/" <> lshow os <> ".dhall )"
+
+generateConfig :: Request -> FilePath -> FilePath -> IO ()
+generateConfig Request{..} configRoot outFile = handle $ do
+  GHC.IO.Encoding.setLocaleEncoding GHC.IO.Encoding.utf8
+
+  Dhall.detailed $ do
+    let inText = LT.fromStrict $ dhallTopExpr (pack configRoot) rConfig rOS rCluster
+    print inText
```
domenkozar: This shouldn't be here right?

installers/Config.hs
domenkozar: `Dhall.Import.load` does call typeOf.

installers/Config.hs
```diff
@@ -0,0 +1,134 @@
+{-# LANGUAGE LambdaCase          #-}
```
domenkozar: This file should also use `NoImplicitPrelude` for consistency

installers/Config.hs
```diff
@@ -0,0 +1,134 @@
...
+  GHC.IO.Encoding.setLocaleEncoding GHC.IO.Encoding.utf8
+
+  Dhall.detailed $ do
+    let inText = LT.fromStrict $ dhallTopExpr (pack configRoot) rConfig rOS rCluster
+    print inText
+    expr <- case Dhall.Parser.exprFromText (Directed "(stdin)" 0 0 0 0) inText of
+              Left  err  -> Control.Exception.throwIO err
+              Right expr -> return expr
+
+    expr' <- Dhall.Import.load expr
```
domenkozar: You can just `fmap` this over Maybe in `expr`

installers/Config.hs
```diff
@@ -0,0 +1,134 @@
...
+    case Dhall.TypeCheck.typeOf expr' of
+      Left  err -> Control.Exception.throwIO err
+      Right _   -> return ()
+
+    json <- case Dhall.JSON.dhallToJSON expr' of
```
domenkozar: Same here, can fmap.

installers/Config.hs
```diff
@@ -0,0 +1,134 @@
...
+              Left err  -> Control.Exception.throwIO err
+              Right json -> return json
+
+    Data.ByteString.writeFile outFile (Data.Yaml.encode json)
+
+handle :: IO a -> IO a
```
domenkozar: Could you add a comment here why this was introduced?

installers/Types.hs
```diff
@@ -0,0 +1,66 @@
...
+  toBool = (== enabled)
+  fromBool :: Bool -> a
+  fromBool x = if x then minBound else maxBound
+  enabled, disabled :: a
+  enabled  = minBound
+  disabled = maxBound
+  opposite :: a -> a
+  opposite = fromBool . not . toBool
+
+lowerShowT :: Show a => a -> Text
```
domenkozar: you can use:

```
λ> :t Universum.show
Universum.show :: (IsString b, Show a) => a -> b
```

installers/Types.hs
deepfire: Respectfully, I can explain why this is needed and how this works,
  so you can see it is not magic.

  First, the why.

  The Flag type encodes type-safe, semantically enriched boolean values,
  that are not prone to the (somewhat) famous issue of Boolean Blindness.

  If you do not trust my judgement that this is a serious problem, please consider
  referring to these, different sources:

  - https://existentialtype.wordpress.com/2011/03/15/boolean-blindness/
  - https://github.com/quchen/articles/blob/master/algebraic-blindness.md

  Also, you once said yourself, that you're afraid of 'iohk-ops' having all these
  booleans, and the possibility of eventually confusing them in the code, leading
  to things like WipeNodeDBs.

  Indeed, that concern of yours has spurred me to implement the solution you see.

  Next, how it works, and why it is not magic.

  The Bounded type class provides minimum and maximum values for an ADT, whose
  constructors are all 0-ary.  This reference to minimum and maximum allows us to
  encode Bool values naturally and consistently, which gives rise to a type class.

  As you can see, the implementation is rather simple -- and I would argue that it
  serves to solve the stated problem (Boolean Blindness) fairly well.

deepfire: cc @rvl, @jmitchell

jmitchell: I share your concerns about boolean blindness, @deepfire. The use of `Bounded` is less obvious to me, though. From [the docs](https://hackage.haskell.org/package/base-4.9.0.0/docs/Prelude.html#t:Bounded):

> The Bounded class may be derived for any enumeration type; minBound is the first constructor listed in the data declaration and maxBound is the last.

The default definitions of `enabled` and `disabled` are the opposite of what I'd expect given that `minBound :: Bool` is `False`. By leaking `Bool` into the ADT the concerns about boolean blindness seep into our logic.

Somewhere we have to implement `Bool -> TestInstaller` because the value depends on the presence or absence of a flag option. As long as we isolate the boolean blindness to this one spot, it's a forgivable and obvious crime. There may be a more elegant way to handle that, but I would prefer we at least get rid of `Bounded`.

domenkozar: On top of Jake's arguments, most importantly: we strive to optimize code for reading, not writing. Code is written once and read many, many times in teams.

It took me a couple of minutes to understand how all of this works while having `TestInstaller -> Bool` used inside our functions would still solve boolean blindness and be understandable in a couple of seconds. 

I looked through the code and most of the instances are not used anyway, except in `TestInstaller`. I do not quite understand they would make sense anyway, since those ADTs have more than two items. It's also not obvious that adding a few term level at the beginning or end will change the behavior of the code. If one would implement a simple pattern match, you'd get a compile error - hence I consider such code magic.

Even `TestInstaller` doesn't need `-> Bool` because its only used once and since you don't pass it to a third party library you can do:

```
MacInstaller.hs:  when (oTestInstaller == TestInstaller) $ do
```

eliminating any need for Bounded instance.

Going an extra mile here would be to do proper naming of ADTs, so it's clear from the name what boolean value it implies: `IsTestInstaller = IsTestInstaller | IsNotTestInstaller`.

installers/Config.hs
deepfire: @domenkozar, are you referring to the import-from-URL edge case in  https://github.com/dhall-lang/dhall-haskell/blob/master/src/Dhall/Import.hs#L652 ?

domenkozar: Ah I see - you copied most of this code from https://github.com/dhall-lang/dhall-json/blob/master/dhall-to-json/Main.hs#L35-L74 - maybe we should submit a PR to dhall-to-json to be able to use it as a haskell library?

installers/Types.hs
deepfire: @domenkozar, @jmitchell, interestingly, I find that your suggestions collide somewhat, and so I'm at a certain impasse:
1. Domen suggests a `TestInstaller -> Bool` function, whereas
2. Jake suggests getting rid of `Bool` altogether
We can't do both, so we need some kind of consensus here : -)

The `Bool -> TestInstaller` is supposed to happen as early as possible -- hence the `flag` option parser, which is generic and allows us to avoid mentioning `Bool` in client code entirely.

Note, though, how `flag` depends on ability to _generically negate_ the value -- which depends on the `Flag` type class which defines this negation.  Please note that we can still hide the unsafe `Bool`-involving operations using module boundaries and exports.

@jmitchell, if I understand you correctly, you don't like the `Bounded` dependency of `Flag` (`Ord` is _not_ a dependency of `Flag`, though!) -- but do you have suggestions on how we could otherwise implement negation generically, without involving `Bounded`?
Note, that the `Bool -> Flag` injection needs not be exposed to client code -- we can hide it in an implementation module, if we want the extra safety.

@domenkozar, I'm completely on-board with you with regards to optimization for reading!

But also, do you think that optimization for reading is an argument against abstraction?

Especially abstraction that allows us to deal with the domain language, instead of `Bool` -- with increased safety?

installers/Config.hs
deepfire: @domenkozar, I found it helpful to see the input expression printed to stdout, to help ensuring things are as they are at key junctions -- since we're dealing with `sh` and `cmd.exe` here..

As an example: https://buildkite.com/input-output-hk/daedalus/builds/429#719c10ca-f6f5-4a4c-8300-492977753b64/22-509

installers/Config.hs
deepfire: @domenkozar, not really -- `Either` is functorial over its second type argument, so over `Right`, not `Left`.

installers/Config.hs
deepfire: @domenkozar, again, `Either` is functorial over its second type argument, so over `Right`, not `Left`.

installers/Types.hs
deepfire: Done.

installers/Config.hs
deepfire: Done.

installers/Config.hs
deepfire: Done.

installers/Config.hs
deepfire: @domenkozar, sure thing!

installers/Types.hs
domenkozar: We are consistent here with Jake.

Jake is saying we don't need `TestInstaller -> Bool` at all for the moment. I say the same here:

```
Even TestInstaller doesn't need -> Bool because its only used once and since you don't pass it to a third party library 
```

What I was saying about `TestInstaller -> Bool` is that if you DO need it, it won't create boolean blindness since you'll be converting to Bool just before you pass it to `base` package function that requires explicit `Bool`.

@deepfire yes, reading is harder for abstractions, especially leaky ones like encoding Booleans into order of ADTs.

@deepfire please read my previous comment, your implemention is less type safe due to being more implicit.

installers/Config.hs
deepfire: https://github.com/dhall-lang/dhall-json/pull/20 is WIP, pending feedback from Gabriel.

installers/Types.hs
deepfire: @domenkozar, oh, I don't mind hiding `TestInstaller -> Bool` at all -- they don't need to be anything above an implementation detail.

Done!

domenkozar: @deepfire you're not addressing my feedback, please re-read my comments.

deepfire: @domenkozar, I'm sorry that you feel that way -- I'm certainly trying to address your feedback.

Let's go through your previous message bit by bit:

> We are consistent here with Jake.
>
> Jake is saying we don't need TestInstaller -> Bool at all for the moment. I say the same here:
> ```
> Even TestInstaller doesn't need -> Bool because its only used once and since you don't pass it to a third party library 
> ```

The unsafe `TestInstaller -> Bool` is no longer dangerously exposed -- I don't think it's fair to say your feedback wasn't addressed on this point.

> What I was saying about `TestInstaller -> Bool` is that if you DO need it, it won't create boolean blindness since you'll be converting to Bool just before you pass it to base package function that requires explicit Bool. This means your function signatures will still have isTestInstaller and then reading the term-level code you can see it becomes a boolean with an explicity cast.

The new code doesn't need or even suggest using `TestInstaller -> Bool` in the client side, so this section doesn't apply, right?  Any time you need a `Bool`, you can just do `oTestInstaller == TestInstaller`, as was intended all the way.  I think we can agree on that.

So the second part of your previous comment was implicitly addressed.

> yes, reading is harder for abstractions, especially leaky ones like encoding Booleans into order of ADTs.
That simply means assigning interpretation, and it's done once -- where as usage happens multiple times -- so it's a trade off for letting `Bool` not to be used in the user code at all.

And the reason it's needed was already described in detail, previously -- which I'll repeat again, because you didn't seem to recognize this in our discussion:

The need for the `Bounded` instance above -- to enable the generic `flag` mechanism, that allows to completely eliminate the exposure to `Bool` in user code.

domenkozar: @deepfire let's come back to the original issue I raised up - there is no need for any of the bounded instances.

The only place in current code where any of ADTs that have bounded instances are converted from or to a Bool is:

```
flag :: Flag a => a -> ArgName -> Char -> Optional HelpMessage -> Parser a
flag effect long ch help = (\case
                               True  -> effect
                               False -> opposite effect) <$> switch long ch help
```

which if you want to keep it abstract but remove the bounded:

```
flag :: Flag a => (Bool -> a) -> ArgName -> Char -> Optional HelpMessage -> Parser a
flag f long ch help = f <$> switch long ch help

toTestInstaller True = IsTestInstaller
toTestInstaller False = IsNotTestInstaller
```

and then call it as `flag toTestInstaller ...`

installers/Config.hs
deepfire: ..merged!

installers/Types.hs
deepfire: Ok, let's do that.

jmitchell: @domenkozar's last comment demonstrates addresses my concerns about `Bounded`, and I support moving in that direction.

deepfire: Interestingly, this is isomorphic to adding an injection `Bool -> Flag a` method to the `Flag` type class.  @domenkozar, what do you think?  It would remove the need to explicitly pass the function, and to care/remember about its name.

domenkozar:
Actually `Flag` is not needed here at all anymore - things keeps getting simplified :)

installers/Types.hs
deepfire: cc @domenkozar

jmitchell: Boolean blindness is amplified when abstractions are built around `Bool`. Since there's only one use case for the `Flag` I suggest we keep the code direct, minimizing the number of places one must look to know what happens when the flag is or is not present.

We can add more abstraction later if there's a need.

jmitchell: ICYMI: https://github.com/input-output-hk/daedalus/pull/749#issuecomment-372459501

deepfire: > Boolean blindness is amplified when abstractions are built around Bool.

Which is equally true for both implementation strategies, be it:
```
instance Flag TestInstaller where
   fromBool True = IsTestInstaller
   fromBool False = IsNotTestInstaller
```
..or..
```
isTestInstallerFromBool :: Bool -> TestInstaller
isTestInstallerFromBool True = IsTestInstaller
isTestInstallerFromBool False = IsNotTestInstaller
```
right?

But I'll do as you suggest.

deepfire: Done.

installers/Config.hs
```diff
@@ -0,0 +1,123 @@
...
+  <*>                optText "output"              'o' "Installer output file"
+  <*> (optional   $
+      (PullReq   <$> optText "pull-request"        'r' "Pull request #"))
+  <*> flag testInstaller     "test-installer"      't' "Test installers after building"
+  <*> pure Buildkite -- NOTE: this is filled in by auto-detection
+
+

+
+lshow :: Show a => a -> Text
```
domenkozar: Use `Universum.show`

installers/MacInstaller.hs
```diff
@@ -135,18 +107,18 @@ makeInstaller cfg = do
...
+  writeLauncherFile dir
 
-  with (makeScriptsDir cfg) $ \scriptsDir -> do
+  with (makeScriptsDir opts) $ \scriptsDir -> do
     let
       pkgargs :: [ T.Text ]
       pkgargs =
            [ "--identifier"
-           , "org." <> appNameLowercase cfg <> ".pkg"
+           , "org.daedalus.pkg"
```
domenkozar: This is a regression for mantis installer

installers/MacInstaller.hs
```diff
@@ -16,101 +16,76 @@ module MacInstaller
...
--- or "false" if it’s not.
-pullRequestFromEnv :: IO (Maybe String)
-pullRequestFromEnv = liftM2 (<|>) (getPR "BUILDKITE_PULL_REQUEST") (getPR "TRAVIS_PULL_REQUEST")
-  where
-    getPR = fmap interpret . lookupEnv
-    interpret Nothing        = Nothing
-    interpret (Just "false") = Nothing
-    interpret (Just num)     = Just num
+import           Config
+import           Types
```
domenkozar: Why was InstallerConfig removed - this was effort from @cleverca22 for mantis installer generalization.

installers/MacInstaller.hs
```diff
@@ -16,101 +16,76 @@ module MacInstaller
...
+  echo "Generating configuration file:  wallet-topology.yaml"
+  generateConfig (Request Macos64 oCluster Topology) "./dhall" "wallet-topology.yaml"
+
+  echo "Packaging frontend"
+  shells "npm run package -- --icon installers/icons/256x256" mempty
+
+  tempInstaller <- makeInstaller opts appRoot
 
-  if shouldSign
+  let shouldSign = oPullReq == Nothing
```
domenkozar: This changed behaviour - previously it would sign on buildkite regardless if it was a PR or not.

installers/MacInstaller.hs
cleverca22: of note, the `predownloadChain` section is no longer of use, but we should keep the rest i think

installers/Types.hs
```diff
@@ -0,0 +1,91 @@
...
+testInstaller True  = TestInstaller
+testInstaller False = DontTestInstaller
+
+flag
+  :: (Bool -> a)           -- ^ Lift a value from Bool.
+  -> ArgName               -- ^ Long form of the switch.
+  -> Char                  -- ^ Short form of the switch.
+  -> Optional HelpMessage  -- ^ Help message.
+  -> Parser a
+flag lift long ch help = lift <$> switch long ch help
```
domenkozar: Is this function needed? It abstracts away `fmap` for one usage occurance.

installers/Config.hs
```diff
@@ -0,0 +1,123 @@
...
+import           Turtle                              (optional)
+import           Turtle.Options
+
+import           Prelude
+import           Types
+
+

+
+diagReadCaseInsensitive :: (Bounded a, Enum a, Read a, Show a) => String -> Maybe a
```
domenkozar: I don't know how to do this better but let's add an example comment here for motivation:

```
λ> data Foo = Bar | Baz deriving (Show, Enum, Bounded)
λ> let x = enumFromTo minBound maxBound :: [Foo]
λ> x
[Bar,Baz]
λ> fmap ((fmap toLower) . show) x
["bar","baz"]
```

installers/Types.hs
```diff
@@ -0,0 +1,91 @@
...
+
+

+
+lowerShowT :: Show a => a -> Text
+lowerShowT = toLower . Universum.show
```
domenkozar: Convention is to suffix transformers with `T` so this becomes an overloaded term, maybe end it with `Text`?

installers/MacInstaller.hs
deepfire: Done

installers/Config.hs
deepfire: Done

installers/Config.hs
deepfire: done

installers/Types.hs
deepfire: Done, renamed to `lshowText`, to keep overhead down

installers/MacInstaller.hs
deepfire: The idea is to use Options for that.  I'm discussing this with @cleverca22.

installers/MacInstaller.hs
deepfire: Do you think we should revert that to always sign on Buildkite?

domenkozar: This PR shouldn't change/touch this behavior since it's out of scope.

If it did, we should discuss it - and for now we sign all buildkite builds since noone except trusted IOHK members can trigger them.

deepfire: Done.

deepfire:
I wish you said it's going to a `Freer` reader -- like, for example https://github.com/lexi-lambda/freer-simple#readme : -)

domenkozar:
Why do you need a Freer Monad here?

deepfire:
@domenkozar, well, it was you who mentioned monadic composition -- and these days you have two ways of doing this:
1. The legacy monad transformer mechanism, and
2. The Free/Freer Monad approach.

domenkozar:
Right :) Do you consider this ready for another review?

deepfire:
@domenkozar, several things remain, but not sure if they belong to/block this PR:

1. the Dhall stack overrides should perhaps be upgraded to their Hackage release: https://github.com/input-output-hk/daedalus/pull/749/files#diff-532314a1aff1436f071a9ee037b6b269R5
2. The cluster list we build for needs.. better parametrisation story, as we discussed previously -- preferably in a single location: https://github.com/input-output-hk/daedalus/pull/749/files#diff-d5d7cd29d8d8a86db9a1d8a53c65bbd6R7

Also, before merging, the Windows build needs to be sorted out -- but this should be fairly self-contained and wouldn't affect review much.

domenkozar:
1. that is fine for now

2. yes - that must be done before mistakes happen.

deepfire:
@domenkozar, so, `installer-clusters.cfg` in the Daedalus root directory now governs what's built.

I think that was the last concern we had..

installers/dhall/check.sh
```diff
@@ -0,0 +1,14 @@
...
+
+set -e
+
+CLUSTERS="$(cat $(dirname $0)/../../installer-clusters.cfg | xargs echo -n)"
+OSES="linux macos64 win64"
+
+cd $(dirname $0)
+for os in ${OSES}
+do for cluster in ${CLUSTERS}
+   do echo "checking ${cluster}-${os}"
```
domenkozar: I still think this is a bad idea - you have now at two places specified how dhall is composed and you'll have to maintain both of them. Developer will change haskell file and dhall files, CI will fail and they'll have to update this file as well. So what's the benefit?

installers/cardano-installer.cabal
```diff
@@ -15,23 +15,33 @@ executable make-installer
...
                      , nsis
+                     , optparse-applicative
+                     , optparse-generic
+                     , optional-args
                      , split
                      , system-filepath
                      , temporary
                      , temporary
                      , text
+                     , trifecta
```
domenkozar: Do we really need trifecta now that we use dhall-json as a lib?

installers/dhall/check.sh
deepfire: @domenkozar, What would you propose as a better way for a developer to type-check all possible combinations of Dhall files?

installers/WindowsInstaller.hs
```diff
@@ -96,19 +101,21 @@ fileSubstString :: Text -> Text -> FilePath -> FilePath -> IO ()
...
 
     forM_ ["ca.conf", "server.conf", "client.conf"] $
         \f-> fileSubstString "OPENSSL_MD" "sha256" f (f <> ".windows")
 
     writeFile "daedalus.nsi" $ nsis $ do
         _ <- constantStr "Version" (str fullVersion)
+        _ <- constantStr "Cluster" (str $ unpack $ lshowText clusterName)
         name "Daedalus ($Version)"                  -- The name of the installer
-        outFile "daedalus-win64-$Version-installer.exe"           -- Where to produce the installer
+        outFile "daedalus-win64-$Version-$Cluster-installer.exe"           -- Where to produce the installer
```
domenkozar: This might break installer uploading logic? /cc @rvl @cleverca22

installers/cardano-installer.cabal
deepfire: Right, good catch! : -)

installers/WindowsInstaller.hs
deepfire: Not for Appveyor, evidently.  But other things, like `iohk-ops` -- sure..

installers/dhall/check.sh
deepfire: Initially, the thing was intended for local development, but then someone (maybe you, even) suggested it should be run as part of CI.

Maybe we could backtrack on the CI run, and keep if for local development (where the person doing the development knows what they do..)

installers/WindowsInstaller.hs
domenkozar: So that has to be fixed as well (until we come up with a way to test it on CI)

installers/dhall/check.sh
domenkozar: I fail to see how it's useful for local development, now the whole logic lies in haskell and has nothing to do with shell anymore?

scripts/build-installer-win64.bat
```diff
@@ -2,19 +2,27 @@ rem DEPENDENCIES:
...
+@echo ###
+@echo ### Will build clusters:  %CLUSTERS%
+@echo ###
+@echo ##############################################################################
 
 set MIN_CARDANO_BYTES=20000000
 set LIBRESSL_VERSION=2.5.3
 set CURL_VERSION=7.54.0
-set CARDANO_BRANCH_DEFAULT=cardano-sl-1.0
+set CARDANO_BRANCH_DEFAULT=release/1.1.0
```
domenkozar: Note that this will conflict with https://github.com/input-output-hk/daedalus/pull/789/files so this change isn't needed.

deepfire: @domenkozar, that creates a kind of a problem -- you cannot really test the PR without that change -- and you cannot (auto-) merge with it included.  Heh : -)

installers/WindowsInstaller.hs
deepfire: How do you propose we test these changes together?

installers/dhall/check.sh
deepfire: @domenkozar, no-no, I mean testing the changes to the Dhall files themselves.

scripts/build-installer-win64.bat
deepfire: We could go with a manual conflict resolution merge commit pushed to this branch, by the way.  That'll probably close the PR automatically.

domenkozar: You can always supply this as argument to the script, if I read it correctly?

installers/WindowsInstaller.hs
domenkozar: create a development S3 bucket and point uploading scripts there :)

installers/dhall/check.sh
domenkozar: I suggest not increasing the scope of this PR, but you can do that if you export haskell functions that do dhall part and them use them through ghci and module reloading.

scripts/build-installer-unix.sh
deepfire: @domenkozar, it's still referenced -- for the Cardano download.

scripts/build-installer-win64.bat
deepfire: Sure, but punting the actual value to a different place won't make the conflict go away, which was our concern, FWIW.

installers/WindowsInstaller.hs
deepfire: It was agreed we should handle this in a separate issue.

installers/dhall/check.sh
deepfire: Right, that's a bit heavy-weight. The current shell approach provides a valuable balance, I believe.

deepfire:
@domenkozar, anything that blocks this from progressing?

deepfire:
@domenkozar, rebase done, your review pending.

installers/WindowsInstaller.hs
jmitchell: Meanwhile, let's not make this change. After we have an issue to track the necessary changes across repos to do this we can make focused PRs just for that.

installers/dhall/check.sh
jmitchell: I would prefer to keep the code DRY if possible, for exactly the reason Domen raised in the initial comment.

Short of that, we should exclude this from CI and add a bold comment near the top of this file to say:

1. it's for illustration purposes only
2. the actual logic for composing the Dhall files is in Haskell

deepfire:
> Good work on such a big change.

Thank you!

> ## Clarity

> Thing that concerns me is that it seems more complicated than it was before -- and less easy to understand because there's another layer. There's lots of stuff added but not much removed.

I've added a README section covering the general area of installer-time generation of runtime YAML configuration files.

> How will daedalus developers start the wallet if the yaml files are missing? I don't see any readme about that.

I'm not 100% positive on how much of a problem that is -- since they can always use the `scripts/build-installer-*` build scripts.

As a longer-term solution we could factor the  config generation as a subcommand of `make-installer`.

In the near term we probably want to balance the risks an absence of this extension introduces against the blocking of dependent PR's.

> ## Security

> Please do remove the command line argument for certificate password. Use the environment variable set by CI and document this variable in the command usage help text.

Done.

> ## Haskell

> In Haskell, why use fully qualified names where the conventional module abbreviations are fine?

It's a matter of preference -- the Dhall upstream code base uses that.  I have pushed a style normalisation commit.

> I would recommend using more of Turtle, not less because it's really good for these kind of scripts.

I generally agree, absolutely.  We've discussed this a lot : -)

> Instead of functions like `unsafeTextToLine` and `lshowText`, use Turtle's `printf` and nice formatting functions.

`lshow` does a transformation that I'm not sure Turtle provides.  Do agree on the rest.

> `Types.hs`: what is the `Request` type? It needs comments. `DeployParams` might be a better name.

Yes, clarified, but opted for `ConfigRequest`, since that's essentially what it is -- parameters that drive the configuration generator.

> `WindowsInstaller.hs`: when implementing new functionality such as invoking npm, put it into new functions. Lots of small functions are easier to understand and work with rather than one snaking main.

Done.

> ## Config file generation

> The `dhall` files - where are the comments?

Good point.  It's a lot of work to properly document all the parameters that drive `cardano-launcher`. It should be noted that the original configuration files don't have that documentation either.

Do you have a suggestion for a shorter-term solution, or you believe this is a blocking issue?

If so, I'll start documenting all of the options. that factor into the output configuration files.

> When a yaml file is generated from a dhall file, this tells me a build tool should be used, not a custom program/script. nix-build would be good for this.

I generally agree, yes. Question of short/long term balance is open, once more.

> It would be better to output yaml files and everything else to a clean temp directory, rather than mixing in with the code.

This is a question of rethinking of how the entire Daedalus build is currently organised, I believe.
Again, short v. longer term balance.

> `installer-clusters.cfg` - what is this file and why? It needs a comment at the top.

The way it is used precludes inline documentation -- I've added that on top of the README, instead.

> ## Building the installer

> `build-installer-unix.sh` - this file should be shrinking, not growing with added logic. The `--clusters` argument should be handled in make-installer.

> `installer-dev-mode.bat` - this file shouldn't be needed. Testing on windows should be no more complicated than `stack exec make-installer`

> `installers/default.nix` - note: I also have changes to this file in devops-673 and will be moving dhall and djall-json into separate files generated by cabal2nix.


> ## Summary

> It's a good change on the whole but legibility needs to be improved in my opinion.

installers/WindowsInstaller.hs
domenkozar: Do we have an issue for this @deepfire ?

installers/dhall/check.sh
domenkozar: This is still not addressed.

deepfire: Ok, I guess I'm confused -- Domen, this was never a part of CI -- now I'm not sure where the idea originated..

Jake, the documentation was added: https://github.com/input-output-hk/daedalus/blob/devops-690-dhall/installers/README.md#generation-of-runtime-configuration-files

installers/WindowsInstaller.hs
deepfire: https://iohk.myjetbrains.com/youtrack/issue/DEVOPS-742, @domenkozar

installers/dhall/check.sh
domenkozar: Following wasn't addressed:

> Short of that, we should exclude this from CI and add a bold comment near the top of this file to say:

> it's for illustration purposes only
> the actual logic for composing the Dhall files is in Haskell

Rodney pointed out that developers now don't have a simple way to generate .yaml files - if we created a separate executable calling `generateConfig` as proposed by Jake to keep things DRY. 

I don't understand why prefer bash script with 14 lines that duplicates behavior is better than 10 lines of haskell just exposing your existing code as something one can run.

deepfire: Ok, will do.

deepfire:
@domenkozar, @nikolaglumac, what would be our plan for getting this into the `master` branch?

Should we leave it as-is, or should I create a rebased PR?

nikolaglumac:
@deepfire we should follow the git-flow approach. I think this should be merged into master with the next release. Do you agree @domenkozar? Is there some reason why we should merge this to master now?

domenkozar:
No reason to treat this as a hot fix.

PR #766, opened 2018-03-01
Revert "[CSL-2307] Do not redirect cardano-node's logs to a file"

georgeee:
Reverts input-output-hk/daedalus#729


installers/launcher-config-mac.yaml
```diff
@@ -22,6 +22,7 @@ nodeArgs:
 - ./wallet-topology.yaml
 nodeDbPath: $HOME/Library/Application Support/Daedalus/DB-1.0
 nodeLogConfig: log-config-prod.yaml
+nodeLogPath: $HOME/Library/Application Support/Daedalus/Logs/cardano-node.log
```
gromakovsky: Actually it's enough to revert it only on Windows. Mac version works fine.

georgeee: Let's be conservative

PR #772, opened 2018-03-03
Update DEFAULT_DAEDALUS_BRANCH for osx

jonathancross:
Not sure if `master` is the best option, but it is better than `cardano-sl-0.4`.
Please let me know if you'd prefer a different branch as default.
PS: I also made the script executable (`u+x`).


domenkozar:
Thanks! this will be removed in https://github.com/input-output-hk/daedalus/pull/789

PR #789, opened 2018-03-14
[DEVOPS-673] Pin cardano-sl to git revision instead of branch

rvl:
With input-output-hk/cardano-sl#2671, there is a `daedalus-bridge` which provides cardano-sl programs and config files.

This change integrates that cardano-sl built by nix into the installer build, according to a specific git revision, rather than downloading the latest build from a branch.

### DIFF
--- a/cardano-sl-src.json
+++ b/cardano-sl-src.json
@@ -1,0 +1,6 @@
+{
+    "url": "https://github.com/input-output-hk/cardano-sl",
+    "fetchSubmodules": "true",
+    "rev": "1bba2fd0183f575752f4529182205e04074db336",
+    "sha256": "1qfi307x7nfp42ldb6sjx30mk7b0zfvd4b48pmyd9g465f7sp420"
+}
--- a/default.nix
+++ b/default.nix
@@ -25,6 +25,10 @@ in stdenv.mkDerivation {
     gnumake
   ];
 
+  passthru = {
+    inherit (cardano-sl-pkgs) daedalus-bridge;
+  };
+
   src = null;
 
 }
--- a/default.nix
+++ b/default.nix
@@ -1,7 +1,21 @@
-with (import (fetchTarball https://github.com/NixOS/nixpkgs/archive/fb235c98d839ae37a639695ad088d19ef8382608.tar.gz) {});
-# NOTE: when bumping nixpkgs, also update nixpkgs-src.json
+let
+  localLib = import ./lib.nix;
+in
+{ system ? builtins.currentSystem
+, config ? {}
+, pkgs ? (import (localLib.fetchNixPkgs) { inherit system config; })
+}:
 
-stdenv.mkDerivation {
+with pkgs;
+
+let
+  # we allow on purpose for cardano-sl to have it's own nixpkgs to avoid rebuilds
+  cardano-sl-src = builtins.fromJSON (builtins.readFile ./cardano-sl-src.json);
+  cardano-sl-pkgs = import (pkgs.fetchgit cardano-sl-src) {
+    gitrev = cardano-sl-src.rev;
+  };
+
+in stdenv.mkDerivation {
   name = "daedalus";
 
   buildInputs = [
--- a/installers/AppVeyor.hs
+++ b/installers/AppVeyor.hs
@@ -1,0 +1,160 @@
+{-# LANGUAGE NoImplicitPrelude, OverloadedStrings, LambdaCase, RecordWildCards, DeriveDataTypeable, DeriveGeneric #-}
+
+module AppVeyor (downloadCardanoSL, AppVeyorError(..)) where
+
+import Universum hiding (get)
+import Data.Aeson
+import GitHub hiding (URL)
+import GitHub.Endpoints.Repos.Statuses
+import qualified GitHub as GH
+import qualified Data.ByteString.Lazy.Char8 as L8
+import qualified Data.Text as T
+import Network.URI
+import Data.Maybe (mapMaybe)
+import Network.Wreq
+import Data.Aeson.Lens
+import Lens.Micro
+import Turtle.Format (printf, (%), d, s, w, Format, makeFormat)
+import GHC.Generics
+
+-- | Gets CardanoSL.zip corresponding to the src json revision from AppVeyor CI
+downloadCardanoSL :: FilePath -> IO L8.ByteString
+downloadCardanoSL srcJson = do
+  src@CardanoSource{..} <- getCardanoRev srcJson
+  printf ("Fetching GitHub CI status for commit "%n%" of "%n%"/"%n%"\n") srcRev srcOwner srcRepo
+  buildURL <- appVeyorURL src
+  printf ("Build URL is "%s%"\n") (getUrl buildURL)
+  jobs <- appVeyorJobs buildURL
+  printf ("Build has "%d%" job(s), status "%ss%".\n") (length jobs)
+    (map (show . jobStatus) jobs)
+  arts <- appVeyorArtifacts $ map jobId jobs
+  printf ("Artifacts are: "%ss%"\n") (map fst arts)
+  case findAppVeyorArtifact "CardanoSL.zip" arts of
+    Just bs -> printf "Done\n" >> pure bs
+    Nothing -> throwM (MissingArtifactsError $ getUrl buildURL)
+
+n :: Format r (Name a -> r)
+n = makeFormat untagName
+
+ss :: Format r ([Text] -> r)
+ss = makeFormat (T.intercalate ",")
+
+----------------------------------------------------------------------------
+
+-- | Given a build status URL, query AppVeyor for job IDs.
+appVeyorJobs :: GH.URL -> IO [Job]
+appVeyorJobs buildURL = do
+  let Just req = buildAPI buildURL
+  r <- asValue =<< get (show req)
+  pure (r ^.. responseBody . buildJobs)
+
+-- | Download all artifacts for all job ids.
+appVeyorArtifacts :: [Text] -> IO [(Text, L8.ByteString)]
+appVeyorArtifacts = fmap concat . mapM artifactForJobId
+
+-- | Filter the artifact we are interested in.
+findAppVeyorArtifact :: Text -> [(Text, L8.ByteString)] -> Maybe L8.ByteString
+findAppVeyorArtifact name arts = head [bs | (name', bs) <- arts, name' == name]
+
+-- | Downloads all artifacts for a job.
+-- Returns the filename and contents.
+artifactForJobId :: Text -> IO [(Text, L8.ByteString)]
+artifactForJobId jobId = do
+  let req = jobArtifactsURI jobId
+  r <- asValue =<< get req
+  forM (r ^.. responseBody . artifactFilenames) $ \filename -> do
+    let req' = req ++ ('/':T.unpack filename)
+    printf ("Downloading "%w%" ... ") req'
+    r' <- get req'
+    printf "done\n"
+    pure (filename, r' ^. responseBody)
+
+-- | Looks into AppVeyor API build info json to get job IDs
+buildJobs :: Monoid r => Getting r Value Job
+buildJobs = key "build" . key "jobs" . _Array . traverse . _JSON
+
+-- | Looks into AppVeyor API job artifacts json to get the filenames
+artifactFilenames :: Monoid r => Getting r Value Text
+artifactFilenames = _Array . traverse . key "fileName" . _String
+
+-- | The AppVeyor API URL of an build, given its web URL.
+buildAPI :: GH.URL -> Maybe URI
+buildAPI = parseAbsoluteURI . T.unpack . T.replace "/project/" "/api/projects/" . getUrl
+
+-- | The AppVeyor API URL to list artifacts belonging to a job.
+jobArtifactsURI :: Text -> String
+jobArtifactsURI jobId = "https://ci.appveyor.com/api/buildJobs/" <> T.unpack jobId <> "/artifacts"
+
+-- | Use GitHub API to find AppVeyor build status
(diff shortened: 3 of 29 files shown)


domenkozar:
This also needs to use `cardano-sl-src.json` on windows

rvl:
The AppVeyor CI branch build passes but PR build fails because their merge results in two cabal files (this PR renames the installer cabal file).

default.nix
```diff
@@ -1,7 +1,21 @@
-with (import (fetchTarball https://github.com/NixOS/nixpkgs/archive/fb235c98d839ae37a639695ad088d19ef8382608.tar.gz) {});
-# NOTE: when bumping nixpkgs, also update nixpkgs-src.json
+let
```
cleverca22: can you rename this to `shell.nix`? the constant changes in here are causing merge conflicts in my devops-13 PR

rvl: OK fixed

lib.nix
```diff
@@ -0,0 +1,19 @@
+let
+  # Allow overriding pinned nixpkgs for debugging purposes via cardano_pkgs
+  fetchNixPkgs = let try = builtins.tryEval <cardano_pkgs>;
```
domenkozar: I think we can get rid of this conditional as it was never used. Also `maybeEnv` :)

rvl:
OK fixed. We will need to decide in which order to merge out current devops branches.

scripts/build-installer-win64.bat
```diff
@@ -69,51 +62,6 @@ call npm install
...
-@echo ###
-@echo ##############################################################################
-
-move   node_modules\daedalus-client-api\log-config-prod.yaml installers\log-config-prod.yaml
-move   node_modules\daedalus-client-api\cardano-node.exe     installers\
-move   node_modules\daedalus-client-api\cardano-launcher.exe installers\
-move   node_modules\daedalus-client-api\configuration.yaml installers\
-move   node_modules\daedalus-client-api\*genesis*.json installers\
-del /f node_modules\daedalus-client-api\*.exe
-
```
deepfire: This is pure happiness of a diff chunk! : -)

installers/MacInstaller.hs
```diff
@@ -82,20 +83,30 @@ makeInstaller opts@Options{..} appRoot = do
   echo "Preparing files ..."
   case oAPI of
     Cardano -> do
+      -- fixme: when DEVOPS-690 adds option parsing, use a
```
deepfire: It did!

installers/AppVeyor.hs
```diff
@@ -0,0 +1,160 @@
...
+statusFor' :: CardanoSource -> IO (Either GH.Error GH.CombinedStatus)
+statusFor' CardanoSource{..} = statusFor noAuth srcOwner srcRepo srcRev
+  where noAuth = BasicAuth "" ""
+
+data JobStatus = JobSuccess | JobFailed | JobRunning
+               | JobUnknown Text deriving (Show, Eq, Generic)
+
+data Job = Job { jobId :: Text
+               , jobStatus :: JobStatus
+               } deriving (Show, Eq, Generic)
```
deepfire: Do you think it might be worth moving type definitions into `Types.hs`?

We already have a `BuildJob` type there (yeah, kind of "fallout" from `DEVOPS-690` being merged first..): https://github.com/input-output-hk/daedalus/blob/develop/installers/Types.hs#L69

installers/AppVeyor.hs
```diff
@@ -0,0 +1,160 @@
...
+getCardanoRev :: FilePath -> IO CardanoSource
+getCardanoRev src = eitherDecode <$> L8.readFile src >>= \case
+  Right rev -> pure rev
+  Left err -> throwM (SourceJSONDecodeError err)
+
+data CardanoSource = CardanoSource
+  { srcOwner :: Name Owner
+  , srcRepo  :: Name Repo
+  , srcRev   :: Name Commit
+  } deriving (Show, Eq)
```
deepfire: Does this type need to be specific to Cardano -- is it actually a Github repo source, in disguise?

installers/AppVeyor.hs
deepfire: ..and this actually reminds me that we actually have a bunch of very similar pre-existing types in `iohk-ops`..

Maybe we should have a discussion (outside of this PR!) about having them merged together..

cardano-sl-src.json
```diff
@@ -0,0 +1,6 @@
+{
+    "url": "https://github.com/input-output-hk/cardano-sl",
+    "fetchSubmodules": "true",
+    "rev": "1bba2fd0183f575752f4529182205e04074db336",
+    "sha256": "1qfi307x7nfp42ldb6sjx30mk7b0zfvd4b48pmyd9g465f7sp420"
+}
```
deepfire: I wonder if we want to (eventually, not now!) have a Github repo that defines versions in a central manner -- so we don't have these source specs littered all around..

installers/AppVeyor.hs
```diff
@@ -0,0 +1,160 @@
...
+import Network.Wreq
+import Data.Aeson.Lens
+import Lens.Micro
+import Turtle.Format (printf, (%), d, s, w, Format, makeFormat)
+import GHC.Generics
+
+-- | Gets CardanoSL.zip corresponding to the src json revision from AppVeyor CI
+downloadCardanoSL :: FilePath -> IO L8.ByteString
+downloadCardanoSL srcJson = do
+  src@CardanoSource{..} <- getCardanoRev srcJson
```
deepfire: I think this begs to be factored into a function of type `SrcSpec -> IO L8.ByteString`, what do you think?

installers/AppVeyor.hs
```diff
@@ -0,0 +1,160 @@
...
+  printf ("Artifacts are: "%ss%"\n") (map fst arts)
+  case findAppVeyorArtifact "CardanoSL.zip" arts of
+    Just bs -> printf "Done\n" >> pure bs
+    Nothing -> throwM (MissingArtifactsError $ getUrl buildURL)
+
+n :: Format r (Name a -> r)
+n = makeFormat untagName
+
+ss :: Format r ([Text] -> r)
+ss = makeFormat (T.intercalate ",")
```
deepfire: I think it might be worth starting to collect these little utility functions together in some kind of a `Misc` file -- together with or as a replacement of https://github.com/input-output-hk/daedalus/blob/develop/installers/Types.hs#L82

installers/WindowsInstaller.hs
```diff
@@ -223,3 +230,22 @@ main opts@Options{..}  = do
...
+  bs <- downloadCardanoSL "../cardano-sl-src.json"
+  let opts = [Zip.OptDestination dst, Zip.OptVerbose]
+  Zip.extractFilesFromArchive opts (Zip.toArchive bs)
+
+printCardanoBuildInfo :: MonadIO io => FilePath -> io ()
+printCardanoBuildInfo dst = do
+  let buildInfo what f = do
+        let f' = decodeString (dst </> f)
+        e <- testfile f'
+        when e $ do
```
deepfire: I wonder if it's valid to have a subset of these files.. and if we should catch any such divergences early.

Just a thought to consider.

deepfire: If we decide not to, maybe we could lift a type out of this?

installers/AppVeyor.hs
```diff
@@ -0,0 +1,160 @@
...
+  parseJSON = withObject "fetchGit source" $ \o -> do
+    (owner, repo) <- parseGitHubURL <$> o .: "url"
+    CardanoSource owner repo <$> o .: "rev"
+
+instance FromJSON URI where
+  parseJSON = withText "Absolute URI" $ \u -> case parseAbsoluteURI (T.unpack u) of
+    Just uri -> pure uri
+    Nothing -> fail "Could not parse absolute URI"
+
+-- | Gets owner/repo from gitcom.com URL
```
deepfire: Seeing how we have `*parsec` in our transitive deps, maybe we could be somewhat defensive and define a proper `Parser` for this?

Again, just a thought to consider.

installers/AppVeyor.hs
```diff
@@ -0,0 +1,160 @@
...
+-- | Looks into AppVeyor API job artifacts json to get the filenames
+artifactFilenames :: Monoid r => Getting r Value Text
+artifactFilenames = _Array . traverse . key "fileName" . _String
+
+-- | The AppVeyor API URL of an build, given its web URL.
+buildAPI :: GH.URL -> Maybe URI
+buildAPI = parseAbsoluteURI . T.unpack . T.replace "/project/" "/api/projects/" . getUrl
+
+-- | The AppVeyor API URL to list artifacts belonging to a job.
+jobArtifactsURI :: Text -> String
```
deepfire: Should this be a `URI`?

installers/AppVeyor.hs
```diff
@@ -0,0 +1,160 @@
...
+jobArtifactsURI jobId = "https://ci.appveyor.com/api/buildJobs/" <> T.unpack jobId <> "/artifacts"
+
+-- | Use GitHub API to find AppVeyor build status
+appVeyorURL :: CardanoSource -> IO GH.URL
```
deepfire: I'm worried we're getting a proliferation of differently-typed URI values..  I don't have a particular suggestion here -- just a thought that we might want to come up with a kind of a generic approach to the problem.

installers/AppVeyor.hs
```diff
@@ -0,0 +1,160 @@
...
+  printf ("Fetching GitHub CI status for commit "%n%" of "%n%"/"%n%"\n") srcRev srcOwner srcRepo
+  buildURL <- appVeyorURL src
+  printf ("Build URL is "%s%"\n") (getUrl buildURL)
+  jobs <- appVeyorJobs buildURL
+  printf ("Build has "%d%" job(s), status "%ss%".\n") (length jobs)
+    (map (show . jobStatus) jobs)
+  arts <- appVeyorArtifacts $ map jobId jobs
+  printf ("Artifacts are: "%ss%"\n") (map fst arts)
+  case findAppVeyorArtifact "CardanoSL.zip" arts of
+    Just bs -> printf "Done\n" >> pure bs
```
deepfire: : -)

deepfire:
This is a very welcome change, in general, but I'm slightly worried by the Appveyor-specific types, which could be made CI-agnostic, instead.

As a separate goalpost we might consider factoring the generic types/ops with the `iohk-ops` codebase.

scripts/build-installer-win64.bat
rvl: There is more which can and should be removed from the `.bat` file.

installers/MacInstaller.hs
rvl: OK, I have added it as a command line argument.

installers/AppVeyor.hs
rvl: I already brought this up at the standup.

installers/AppVeyor.hs
rvl: What do you propose?

cardano-sl-src.json
rvl: Why would you need to do that?

installers/AppVeyor.hs
rvl: Why?

installers/AppVeyor.hs
rvl: `n` is specific to `GitHub.Name` type and `ss` hardcodes the separator string, so no.

installers/WindowsInstaller.hs
rvl: It's just the logging which was previously in the awful `.bat` file.

installers/AppVeyor.hs
rvl: Why?

installers/AppVeyor.hs
rvl: No, because of the types of `wreq`.

installers/AppVeyor.hs
rvl: Pretty much unavoidable if you have the URL type from GitHub which is just a newtype Text wrapper. I just went with the way which involved minimal URL parsing.

rvl:
Thanks @deepfire 

The `AppVeyor.hs` file is going to be deleted when we move to Buildkite. Domen pointed out that I should have just put the `CardanoSL.zip` on a S3 bucket with the git revision in the filename. That would have been simpler but it's done now.

You are right that we can move common code into a library within `iohk-ops`.

cardano-sl-src.json
deepfire: To establish a single source of truth, and ensure that all of our builds get the same inputs.

In the end it will translate to higher Hydra cache hit rate and less confusion.

installers/AppVeyor.hs
deepfire: We already have something like this: https://github.com/input-output-hk/iohk-ops/blob/develop/iohk/Nix.hs#L29, which is more generic in purpose.  We could use something similar here, without being tied to Cardano specifics.

installers/AppVeyor.hs
deepfire: Being defensive, as I said.

installers/AppVeyor.hs
deepfire: To have more orthogonal, reusable code, I guess.

installers/WindowsInstaller.hs
deepfire: An incremental improvement, right?

cardano-sl-src.json
domenkozar: In practice that won't work - we had situations where we deployed different version of cardano-sl to users than it's running in the network. Just to allow shipping hotfixes faster.

installers/AppVeyor.hs
domenkozar: I suggest we factor out this later on into a library to handle pinning in such style.

PR #797, opened 2018-03-16
[DEVOPS-13] linux installer and yarn2nix build

cleverca22:


default.nix
```diff
@@ -1,16 +1,98 @@
...
+let
+  defaults = {
+    master_config = {
+      # TODO DEVOPS-673
+      cardano_rev = "eef095";
+      cardano_hash = "151qjqswrcscr2afsc6am4figw09hyxr80nd3dv3c35dvp2xx4rp";
+    };
+    pkgs = import (import ./fetchNixpkgs.nix (builtins.fromJSON (builtins.readFile ./nixpkgs-src.json))) { config = {}; overlays = []; };
+  };
+in { cluster ? "mainnet", master_config ? defaults.master_config, pkgs ? defaults.pkgs, buildNr ? "nix" }:
```
domenkozar: btw, top-level function arguments are recursive you can include defaults here:

```
{ cluster ? "mainnet"
, master_config ? { ... }
, pkgs ? import (import ./fetchNixpkgs.nix (builtins.fromJSON (builtins.readFile ./nixpkgs-src.json))) { config = {}; overlays = []; }
, buildNr ? "nix" }:
```

source/main/utils/getRuntimeFolderPath.js
```diff
@@ -9,6 +9,11 @@ export default (platform, env, appName) => {
       return path.join(env.APPDATA, appName);
     }
     case 'linux': {
+      var DAEDALUS_DIR = env.DAEDALUS_DIR;
```
domenkozar: cc @nikolaglumac

source/main/webpack.config.js
```diff
@@ -22,6 +26,15 @@ module.exports = {
     __dirname: false,
     __filename: false
   },
+  plugins: [
```
domenkozar: cc @nikolaglumac

source/main/windows/main.js
```diff
@@ -37,7 +40,7 @@ export const createMainWindow = () => {
 
   window.loadURL(`file://${__dirname}/../renderer/index.html`);
   window.on('page-title-updated', event => { event.preventDefault(); });
-  window.setTitle(`Daedalus (${environment.DAEDALUS_VERSION || 'dev'})`);
+  window.setTitle(`Daedalus (${environment.build})`);
```
domenkozar: cc @nikolaglumac

default.nix
cleverca22: yeah, but that gets a bit messy with the nixpkgs stuff in there
the master_config is also going away with the changes to pin cardano

source/main/utils/getRuntimeFolderPath.js
cleverca22: without this change, daedalus logs go to an entirely different path from cardano logs, and then daedalus cant upload cardano logs when reporting issues

source/main/webpack.config.js
cleverca22: i believe this bug was introduced during the webpack3 changes, and it results in those variables not being embeded into the code on the main.js side, this fixes it

source/main/windows/main.js
cleverca22: webpack didnt detect this, because its indirectly accessing the env vars
but the environment object already has a dedicated field for this value, and default value logic
that resulted it in requiring a `$DAEDALUS_VERSION` at runtime to have any effect

nikolaglumac:
@domenkozar @cleverca22 please don't merge this PR until we review it first. Thanks!

source/main/webpack.config.js
```diff
@@ -20,8 +24,18 @@ module.exports = {
...
+    __filename: false,
   },
+  plugins: [
+    new webpack.DefinePlugin({
+      'process.env.API': JSON.stringify(process.env.API || 'ada'),
+      'process.env.NETWORK': JSON.stringify(process.env.NETWORK || 'development'),
+      'process.env.MOBX_DEV_TOOLS': process.env.MOBX_DEV_TOOLS || 0,
+      'process.env.DAEDALUS_VERSION': JSON.stringify(process.env.DAEDALUS_VERSION || 'dev'),
+      'process.env.REPORT_URL': JSON.stringify(reportUrl),
+      'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV || 'dev'),
```
nikolaglumac: @cleverca22 please change the fallback value from `dev` into `development`

source/renderer/webpack.config.js
```diff
@@ -72,7 +72,8 @@ module.exports = {
       'process.env.NETWORK': JSON.stringify(process.env.NETWORK || 'development'),
       'process.env.MOBX_DEV_TOOLS': process.env.MOBX_DEV_TOOLS || 0,
       'process.env.DAEDALUS_VERSION': JSON.stringify(process.env.DAEDALUS_VERSION || 'dev'),
-      'process.env.REPORT_URL': JSON.stringify(reportUrl)
+      'process.env.REPORT_URL': JSON.stringify(reportUrl),
+      'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV || 'dev'),
```
nikolaglumac: @cleverca22 please change the fallback value from `dev` into `development`

source/main/windows/main.js
```diff
@@ -2,13 +2,17 @@ import path from 'path';
 import { app, BrowserWindow, ipcMain, Menu } from 'electron';
 import environment from '../../common/environment';
 import ipcApi from '../ipc-api';
+import { runtimeFolderPath } from '../config';
 
 export const createMainWindow = () => {
+  const iconPath = runtimeFolderPath + '/icon.png';
```
nikolaglumac: @cleverca22 please use `path.join` here to keep it cross-platform compatible.

source/main/windows/main.js
```diff
@@ -2,13 +2,17 @@ import path from 'path';
...
+
   // Construct new BrowserWindow
   const window = new BrowserWindow({
     show: false,
     width: 1150,
-    height: 870
+    height: 870,
+    icon: iconPath,
```
nikolaglumac: @cleverca22 what is the purpose of this icon? It is fine to add it I guess, but I can not locate the `icon.png` on `macOS` install.
Also we have AboutWindow - not sure if we need this icon there too?

cleverca22: on linux, this is the only way to put an icon on the window
i can add some if statements to make it do nothing on other platforms

source/main/windows/main.js
nikolaglumac: @cleverca22 my suggestion is to use something like:
```
path.join(runtimeFolderPath, 'icon.png');
```

source/main/windows/main.js
nikolaglumac: We have checked in macOS and Windows and it looks as this icon import has no affect.
It could be a Linux specific icon so I don't think any "ifs" are necessary.

PR #803, opened 2018-03-21
Fix/ddw 196 Add About window icon for Linux build

nikolaglumac:
This PR adds missing Daedalus icon to the About window on Linux build.

Previously we were showing a generic icon:
screenshot_2018-03-20_18-50-18

### DIFF
--- a/source/main/windows/main.js
+++ b/source/main/windows/main.js
@@ -5,14 +5,14 @@ import ipcApi from '../ipc-api';
 import { runtimeFolderPath } from '../config';
 
 export const createMainWindow = () => {
-  var params = {
+  const params = {
     show: false,
     width: 1150,
     height: 870,
   };
   if (process.platform === 'linux') {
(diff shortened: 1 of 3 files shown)


source/main/windows/main.js
```diff
@@ -5,14 +5,14 @@ import ipcApi from '../ipc-api';
 import { runtimeFolderPath } from '../config';
 
 export const createMainWindow = () => {
-  var params = {
+  const params = {
```
cleverca22: i thought const would prevent further changes to the object?

nikolaglumac: @cleverca22 In JavaScript, `const` means that the identifier can’t be reassigned. (Not to be confused with immutable values. Unlike true immutable datatypes such as those produced by Immutable.js and Mori, a `const` object can have properties mutated.)

cleverca22: ahh

PR #804, opened 2018-03-21
Fix/ddw 194 fix missing icons in webpack fork branch

nikolaglumac:
This PR fixes an issue caused by the replacement of `raw-loader` with `svg-inline-loader` which strips out `height` and `width` svg file attributes. This was identified within wallet navigation as the icons were missing in the navigation tabs. Also, the `SVGInline` dependency is updated to the latest version and both `package-lock.js` and `yarn.lock` have been updated.

### Before
missing-icons

### After
screen shot 2018-03-21 at 12 28 38

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,12 +4,6 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "7zip": {
-      "version": "0.0.6",
-      "resolved": "https://registry.npmjs.org/7zip/-/7zip-0.0.6.tgz",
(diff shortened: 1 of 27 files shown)


DominikGuzei:
@nikolaglumac i don't see any change related to the webpack setup?

nikolaglumac:
@DominikGuzei correct! The easy fix was to revert the svg loader to `raw-loader` instead of `svg-inline-loader` BUT I decided it is better approach to fix the CSS issues and keep the new loader in place ;)

PR #808, opened 2018-03-23
also include package.json

cleverca22:


scripts/package.js
```diff
@@ -29,6 +29,7 @@ const DEFAULT_OPTS = {
     // Ignore anything but dist and tls folders
     '^/(?!dist).*($|/)',
     '^/(?!tls).*($|/)',
+    '^/package.json',
```
DominikGuzei: Unfortunately that wont work … since these patterns are excluding each other … it needs to be something like `/^(?!(\/dist|\/tls|\/package\.json)).*($|\/` but im running into other problems: when i correctly ignore all `node_modules` the packager has some troubles.

PR #823, opened 2018-04-02
DO NOT MERGE: staging installers for the 1.1.1/0.9.1 release

deepfire:


deepfire:
Got an incoherent build (different Cardano commits in Windows/OS X builds), rebuilding..

PR #829, opened 2018-04-03
Chore/ddw 146 "Attach logs" should be on by default for support requests

nikolaglumac:
This PR sets the "Attach logs" switch by default to true within the support dialog:

screen shot 2018-04-03 at 16 51 57

...it also fixed several flow-type issues.


Belei3:
I do not understand at all how it works, and what needs to be done, on normal wallets, downloaded the required password and all, even if there is a problem, it reinstalled and all the files were transferred, launched and there are no problems. Everything is working!! Programmers need to be at least. !!! I lost 505 ADA, there is NO MORE WISH I WANT TO DO THIS. Thanks and good luck.

nikolaglumac:
@Belei3 sorry to hear you are having issues. Can I ask you to use the send-bug-report feature from Daedalus UI? That way our support team will get all the information required to help you resolve the problem. Thanks!

PR #831, opened 2018-04-03
Chore/ddw 177 Update acceptance test README

nikolaglumac:
This PR updates both main Daedalus README file as well as features README file with the new acceptance tests setup information.

### DIFF
--- a/features/README.md
+++ b/features/README.md
@@ -1,35 +1,22 @@
-# Running Deadalus acceptance tests
+# Running Daedalus acceptance tests
 
 
-1. Make sure you have correct node/npm versions installed on your machine (node v6.x and npm v3.x)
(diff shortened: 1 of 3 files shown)


features/README.md
```diff
@@ -1,35 +1,22 @@
...
+5. Run Daedalus frontend tests:
+
+```bash
 $ cd daedalus/
 $ npm run test
 ```
 
-There is a total of 46 different acceptance tests.
-Whole test suite takes around 5 minutes to finish.
+There is a total of 60 different acceptance tests.
```
DominikGuzei: Not sure we should put concrete numbers here as this will often change in the future.

nikolaglumac: It's more of a show off thing I guess :)
But to be honest our total test cases count has not been changing that much...

PR #833, opened 2018-04-03
Chore/ddw 165 Daedalus logs should be in UTC instead of local time

nikolaglumac:
This PR updates the date in Daedalus logs so that it is always shown in UTC instead of local time.

Before (local time - UTC+2):
```
[2018-04-03 21:37:34.0478] [debug] Local difficulty changed: 1124
```

After (UTC time - no timezone info:
```
[2018-04-03T19:37:34.0478Z] [debug] Local difficulty changed: 1124
```


DominikGuzei:
Works great 👍 merging now!

PR #839, opened 2018-04-05
[CSL-2435] Use 'TextEncoder' to measure correct length of UTF-8 strings

KtorZ:
JavaScript length method (with ES6) actually counts unicode punycodes of the UTF-16 encoding of the given string. Apart from messing up with surrogate pairs (counting for instance 💩 as two characters because encoded with a pair or punycodes), it also leads to unexpected results when characters have a different UTF-8 and UTF-16 encoding. 

Since requests are made using UTF-8 encoding, the `Content-Length` of these requests should match the actual length when payloads are UTF-8 encoded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Changelog
 - Updated moment.js dependency to the latest version which fixes ReDOS vulnerability ([PR 782](https://github.com/input-output-hk/daedalus/pull/782))
 - Fixed bug causing only 5 transactions to be shown on the transaction list screen ([PR 778](https://github.com/input-output-hk/daedalus/pull/778))
 - Fixed broken copy & paste context menu actions on Windows ([PR 817](https://github.com/input-output-hk/daedalus/pull/817))
+- Fixed non-latin character in wallet name bug ([PR 839](https://github.com/input-output-hk/daedalus/pull/839))
 
 ### Chores
 
--- a/features/wallet-settings.feature
+++ b/features/wallet-settings.feature
@@ -59,6 +59,15 @@ Feature: Wallet Settings
     And I click outside "name" input field
     Then I should see new wallet name "first Edited"
 
+  Scenario: User renames Wallet to a name which includes non-latin characters
+    Given I am on the "first" wallet "settings" screen
+    And I click on "name" input field
+    And I enter new wallet name:
+    | name     |
+    | キュビズム |
+    And I click outside "name" input field
+    Then I should see new wallet name "キュビズム"
+
   Scenario: User changes Wallet assurance level
     Given I am on the "first" wallet "settings" screen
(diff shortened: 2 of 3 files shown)


nikolaglumac:
We should add acceptance tests which cover this issue and first run it against existing code which should fail after which we can run it against the fixed code to confirm that it works.

nikolaglumac:
@KtorZ @DominikGuzei @darko-mijic I have ported the work to a branch that targets `release/0.9.1` branch: https://github.com/input-output-hk/daedalus/pull/840
...there I have added acceptance test that covers this bug.
We need to merge both PRs. I will add acceptance tests to this branch as well.

KtorZ:
Awesome :D

nikolaglumac:
Acceptance tests added in this PR too ;)

cleverca22:
i think this change is encoding it twice, once with an unknown method (which happens to work right now)
then again with TextEncoder, purely to get the size
```> str = "キュビズ";
'キュビズ'
> Buffer(str)
<Buffer e3 82 ad e3 83 a5 e3 83 93 e3 82 ba>```
but with this, you encode it once, measure the raw length of the bytes, then give the http req the raw bytes

see also: https://github.com/nodejs/node/blob/b38c81cb449a822ab45e6caa210d070b91a59836/lib/_http_outgoing.js#L756-L774

nikolaglumac:
I can confirm that all acceptance tests are passing 👍

nikolaglumac:
@KtorZ @DominikGuzei @darko-mijic what do you guys think of @cleverca22's proposal?

nikolaglumac:
My manual testing has also proven that the fix works! 🎉

DominikGuzei:
If @cleverca22's proposal works the same way it would be cleaner IMO.

nikolaglumac:
I don't know if it works. I have tested the existing one.
Can you please try to implement it?
I would like to get this PR merged until 3PM today :(

KtorZ:
I'd definitely agree with @cleverca22 if we were in a Node.js env (see the original ticket actually). `Buffer` in node infers utf8 by default. It seems that browsers now (since ES6) consider strings are being utf16, hence the issue we have here. Node.js Buffers aren't accessible on browsers however which come with different alternatives like Int8Array.

What the text encoder does is simply to read the string as utf8 encoded and create a corresponding UInt8Array out of it.

nikolaglumac:
We will go with what we have at the moment due to time restraints.
We can continue the debate on this technical detail in the next sprint within this card which I have created: https://iohk.myjetbrains.com/youtrack/issue/DDW-218

nikolaglumac:
Thanks for all of the efforts @KtorZ @cleverca22 @DominikGuzei ! This is a great improvement 🎉

PR #840, opened 2018-04-05
[CSL-2435] Use 'TextEncoder' to measure correct length of UTF-8 strings

nikolaglumac:
JavaScript length method (with ES6) actually counts unicode punycodes of the UTF-16 encoding of the given string. Apart from messing up with surrogate pairs (counting for instance 💩 as two characters because encoded with a pair or punycodes), it also leads to unexpected results when characters have a different UTF-8 and UTF-16 encoding.

Since requests are made using UTF-8 encoding, the Content-Length of these requests should match the actual length when payloads are UTF-8 encoded.

This PR is a port of https://github.com/input-output-hk/daedalus/pull/839

### Acceptance tests results:

**Before the fix:**
screen shot 2018-04-05 at 11 59 19

**After the fix:**
screen shot 2018-04-05 at 12 00 07

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Changelog
 
 - Fixed bug causing only 5 transactions to be shown on the transaction list screen ([PR 778](https://github.com/input-output-hk/daedalus/pull/778))
 - Fixed broken copy & paste context menu actions on Windows ([PR 817](https://github.com/input-output-hk/daedalus/pull/817))
-
+- Fixed non-latin character in wallet name bug ([PR 840](https://github.com/input-output-hk/daedalus/pull/840))
 
 ## 0.9.0
 =======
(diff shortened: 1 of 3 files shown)


nikolaglumac:
I can confirm that all our acceptance tests are passing!

nikolaglumac:
My manual testing has also proven that the fix works! 🎉

nikolaglumac:
@DominikGuzei @darko-mijic we need to get this PR merged as soon as possible so that we can resume the release process...

DominikGuzei:
@nikolaglumac you forgot to push the acceptance tests implementation?

nikolaglumac:
No @DominikGuzei - I have added them in the first commit: https://github.com/input-output-hk/daedalus/pull/840/commits/521d30a566df128fc5e8db862b40555013c0eb80
(perhaps this was not the best idea in the world)...

DominikGuzei:
@nikolaglumac ah sorry … i just saw the feature file and thought there are some implementation steps missing - but you just reused existing functionality 👍

DominikGuzei:
@nikolaglumac i will merge this now as it works as intended. Please feel free to create an issue on YouTrack with the proposal by @cleverca22

nikolaglumac:
Yes that single test captures the issue (in the code without the fix)

PR #841, opened 2018-04-05
[scripts/i18n tool] New 'i18n_check.js' to find missing and unused translation entries

gubatron:
I created this to find out what translation entries have been added to the base english `en-US.json` file that might be missing in other translation files.

_If there's a tool for this purpose already please let me know and disregard this PR._

Here is a sample of how the tool has worked for me after being 186 commits behind master while I was working on the Spanish translation.

screen shot 2018-04-05 at 12 12 05 pm

Note: I'm not sure if `/scripts` is the right place to put this tool.

### DIFF
--- a/scripts/i18n_check.js
+++ b/scripts/i18n_check.js
@@ -1,0 +1,123 @@
+//
+// Requires: npm install -G jsonfile
+// Author: Angel Leon (@gubatron)
+// Created on Apr 5th, 2018
+//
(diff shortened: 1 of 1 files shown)


nikolaglumac:
Thank you for your contribution!
We already have a script for that - `npm run manage:translations`.
Also if you open the translation files you will notice that all untranslated keys have values starting with `!!!`.

PR #842, opened 2018-04-06
[DEVOPS-13] disable linux updating

cleverca22:


disassembler:
We'll keep this open in case major issues arise, but @cleverca22 is working to complete the update testing on DEVOPS-13 today.

PR #861, opened 2018-04-12
Chore/ddw 229 Tell user that only one wallet can be restored at the same time

nikolaglumac:
This PR prevents opening of restore and import dialogs if there is already one active restoration underway.

### TODO:
- [x] Inject missing Japanese translation

screen shot 2018-04-12 at 12 58 37


nikolaglumac:
@DominikGuzei translations are now in! Merging this one without waiting for the CI since this was just a minor change...

PR #871, opened 2018-04-18
Fix/ddw 237 Update paper wallet certificate generation to use more random bits 

DominikGuzei:
This PR changes how paper wallet certificates are generated after input from audit results. Salt now uses more bits for entropy and`rust-cardano-crypto` scramble API which now requires 8 bytes for the IV param (27-word mnemonic instead of 24).

After this change paper wallet certificate has 27 mnemonic words in total, 18 words for scrambled mnemonics and 9 words for the password.

screen shot 2018-04-18 at 16 12 00


DominikGuzei:
@darko-mijic we need to update the acceptance tests too!

nikolaglumac:
Great work guys!
Both manual testing and running acceptance tests is working great!

screen shot 2018-04-18 at 19 24 03

PR #874, opened 2018-04-19
[DEVOPS-59] [DEVOPS-771] Installers: Apply versioning scheme

rvl:
* Version is no longer specified in appveyor.yml or .buildkite/pipeline.yml.
    
* No more version logic in batch file/shell scripts.

* Daedalus version is found in package.json.

* Cardano version is found in
  - Mac: `version` file within `daedalus-bridge` derivation.
  - Windows: runs `cardano-node.exe --version`
  - Nix: `daedalus-bridge.version` attribute

* There is a common function to determine the installer filename.

* There is a little bit of code churn from changing from the base FilePath API to the Turtle FilePath API.

* Split the MacInstaller.makeInstaller function into smaller parts.

* Added some basic test cases of installer builder and config file generation.


scripts/osx-build-fresh-daedalus.sh
```diff
@@ -52,7 +52,5 @@ git clone ${URL}
 pushd daedalus
```
deepfire: I think these `*-build-fresh-daedalus.*` scripts can be dropped, really, as they're too specialized.

installers/Spec.hs
```diff
@@ -0,0 +1,113 @@
...
+main :: IO ()
+main = hspec $ do
+  describe "Utility functions" utilSpec
+  describe "MacInstaller build" macBuildSpec
+  describe "Config generation" configSpec
+
+macBuildSpec :: Spec
+macBuildSpec = do
+  describe "The whole thing" $ do
+    it "Runs through the whole installer build" $ runManaged $ do
```
disassembler: Installer build tests!

scripts/osx-build-fresh-daedalus.sh
rvl: OK, fixed.

PR #937, opened 2018-05-22
[R120-20] Remove open dir if it exists during windows installation

cleverca22:


installers/WindowsInstaller.hs
```diff
@@ -146,6 +146,8 @@ writeInstallerNSIS (Version fullVersion') clusterName = do
                 createDirectory "$APPDATA\\Daedalus\\Secrets-1.0"
                 createDirectory "$APPDATA\\Daedalus\\Logs"
                 createDirectory "$APPDATA\\Daedalus\\Logs\\pub"
+                iff_ (fileExists "$APPDATA\\Daedalus\\Wallet-1.0\\open\\*.*") $
+                    rmdir [] "$APPDATA\\Daedalus\\Wallet-1.0\\open"
```
deepfire: Looks like a good, surgically-positioned migration recipe.

PR #953, opened 2018-05-31
[DEVOPS-872] disable allow skip files and adds a launcher lock file

disassembler:
This PR disables allow skip files so users can't ignore errors if cardano-node is already running. It also adds a lockfile to cardano-launcher so multiple launchers can't be running at the same time.


rvl:
Could the installer fix and launcher fix be separate PRs so they can be reviewed separately?

I'm not convinced any launcher change is necessary in a hotfix release, although some locking in the launcher would be a good feature for a later release.

jmitchell:
@disassembler, this should be target `release/0.11.0`.

disassembler:
rebased off of release branch.

disassembler:
Tested on WIndows just now. The lockfile prevents the installer from proceeding if launcher is still running.

nikolaglumac:
@darko-mijic can you also check out the screenshot @disassembler shared?

PR #985, opened 2018-06-20
[DEVOPS-919] installers: provide BUILD_NUMBER and API_VERSION

rvl:
* Renames the variable `DAEDALUS_VERSION` to `BUILD_NUMBER` because that's what it contains.
* Makes sure that `BUILD_NUMBER` is set for the Windows and macOS builds.
* Adds new variable `API_VERSION` and sets to the cardano-sl version known at build time.
* Uses these new ENV variables in place of hardcoded translation key files versions

### DIFF
--- a/source/renderer/app/i18n/locales/en-US.json
+++ b/source/renderer/app/i18n/locales/en-US.json
@@ -40,8 +40,6 @@
   "cardano.node.update.notification.titleWithoutVersion": "Daedalus and Cardano node update is available",
   "environment.apiName.cardano": "Cardano",
   "environment.apiName.mantis": "Mantis",
-  "environment.apiVersion.cardano": "1.3.0",
-  "environment.apiVersion.mantis": "1.0 rc1",
   "environment.currency.ada": "ADA",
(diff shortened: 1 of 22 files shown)


nikolaglumac:
@rvl I have removed all hardcoded versions from the translation files and now we use ENV variables directly...

source/renderer/app/i18n/locales/en-US.json
```diff
@@ -376,4 +374,4 @@
   "wallet.transactions.no.transactions": "No transactions",
   "wallet.transactions.no.transactions.found": "No transactions found",
   "wallet.transactions.search.hint": "!!!Search transaction"
-}
+}
```
disassembler: why no newline at end of file?

nikolaglumac: That’s automatic output od the translation manager...

jmitchell:
The macOS build failed with `Error: Cannot find module 'webpack'`. That started with the first commit.

rvl:
OK fixed the build problem and tested on macOS now.

PR #991, opened 2018-06-22
[DEVOPS-398] Installers: Add config for public testnet

rvl:
Configures the testnet installer so that the wallet can connect to the public testnet.

This PR is for the `release/0.11.0` branch. Merging this change won't have any impact on the mainnet or staging builds, so may as well be done now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+## vNext
+
+### Features
+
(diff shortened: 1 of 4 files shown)


nikolaglumac:
@rvl I think we should not merge it into `release/0.11.0` to avoid any risk.
We'll retarget this PR into `release/0.11.1` (branch to be created next week).
cc @darko-mijic @DominikGuzei @jmitchell

PR #992, opened 2018-06-22
[DEVOPS-922] Different ports for mainnet/staging/testnet

rvl:
This change will be targeting release/0.11.0 or release/0.11.1.

The `WALLET_PORT` environment variable will be set at webpack build time based on the target network. It is available to use from the `environment.WALLET_PORT` global.

The cardano launcher will use the same port.

This will allow mainnet and testnet builds of Daedalus to run at the same time.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -25,7 +25,8 @@ const createElectronServer = (env, args = []) => {
     spawnOpt: {
       env: Object.assign({}, process.env, env),
       args,
-    }
+    },
+    port: process.env.ELECTRON_PORT
   });
 };
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -25,8 +25,7 @@ const createElectronServer = (env, args = []) => {
     spawnOpt: {
       env: Object.assign({}, process.env, env),
       args,
-    },
-    port: process.env.ELECTRON_PORT
+    }
   });
 };
 
--- a/source/renderer/webpack.config.js
+++ b/source/renderer/webpack.config.js
@@ -90,6 +90,7 @@ module.exports = {
       'process.env.MOBX_DEV_TOOLS': process.env.MOBX_DEV_TOOLS || 0,
       'process.env.BUILD_NUMBER': JSON.stringify(process.env.BUILD_NUMBER || 'dev'),
       'process.env.REPORT_URL': JSON.stringify(reportUrl),
+      'process.env.WALLET_PORT': JSON.stringify(process.env.WALLET_PORT || ''),
     }),
     new AutoDllPlugin({
       filename: 'vendor.dll.js',
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+## vNext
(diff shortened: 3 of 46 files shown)


source/renderer/webpack.config.js
```diff
@@ -90,6 +90,7 @@ module.exports = {
       'process.env.MOBX_DEV_TOOLS': process.env.MOBX_DEV_TOOLS || 0,
       'process.env.BUILD_NUMBER': JSON.stringify(process.env.BUILD_NUMBER || 'dev'),
       'process.env.REPORT_URL': JSON.stringify(reportUrl),
+      'process.env.WALLET_PORT': JSON.stringify(process.env.WALLET_PORT || ''),
```
nikolaglumac: @rvl here we'll need fallback port `8090` instead of empty string to keep Daedalus working in development env...

nikolaglumac:
Thanks for the PR @rvl!
I think we won't be merging it into `release/0.11.0` -> it should target `release/0.11.1` (hotfix!).
I will create `release/0.11.1` branch next week.
cc @darko-mijic @DominikGuzei @jmitchell

nikolaglumac:
(closed by mistake!)

source/renderer/webpack.config.js
cleverca22: environment.js has a default, and i think it will trigger when this part sets it to `''`?

nikolaglumac: That is a valid point. Let's keep it as is...

daniloprates:
`WALLET_PORT` is working fine, however the Electron server will fail. By using a different port multiple instances work well. I have created an `ELECTRON_PORT` environment variable. Not sure if it's the best naming.

image

nikolaglumac:
@rvl I have rebased this PR to target `release/0.11.1`
cc @darko-mijic @jmitchell

gulpfile.js
```diff
@@ -25,7 +25,8 @@ const createElectronServer = (env, args = []) => {
     spawnOpt: {
       env: Object.assign({}, process.env, env),
       args,
-    }
+    },
+    port: process.env.ELECTRON_PORT
```
nikolaglumac: @daniloprates this is not what we are aiming for :(
It is my fault as I did not explain the task good enough...
It is the `WALLET_PORT` that we need and not `ELECTRON_PORT`.
Backend will be running on different ports for different networks. Until now we always used port `8090` to contact it. This has to be changed so that we use this variable: https://github.com/input-output-hk/daedalus/blob/devops-922-testnet-parallel/source/common/environment.js#L21 in all of our API calls (for Cardano only!). This means that we need to replace hardcoded `8090` port from all of the places and use the one from environment file.
This is one of the places where we need to do that: https://github.com/input-output-hk/daedalus/blob/devops-922-testnet-parallel/source/renderer/app/api/ada/adaTxFee.js#L22
(there are many more)
Thanks for taking care of this! :)

daniloprates: Sorry for the misunderstanding 👎Now it's done 👍

PR #993, opened 2018-06-22
[DEVOPS-919] installers: fix missing backend in filenames

rvl:
I forgot to update the installer filenames to add the backend name which was removed from the get version function.

### DIFF
--- a/installers/common/Types.hs
+++ b/installers/common/Types.hs
@@ -10,6 +10,7 @@ module Types
   (  -- * Atomic types
     OS(..)
   , Cluster(..)
+  , Backend(..)
   , Config(..), configFilename
   , ConfigRequest(..)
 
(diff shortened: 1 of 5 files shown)


rvl:
Oh bugger, something's wrong with the windows installer filenames.

rvl:
Fixed.

installers/common/Types.hs
```diff
@@ -57,6 +58,11 @@ data Cluster
   | Testnet
   deriving (Bounded, Enum, Eq, Read, Show)
 
+data Backend
+  = Cardano { cardanoDaedalusBridge :: FilePath }
```
jmitchell: I realize this was just moved from one place to another, but it's an anti-pattern to mix sum and record types like this. It results in partial functions. For instance, given a type of `Backend` you can now call `cardanoDaedalusBridge`, even when the value is `Mantis`.

One solution is to make a separate type for the record and use it as a parameter of the `Cardano` constructor.

jmitchell: Given the simplicity of the constructor here and the fact that `cardanoDaedalusBridge` isn't used anywhere I recommend simply changing this to `Cardano FilePath`.

rvl: Yeah I wrote the original, fully aware of the anti-pattern. Forgive my sins. I used the record type just to label what the parameter is, not to actually be called. Pushed a fix.

PR #1006, opened 2018-07-10
[DEVOPS-487] AppImage build

rvl:
Works on Ubuntu 18.04 and NixOS (through `appimage-run`).

## How it works

- All shared libraries required for running `cardano-node`, etc. are copied to `./usr/lib` inside the AppImage. Nix store RPATHs are removed using `patchelf`.
- Electron binary distribution is used to run electron. This does not share libraries with cardano-node. Electron binary distribution relies on many libraries provided by the underlying OS. This is the AppImage way.

## Advantages

- Simpler for users, doesn't create directories in unexpected places.
- Simpler for developers because the AppImage runtime is provided by someone else.
- Uses FUSE and doesn't require user namespaces.
- It's simple to uninstall the application.

## How are updates handled?

The `$APPIMAGE` variable points to the AppImage file location and is used by the update script to replace the old AppImage file with a new one.

This policy could be changed. For example, it could put the new version beside the old version instead of replacing it. It could also rename AppImage file after updating.

There is currently a problem with updating. The design of `cardano-launcher` means that it is not able to restart itself. It can only restart the node and frontend processes. `cardano-launcher` will probably need a tweak to work with AppImage.

## How are `.desktop` files handled?

Desktop integration is handled the same way as for other AppImages. It's optional. Users can install `appimaged`.

## Network setting not baked in

In this branch, the `NETWORK` and `REPORT_URL` variables aren't baked in so that the webpack build can be shared. These variables are set by the launcher script.

## Plain nixpkgs derivation

This will install a multi-network Daedalus:

    nix-env -f default.nix -iA daedalus

This will install a specific network Daedalus including desktop launcher:

    nix-env -f default.nix -iA testnet.daedalus-desktop

## Split up `make-installer`

The `make-installer` program is split into separate executables in order to provide a `generate-launcher-config` command. The program is split and refactored with future nixification of macOS and Windows installers in mind.

## Source filtering

Nixpkgs `lib.cleanSourceWith` is used with filters in the local `lib.nix` file.

## Re-organised `default.nix` and `release.nix`

`release.nix` needed to be reorganised to support Hydra building for multiple systems.

`default.nix` is cleaned up and `self.callPackage ./installers/....nix {}` is used for everything.

https://hydra.iohk.io/jobset/serokell/daedalus-pr-1006

## Todo

- [x] Fix missing icon of running application.
- [x] Replace bundling of `libgconf-2.so.4` with `fetchurl` of an ubuntu deb file.
- [x] Test update system.
- [x] Builds both nix-bundle installer and AppImage.


installers/yarn2nix.nix
```diff
@@ -1,30 +1,27 @@
...
+  yarnLock = ../yarn.lock;
+  packageJSON = ../package.json;
+
+  API = backend.api;
+  API_VERSION = backend.version;
   CI = "nix";
-  NETWORK = networkMap.${cluster};
-  BUILD_NUMBER = "${toString buildNum}";
+  DAEDALUS_VERSION = "${version}";
+  # BUILD_NUMBER = toString buildNum;
```
angerman: Should we just drop this instead of leaving it commented out?

rvl: Yes - fixed.

nikolaglumac:
@rvl what is the state of this PR?

rvl:
@nikolaglumac it has been ready to merge for a while and has been sitting waiting for review.
Every so often a change in the develop branch breaks this PR and I have to fix it.

nikolaglumac:
@rvl sorry for that! I don't see it has been approved? This is a PR for which we need approval from your team members I guess. Will these changes have any impact on our standard development workflow?

rvl:
I'm hoping that it won't affect existing workflows, but will make reading and modifying the nix code easier.

deepfire:
@rvl, one thing that bothers me a lot is that you mention system-provided Electron libraries in https://iohk.myjetbrains.com/youtrack/issue/DEVOPS-487#comment=93-25083.

Frankly, I find it the weakest point of the proposal, but I'm also not a web developer, and I find it hard to assess how often API breakages happen in the Electron world.

Mayme @nikolaglumac can help us here?

deepfire:
One thing, I guess, is to try figuring out what is the tentative spread of Electron libraries across the landscape of Linux distributions.

deepfire:
Another thing, I guess, is that it somewhat neuters the assurance that we get from our QA, or increases the load they get -- or both.

I.e. now, all properties of the Linux system we depend on are fairly simple and low-level -- like the user namespaces.  So the inter-distribution variance is fairly low, and similarly any changes to be introduced by distribution-level upgrades are fairly minimal.

On the other hand, once we admit Chromium/Electron into the basis we depend upon for compatibility.. we make the situation significantly more complicated.

So to establish the _same_ level of assurance we'd have to do _more_ QA:
1. on a _wider_ variety of systems (since Electron packaging is different)
2. more often, to account for distribution/Electron upgrades

I find this a potential source of user tickets.. : -/

nikolaglumac:
Let's not merge this one until we have the `release/0.12.0` branch cut off...

probonopd:
What is holding you back from publishing an AppImage?

rvl:
Hi @probonopd, thanks for asking! In the case of this PR, it's not a priority at present and the code has gotten out of date with the main development branch. We have a self-extracting installer which works well enough.

Additionally, due to the way our auto update system is implemented, the launcher code needed changes to support AppImage. So that's why we're using the installer for now.

We are going to rework the launcher code in future, so I will make sure the design can accommodate AppImage. I will also try to update this PR soon so that we can have AppImage builds just for testing purposes.

disassembler:
Stale, closing.

PR #1010, opened 2018-07-13
[DEVOPS-944] Bump cardano-sl to latest release/1.3.0

rvl:
This brings Daedalus up to date with cardano-sl 1.3.0 after input-output-hk/cardano-sl#3235 and input-output-hk/cardano-sl#3242. These are fixes to block streaming and don't really affect the wallet.

### DIFF
--- a/cardano-sl-src.json
+++ b/cardano-sl-src.json
@@ -1,6 +1,6 @@
 {
   "url": "https://github.com/input-output-hk/cardano-sl",
-  "rev": "f7b6661121f1eead48043d99b9081d5ab52ff519",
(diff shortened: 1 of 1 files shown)


jmitchell:
AppVeyor is failing with this error.

```
Artifacts are: 
make-installer: MissingArtifactsError "https://ci.appveyor.com/project/input-output/cardano-sl/build/1.0.29491"
dontfail
FATAL: failed to build installer
```

I haven't yet verified whether that URL is as expected, but that's a good place to look.

rvl:
Yes, this happened because the cardano-sl appveyor build for this commit failed because of the 90 minute timeout. So we will need to restart that build, then restart this build.

PR #1026, opened 2018-07-27
Amias ddw 369

amias-iohk:
This PR makes the nodelLogPath entry in launcher-config.yaml optional , this will have the effect of disabling logging to cardano-node.log which was deamed to be causing a performance issue

### Testing
- [x]  Installer from this PR to be tested 
- [x]  Should start up on all platforms
- [x]  Do a restore on a MacOS vm to verfiy performance fix
- [x]  Check pub logging is still working

### After Review:
- [ ] Merge PR
- [ ] Delete source branch
- [ ] Move DDW-369 ticket to `done` on the Youtrack board


amias-iohk:
Could we get Duncan to review this , he doesn't seem to be visible to github

PR #1027, opened 2018-07-27
[DDW-369] Disable nodeLogPath in dhall files

disassembler:
This PR makes the nodelLogPath entry in launcher-config.yaml optional , this will have the effect of disabling logging to cardano-node.log which was deamed to be causing a performance issue

### Testing
- [x]  Installer from this PR to be tested 
- [x]  Should start up on all platforms
- [x]  Do a restore on a MacOS vm to verfiy performance fix
- [x]  Check pub logging is still working

### After Review:
- [ ] Merge PR
- [ ] Delete source branch
- [ ] Move DDW-369 ticket to `done` on the Youtrack board

### DIFF
--- a/installers/dhall/linux64.dhall
+++ b/installers/dhall/linux64.dhall
@@ -19,8 +19,7 @@ in
   , nodePath            = "cardano-node"
   , nodeDbPath          = "${dataDir}/DB/"
   , nodeLogConfig       = "\${DAEDALUS_CONFIG}/daedalus.yaml"
-  , nodeLogPath         = "${dataDir}/Logs/cardano-node.log"
-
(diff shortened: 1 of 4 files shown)


amias-iohk:
Testing was completed on mac vm (10.13) and a Windows 7 laptop using previous PR build  , logging changes where observed and app was faster to operate. Restore time was as faster than previous release build.

cleverca22:
of note, there was a bug near this config on a version of mac (forget which one), that resulted in daedalus not starting at all when ran, so it should be tested on all mac versions we have access to

deepfire:
This is a repeat of #1026, since it was closed, but we can't reopen closed PRs.

PR #1138, opened 2018-10-23
[DEVOPS-1090] Update nixpkgs from 17.09 -> 18.09

rvl:
Updates the pinned nixpkgs version to the latest stable version.

This will probably help with nix building Daedalus on macOS Mojave.

However the main problem for Mojave build support is Cardano SL, and its nixpkgs isn't updated yet.


disassembler:
Iohk--nix addressed this.

PR #1141, opened 2018-10-24
[DEVOPS-1109] cli argument to inhibit console log: --log-console-off

CodiePP:
Signed-off-by: Alexander Diemand <codieplusplus@apax.net>

This PR adds the `--log-console-off` cli argument that the node will not log to stdout.

YT: DEVOPS-1109
cardano-sl PR: https://github.com/input-output-hk/cardano-sl/pull/3785
---

### Basics

- [x] PR is updated to the most recent version of target branch (and there are no conflicts)
- [~] PR has good description that summarizes all changes and shows some screenshots or animated GIFs of important UI changes
- [~] CHANGELOG entry has been added and is linked to the correct PR on GitHub
- [ ] Automated tests: All acceptance tests are passing (`yarn run test`)
- [ ] Manual tests (minimum tests should cover newly added feature/fix): App works correctly in *development* build (`yarn run dev`)
- [ ] Manual tests (minimum tests should cover newly added feature/fix): App works correctly in *production* build (`yarn run package` / CI builds)
- [ ] There are no *flow* errors or warnings (`yarn run flow:test`)
- [ ] There are no *lint* errors or warnings (`yarn run lint`)
- [ ] Text changes are proofread and approved (Jane Wild)
- [ ] There are no missing translations (running `yarn run manage:translations` produces no changes)
- [ ] UI changes look good in all themes (Alexander Rukin)
- [ ] Storybook works and no stories are broken (`yarn run storybook`)
- [ ] In case of dependency changes `yarn.lock` file is updated

### Code Quality
- [ ] Important parts of the code are properly documented and commented
- [ ] Code is properly typed with flow
- [ ] React components are split-up enough to avoid unnecessary re-rendering
- [ ] Any code that only works in Electron is neatly separated from components

### Testing
- [ ] New feature / change is covered by acceptance tests
- [ ] All existing acceptance tests are still up-to-date
- [ ] New feature / change is covered by Daedalus Testing scenario
- [ ] All existing Daedalus Testing scenarios are still up-to-date

### After Review:
- [x] Merge PR
- [x] Delete source branch
- [x] Move ticket to `done` on the Youtrack board

PR #1144, opened 2018-10-24
[DEVOPS-1109] cli argument to inhibit console log: --log-console-off

CodiePP:
replaces PR #1141, now against `release/0.12.0`


nikolaglumac:
@CodiePP Cardano revision you have used in this PR does not point to the Cardano `develop` branch right? We should change this once your Cardano changes are merged into develop...

PR #1157, opened 2018-11-01
[DDW-256] No disk space overlay

daniloprates:
This PR Implements a better handling for no-diskspace error

## Todo:

- [x] No disk space overlay
- [ ] Container functions and variables
- [ ] English copy - @darko-mijic 
- [ ] Japanese translation - @darko-mijic 

### Before

image

### After

image

---

### Basics

- [ ] PR is updated to the most recent version of target branch (and there are no conflicts)
- [ ] PR has good description that summarizes all changes and shows some screenshots or animated GIFs of important UI changes
- [ ] CHANGELOG entry has been added and is linked to the correct PR on GitHub
- [ ] Automated tests: All acceptance tests are passing (`yarn run test`)
- [ ] Manual tests (minimum tests should cover newly added feature/fix): App works correctly in *development* build (`yarn run dev`)
- [ ] Manual tests (minimum tests should cover newly added feature/fix): App works correctly in *production* build (`yarn run package` / CI builds)
- [ ] There are no *flow* errors or warnings (`yarn run flow:test`)
- [ ] There are no *lint* errors or warnings (`yarn run lint`)
- [ ] Text changes are proofread and approved (Jane Wild)
- [ ] There are no missing translations (running `yarn run manage:translations` produces no changes)
- [ ] UI changes look good in all themes (Alexander Rukin)
- [ ] Storybook works and no stories are broken (`yarn run storybook`)

### Code Quality
- [ ] Important parts of the code are properly documented and commented
- [ ] React components are split-up enough to avoid unnecessary re-rendering
- [ ] Any code that only works in Electron is neatly separated from components

### Testing
- [ ] New feature / change is covered by acceptance tests
- [ ] New feature / change is covered by Daedalus Testing scenario

### After Review:
- [ ] Merge PR
- [ ] Delete source branch
- [ ] Move ticket to `done` on the Youtrack board


nikolaglumac:
@daniloprates looks like you forgot to add `prettysize` plugin into `package.json` and `yarn.lockfile`?
screen shot 2018-11-06 at 10 56 54

daniloprates:
@darko-mijic I've adapted the text to accommodate the changes, as shown below and the screenshot above. Could you confirm if it's OK? Also, we need new Japanese translations. Thanks.

> **There is not enough disk space left on your device.**
> Daedalus requires at least 2 GB to operate. Please free up some disk space to continue.

daniloprates:
Still `WIP` as it needs confirmation for the overlay text, but the functionality is ready and can be reviewed.

DominikGuzei:
Checking this now now …

daniloprates:
Dear reviewers, I have added the correct copy, with its calculations. It's ready for review, only missing the Japanese translation.

daniloprates:
Japanese translations ✔ PR is ready.

nikolaglumac:
@daniloprates could we cover this feature with acceptance tests? I think that's very important 🙏

nikolaglumac:
@daniloprates I have merged latest develop into your branch.
It looks like there are some lint and flow errors. Can you please sort those out? Thanks!

screen shot 2018-12-14 at 13 16 31
screen shot 2018-12-14 at 13 16 36

daniloprates:
Acceptance tests are done. PR ready to be reviewed.

daniloprates:
@nikolaglumac ✅

daniloprates:
@nikolaglumac regarding the disk space, it seems that info screen is not completely acurate:

image
image
image
image

daniloprates:
Maybe we should inform the disk total space in the overlay, or maybe find out how to show the total including these protected bytes.

nikolaglumac:
@daniloprates are you sure these are protected bytes? It appears to me that this discrepancy in total size could be related to disk size vs. partition size?

daniloprates:
> @daniloprates are you sure these are protected bytes? It appears to me that this discrepancy in total size could be related to disk size vs. partition size?

Yes, it could be. But are still some non-available bytes. I'll try to find another package that also returns the total available, including this non-available part.

daniloprates:
@nikolaglumac tested 3 other packages ([`hdd-space`](https://www.npmjs.com/package/hdd-space), [`diskusage`](https://www.npmjs.com/package/diskusage) and [`diskspace`](https://www.npmjs.com/package/diskspace)) and they all return only this real available space. Maybe we can check using node directly, but it's quite complicated, as it returns all the partitions, volumes and any other attached devides.

nikolaglumac:
@daniloprates let's keep the current solution. The important part is that the available space is correctly estimated and that part is fine.

daniloprates:
@nikolaglumac OK. I'll work in the other part. I think I know how to do the check before cardano starts, but I might need your help on that.

features/no-disk-space.feature
```diff
@@ -0,0 +1,16 @@
+Feature: No Disk Space Overlay
+
+  Background:
+    Given I have completed the basic setup
+
+  Scenario: I set the required space to 1 KB
```
nikolaglumac: @daniloprates the scenario title should be more descriptive - e.g. "No disk space overlay is not shown when there is enough available disk space"

features/no-disk-space.feature
```diff
@@ -0,0 +1,16 @@
...
+    Given I set the required space to 1 KB
+    And I check the disk space
+    Then The No Disk Space overlay should be hidden
+
+  Scenario: I set the required space to 100 TB
```
nikolaglumac: @daniloprates the scenario title should be more descriptive - e.g. "No disk space overlay is shown when there is not enough available disk space and hidden once more disk space becomes available"

features/no-disk-space.feature
```diff
@@ -0,0 +1,16 @@
...
+    Given I set the required space to 100 TB
+    And I check the disk space
+    Then The No Disk Space overlay should be visible
+    And After 10 seconds
```
nikolaglumac: @daniloprates this test step should be replaced with a more descriptive step such as "After I free up disk space".

source/main/cardano/setupCardanoNodeMode.js
```diff
@@ -43,7 +44,11 @@ const startCardanoNode = (node: CardanoNode, launcherConfig: Object) => {
   return node.start(config);
 };
 
-const restartCardanoNode = async (node: CardanoNode) => {
+export const restartCardanoNode = async (node: CardanoNode) => {
+  if (
+    node.state === CardanoNodeStates.STARTING ||
+    node.state === CardanoNodeStates.RUNNING
```
nikolaglumac: @daniloprates do we really want to prevent restart in case Cardano node is in `RUNNING` state? It looks to me we should allow the restart in this case and perhaps prevent it in other non-final states such as: `STOPPING, EXITING, UPDATING`...

features/no-disk-space.feature
daniloprates: What actually happens after the 10 seconds is that it will check using the default `DISK_SPACE_REQUIRED` param, therefore I replaced this step with `I set the required space to 1 KB`.

source/main/cardano/setupCardanoNodeMode.js
daniloprates: You're right, I moved the node.state checking into the `onCheckDiskSpace` function, as in this case it doesn't make sense to restart it if it's already running.

daniloprates:
@nikolaglumac 
> logging says I have diskSpaceMissing but there is no notification shown :(

It didn't reach the `diskSpaceRequired` - 10% to trigger the `noDiskSpace` variable:
* If this space is lower than 2 GB - 10% we don't start the node or stop it if it is running, we show the red screen explaining that Daedalus requires at least 2 GB of free space but we strongly

daniloprates:
@nikolaglumac 
> we will still have our pollers running and throwing errors.

✅ [4a3f259](https://github.com/input-output-hk/daedalus/pull/1157/commits/4a3f2591a5ed6946e570f5ef0fe0c4f58520a566)

source/common/ipc-api.js
```diff
@@ -33,3 +33,9 @@ export const GET_GPU_STATUS = {
   SUCCESS: `${GET_GPU_STATUS_CHANNEL}-success`,
   ERROR: `${GET_GPU_STATUS_CHANNEL}-error`,
 };
+
+const GET_DISK_SPACE_STATUS_CHANNEL = 'get-disk-space-status';
+export const GET_DISK_SPACE_STATUS = {
+  REQUEST: GET_DISK_SPACE_STATUS_CHANNEL,
+  SUCCESS: `${GET_DISK_SPACE_STATUS_CHANNEL}-success`,
+};
```
DominikGuzei: Please refactor this dynamically typed constants-based (old) way of ipc channel communication to our new (3-4 months already) Flow-typed, promise based ipc channels: https://github.com/input-output-hk/daedalus/tree/develop/source/common/ipc 😉

source/main/index.js
```diff
@@ -96,6 +98,30 @@ const onAppReady = async () => {
...
+    if (cardanoNode) {
+      if (notEnoughSpace) {
+        try {
+          cardanoNode.stop();
+        } catch (e) {} // eslint-disable-line
+      } else if (
+        cardanoNode.state !== CardanoNodeStates.STARTING &&
+        cardanoNode.state !== CardanoNodeStates.RUNNING
+      ) {
+        restartCardanoNode(cardanoNode);
```
DominikGuzei: Why not use `node.restart()` directly here?

source/main/utils/HandleDiskSpace.js
```diff
@@ -0,0 +1,71 @@
+// @flow
```
DominikGuzei: Since this is not a class i would rather use lower case letter for the first char of the file name (`handleDiskSpace` or maybe even `disk-space.js`)

source/main/utils/HandleDiskSpace.js
```diff
@@ -0,0 +1,71 @@
...
+import { Logger } from './logging';
+import {
+  DISK_SPACE_REQUIRED,
+  DISK_SPACE_REQUIRED_MARGIN_PERCENTAGE,
+  DISK_SPACE_CHECK_LONG_INTERVAL,
+  DISK_SPACE_CHECK_SHORT_INTERVAL,
+  DISK_SPACE_RECOMMENDED_PERCENTAGE
+} from '../config';
+
+export default (
```
DominikGuzei: I think we agreed on always using explicit exports like `export const handleDiskSpace = …` (because this makes refactoring tasks easier, since the default export can be named arbitrarily in the import location and thus refactoring tools do not pick that up correctly)

source/main/utils/HandleDiskSpace.js
```diff
@@ -0,0 +1,71 @@
...
+  mainWindow: BrowserWindow,
+  onCheckDiskSpace?: Function,
+) => {
+  const path = environment.isWindows ? 'C:' : '/';
```
DominikGuzei: Maybe there is a better way to let node.js handle the os differences?

source/main/index.js
```diff
@@ -96,6 +98,30 @@ const onAppReady = async () => {
   const isInSafeMode = includes(process.argv.slice(1), '--safe-mode');
 
   mainWindow = createMainWindow(isInSafeMode);
+
+  const onCheckDiskSpace = ({ notEnoughSpace }: CheckDiskSpaceResponse) => {
+    if (cardanoNode) {
+      if (notEnoughSpace) {
+        try {
+          cardanoNode.stop();
```
nikolaglumac: @daniloprates @DominikGuzei we should perhaps check if cardanoNode is already stopped before trying to stop it again. That way we would avoid many stopping attempts: 

screen shot 2019-01-08 at 13 12 06

nikolaglumac:
> @nikolaglumac
> 
> > logging says I have diskSpaceMissing but there is no notification shown :(
> 
> It didn't reach the `diskSpaceRequired` - 10% to trigger the `noDiskSpace` variable:
> 
> * If this space is lower than 2 GB - 10% we don't start the node or stop it if it is running, we show the red screen explaining that Daedalus requires at least 2 GB of free space but we strongly

You are right! Works great 🎉

source/main/index.js
```diff
@@ -6,6 +6,7 @@ import { includes } from 'lodash';
 import { Logger } from '../common/logging';
 import { setupLogging } from './utils/setupLogging';
 import { makeEnvironmentGlobal } from './utils/makeEnvironmentGlobal';
+import HandleDiskSpace from './utils/HandleDiskSpace';
```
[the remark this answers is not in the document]
daniloprates: Sure! It's done

source/common/ipc-api.js
daniloprates: ✅

source/main/index.js
daniloprates: That's odd, I tried before and got some errors. Must have done something wrong. Changed, now.

source/main/utils/HandleDiskSpace.js
daniloprates: ✅

source/main/index.js
```diff
@@ -96,6 +98,30 @@ const onAppReady = async () => {
   const isInSafeMode = includes(process.argv.slice(1), '--safe-mode');
 
   mainWindow = createMainWindow(isInSafeMode);
+
+  const onCheckDiskSpace = ({ notEnoughSpace }: CheckDiskSpaceResponse) => {
+    if (cardanoNode) {
```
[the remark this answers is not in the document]
daniloprates: ✅

source/main/utils/HandleDiskSpace.js
daniloprates: I couldn't find a way. I got this code from one the [packages](https://www.npmjs.com/package/diskusage) that check disk space.

source/main/index.js
nikolaglumac: This probably due to our automatic restart feature which is used during initial Daedalus start.
In case cardano-node exits Daedalus will try to start it again for 5 consecutive times before giving up. Not sure how we can bypass that in this case. Help appreciated @DominikGuzei 🙏

nikolaglumac: @daniloprates I guess your original code simply shuts it down every of these 5 restart attempts. Which makes it look like it is working BUT we should avoid even trying to restart it if we have shut it down due to not-enough disk space.

nikolaglumac:
@daniloprates please remove the `WIP` tag once this PR is ready for the final review 🙏

source/main/index.js
daniloprates: Not sure how to deal with it. I have added the [condition](https://github.com/input-output-hk/daedalus/pull/1157/commits/b1b24475d2c54294d464e00f3722b0c5aac30155#diff-286decdfc7881f7334e73465dab511ffR106), but as the node restarts (it won't trigger the `node.restartNode` nor the `node.start` functions, though), it's still calling `node.stop` every 10 secs.

nikolaglumac: I will sort this out @daniloprates - no worries!

nikolaglumac: @daniloprates fixed in: https://github.com/input-output-hk/daedalus/pull/1157/commits/245017e19ea98e06e99b61460cdc5768228b9ec3

nikolaglumac:
We'll need to improve this feature...
The problem we are seeing while running manual tests is that the 10-min pollers we run when there's enough space are not good enough. Given that user starts with exactly 2GB of free disk space and Daedalus has just started syncing it will run out of disk space much sooner than our next disk-space check is run. The problem in this scenario is that things start to fall apart when we reach very low disk space - before the ENOSPC error is thrown... This means Daedalus will crash before we even show the not-enough-disk-space error screen.
In order to avoid this from happening we should implement dynamic disk-space-check polling. Such which would run more frequently as we get closer to the 2GB of free disk space.
The exact intervals are yet to be decided but a logic such as this one should be good enough:
- if there is more than 100GB of free disk space the check should be run every 10 mins,
- if there is more than 20 GB and less than 100GB of free disk space the check should run every 5 mins,
- if there is more than 10 GB and less than 20 GB of free disk space the check should run every 2 mins,
- if there is more than 5 GB and less than 10 GB of free disk space the check should run ever 1 min,
- if there is more than 2 GB and less than 5 GB of free disk space the check should run every 20 sec.

cc @daniloprates @DominikGuzei @darko-mijic @DmitriiGaico

nikolaglumac:
@DominikGuzei as I have pointed out in my comment - this was just a suggestion.
Regardless of the disk space operation cost, I would prefer not to run this check too often if there is no reason to do so (aka. if there is a bunch of free space). Of course, we don't need to have 10 different intervals. We could probably get by with having 3...

DominikGuzei:
i was just asking because you have proposed a complicated logic for the interval based on free disc space … now i wondered why 😉

nikolaglumac:
I still think we should have it like that - just simpler ;)

nikolaglumac:
@DominikGuzei I have added only one more interval: https://github.com/input-output-hk/daedalus/pull/1157/commits/23de0d77d8b9b4aa3831542fd49485e29200bdea

Here's the new setup:
| Available disk space | Interval |
|----------------------|:--------:|
| >= 0 && < 2 GB       |  10 sec  |
| >= 2 && < 4 GB       |   1 min  |
| >= 4 GB              |  10 min  |

DominikGuzei:
Looks good @nikolaglumac 👍

PR #1247, opened 2019-01-03
[DDW-541] Fix blank window rendering issue

nikolaglumac:
This PR reverts Electron to version `3.0.13` and keeps the `electron-chromedriver: 4.0.0-beta.1` as this combination fixes all blank window rendering issues and keeps the acceptance tests working too.

## Todos:

- [x] Find a way to re-establish `enableRemoteModule: isTest` and in the same time avoid the blank screen issues...

### Installer:
screen shot 2019-01-03 at 11 30 46

### Development: 
screen shot 2019-01-03 at 13 29 46

### The problem is caused by `electron-storage` package which requires `remote` in order to run (which we prevent by setting `enableRemoteModule: false` for non-test environments) :
screen shot 2019-01-03 at 14 18 54

---

### Basics

- [ ] PR is updated to the most recent version of target branch (and there are no conflicts)
- [ ] PR has good description that summarizes all changes and shows some screenshots or animated GIFs of important UI changes
- [ ] CHANGELOG entry has been added and is linked to the correct PR on GitHub
- [ ] Automated tests: All acceptance tests are passing (`yarn run test`)
- [ ] Manual tests (minimum tests should cover newly added feature/fix): App works correctly in *development* build (`yarn run dev`)
- [ ] Manual tests (minimum tests should cover newly added feature/fix): App works correctly in *production* build (`yarn run package` / CI builds)
- [ ] There are no *flow* errors or warnings (`yarn run flow:test`)
- [ ] There are no *lint* errors or warnings (`yarn run lint`)
- [ ] Text changes are proofread and approved (Jane Wild)
- [ ] There are no missing translations (running `yarn run manage:translations` produces no changes)
- [ ] UI changes look good in all themes (Alexander Rukin)
- [ ] Storybook works and no stories are broken (`yarn run storybook`)

### Code Quality
- [ ] Important parts of the code are properly documented and commented
- [ ] React components are split-up enough to avoid unnecessary re-rendering
- [ ] Any code that only works in Electron is neatly separated from components

### Testing
- [ ] New feature / change is covered by acceptance tests
- [ ] New feature / change is covered by Daedalus Testing scenario

### After Review:
- [ ] Merge PR
- [ ] Delete source branch
- [ ] Move ticket to `done` on the Youtrack board

### DIFF
--- a/installers/nix/electron.nix
+++ b/installers/nix/electron.nix
@@ -1,0 +1,78 @@
+{ stdenv, libXScrnSaver, makeWrapper, fetchurl, unzip, atomEnv, libuuid, at-spi2-atk }:
+
+let
+  version = "4.0.0";
+  name = "electron-${version}";
+
+  throwSystem = throw "Unsupported system: ${stdenv.hostPlatform.system}";
+
+  meta = with stdenv.lib; {
+    description = "Cross platform desktop application shell";
+    homepage = https://github.com/electron/electron;
+    license = licenses.mit;
+    maintainers = with maintainers; [ travisbhartwell manveru ];
+    platforms = [ "x86_64-darwin" "x86_64-linux" "i686-linux" "armv7l-linux" "aarch64-linux" ];
+  };
+
+  linux = {
+    inherit name version meta;
+
+    src = {
+      i686-linux = fetchurl {
+        url = "https://github.com/electron/electron/releases/download/v${version}/electron-v${version}-linux-ia32.zip";
+        sha256 = "0yv2f7yf6ingjysswpnpnvqsjkdkp2rd4laawhziifzbfjda4yws";
+      };
+      x86_64-linux = fetchurl {
+        url = "https://github.com/electron/electron/releases/download/v${version}/electron-v${version}-linux-x64.zip";
+        sha256 = "1kh2jds7jra9f1vcn2z1193cxcyvfxkldim4b9ij7chj9xzxwgln";
+      };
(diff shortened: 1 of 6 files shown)


nikolaglumac:
@DominikGuzei I will need your help regarding this one 🙏

nikolaglumac:
> @nikolaglumac does this only happen on electron 4?

@DominikGuzei interestingly, it is not happening on Electron 3 😱

DominikGuzei:
@nikolaglumac yeah i think they improved the security in electron 4 and now `enableRemoteModule: false` really, really disables it (also in preload.js) … i read somewhere about incidents about the `remote` module leaking into the global scope when it was exposed in the wrong way inside `preload.js` … and to be honest i think it's a good thing anyway to completely remove the direct / indirect access to `remote` and move all that logic to the main process.

nikolaglumac:
@DominikGuzei I have now downgraded Electron to `3.0.13` and the issue is gone which means you are right - Electron 4 improved security is the culprit.

A preferred solution for this problem would be as you said - removal of access to `remote` from the rendering process. Unfortunately, it looks like Electron 4 build fails both on Windows and Linux (even with my fix from this PR). Thus, I will push an update with the downgrade to Electron 3.0.13 and revert of the `enableRemoteModule` change.
The original reason why we have upgraded to Electron 4 was because of acceptance test issues.
These issues are fixed by upgrading `electron-chromedriver` to v4 which enables us to use Electron 3 🎉

PR #1514, opened 2019-08-07
Nix dependency fix

Sam-Jeston:
**There were 2 issues here:**

**1. yarn2nix bug**
The bug is defined in this issue: https://github.com/moretea/yarn2nix/issues/53

Due to a new linting dependency, we ended up with this in our yarn.lock:

```
"@types/events@*":
  version "3.0.0"
  resolved "https://registry.yarnpkg.com/@types/events/-/events-3.0.0.tgz#2862f3f58a9a7f7c3e78d79f130dd4d71c25c2a7"

events@^3.0.0:
  version "3.0.0"
  resolved "https://registry.yarnpkg.com/events/-/events-3.0.0.tgz#9a0a0dfaf62893d92b875b8f2698ca4114973e88"
```

The existing release of yarn2nix fails to differentiate between @types and regular dependencies if their version is exactly the same.

To resolve this, I just rolled the lint dependency back a major version. It still worked and resolved this issue. I ran `yarn run lint` and `yarn run lint:fix` to verify the package still works.

We can upgrade to the newer version of yarn2nix to prevent this in the future (although its an unlikely outcome). When we tried the newer version of yarn2nix, unexpectedly some packages had to be changed in the package.json (for no apparent reason), and I had to update the bable config handling to get flow passing, and I'd prefer avoid that.

**2. yarn.lock version changes**
By deleting and recreating the yarn.lock, we lose its value. In this instance, a dependency of a dependency had gone up a minor version, and this caused the break as it had been loosely specified by that parent dependency. I've now specified this resolutely in the package.json, but not deleting and re-creating the yarn.lock will protect us from this.


nikolaglumac:
This broken CI nix issue was caused by this PR: https://github.com/input-output-hk/daedalus/pull/1483
cc @daniloprates @tomothespian 

@tomothespian we need to make sure if we update these packages that you added for style linting that CI is passing before merging it.

PR #1567, opened 2019-09-13
[DDW-897] Rust node integration

Sam-Jeston:
This PR:

 - Changes the default Cardano node implementation to be Rust node to facilitate the incentivized testnet
 - Updates the V2 API importer to be Rust node style, not http-bridge
 - Adds the nodeImplementation and isIncentivizedTestnet to the relevent stores

My thoughts are that incentivized testnet features moving forward can be driven from 

```
const { networkStatus } = stores
networkStatus.isIncentivizedTestnet
```

@tomothespian this branch has moved us to the Edge version (for the short term) of `cardano-wallet`. You might want to sanity check this branch to see if things that were working are broken. I expect this might be the case, but I'm not sure what state you had the branch in. 

@cleverca22 I kept the config in the launcher.dhall after our conversation because it felt to me like each node should have a configuration file, and baking the value into the compiled JS wouldn't be ideal. I think this is a slightly more future proof approach as well. If we ever want to change out nodes implementation in realtime, then we will already expect to read and set the implementation from the LauncherConfig, instead of a single baked in ENV var


Sam-Jeston:
Ah yes I did see that @nikolaglumac, but wasn't sure if it was due to an unstable feature @tomothespian had worked on. Now that I know that it isn't, I will investigate

nikolaglumac:
Nope @Sam-Jeston - the `v2-integration` branch should be fully functional as far as the wallet handling is concerned.

Sam-Jeston:
@nikolaglumac this is resolved.

When we go to start Jormungandr, we check to see if a secret file exists in the state directory, as this is required to run the node. If it doesn't, we create it and create the block0.bin from a default testnet genesis.yml. This yml references the public key of the private key stored in the secret file, hence the string interpolation prior to writing the custom genesis.yml to disk.

PR #2258, opened 2020-12-01
Fix empty disk space calculation for some Linux distributions

rvl:
The Linux installer runs Daedalus inside a chroot, which lives in a temporary directory.

Some - perhaps most - Linux distros have the temporary directory as a tmpfs (RAM drive).

So when Daedalus checks the free space in /, it is reporting the amount of free space on the tmpfs - typically only a few GB.

This change makes Daedalus check the free space for the state directory.

In any case, this is a good thing to do, because the state directory (in the user's home directory) may be on a different partition to /.

### Testing

I haven't done any testing on this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Changelog
 
 ### Fixes
 
+- Fixed empty disk space calculation for some Linux distributions ([PR 2258](https://github.com/input-output-hk/daedalus/pull/2258))
 - Fixed the overlap of the "X" button in the stake pools search box ([PR 2251](https://github.com/input-output-hk/daedalus/pull/2251))
 
 ## 2.5.0
--- a/source/main/config.js
+++ b/source/main/config.js
@@ -125,7 +125,7 @@ export const {
 export const appLogsFolderPath = logsPrefix;
 export const pubLogsFolderPath = path.join(appLogsFolderPath, 'pub');
 export const stateDirectoryPath = stateDir;
-export const stateDrive = isWindows ? stateDirectoryPath.slice(0, 2) : '/';
+export const stateDrive = isWindows ? stateDirectoryPath.slice(0, 2) : stateDir;
 export const buildLabel = getBuildLabel(
   build,
   network,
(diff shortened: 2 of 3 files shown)


gnpf:
@rvl @nikolaglumac I compared this with an old build on all platforms and the information about the disk space seems to be as accurate with this fix. Here are the results:

**Another build:**
[Windows](https://drive.google.com/file/d/1ZT8NciCSMmcz7_T24EhffNy3RsZuviBt/view?usp=sharing)
[Mac](https://drive.google.com/file/d/1pFAEAIehiracy-W5nXmh62P_xOZQLAf2/view?usp=sharing)
[Linux](https://drive.google.com/file/d/1VTb0K-sVLi-fkjn8Ep9vbPsZfMQUDxdz/view?usp=sharing)

**Build 15394:**
[Windows](https://drive.google.com/file/d/1nkmaPeZJ4T5ttAuG_1v3D4uIILochfYX/view?usp=sharing)
[Mac](https://drive.google.com/file/d/1aI671rSIaAtxVuUtpQ3TNtRWhX-PdxyV/view?usp=sharing) 
[Linux](https://drive.google.com/file/d/1-5ondpfnm6TOwbkJce0KuFPuWzqQRDKt/view?usp=sharing)

However, I couldn't reproduce this issue [initially](https://input-output-rnd.slack.com/archives/C7PJ29FME/p1600378302067000), even with a partitioned disk. Should we ask @KtorZ to review this too?

rvl:
I could reproduce on Linux Daedalus flight:

Screenshot from 2020-12-02 15-26-37

And it appears to be fixed on build 15394 testnet.

Screenshot from 2020-12-02 15-27-48

nikolaglumac (approved):
Great fix @rvl 🎉 
I have checked the [docs](https://github.com/Alex-D/check-disk-space#usage) and we can actually provide the full state directory path to all platforms.
That is what I updated [here](https://github.com/input-output-hk/daedalus/commit/4d2f9d320dbde2b77dd3d6fc2654078486714191) - simpler code - simpler life for us all :-)

PR #2774, opened 2021-12-03
Update README.md

Fresheyeball:


danielmain:
Hi @Fresheyeball thank you for the PR. I don't know if I understand, but why would you want to install Daedalus via nix-env? This documentation is intended for developers how to run Daedalus from the source code. If you are a Daedalus user not a developer, please download it (tested and signed) from [daedaluswallet.io](https://daedaluswallet.io/en/download/)

danielmain:
Closing due inactivity

Fresheyeball:
@danielmain so for example, I want to install Daedalus on my NixOS machine. How to do so is not documented anywhere I could find. I know people on NixOS who are running it in a nix-shell, (not looking to hack) just to run the wallet locally. I felt it would be helpful to add a line about how it can be installed on NixOS, the easiest way I could find was with `nix-env`.

ursi:
I will also suggest potentially using this instead, so the user can just copy-paste the command to install, instead of having to clone the repo.
```
nix-env -iE '_: (import (builtins.fetchGit "https://github.com/input-output-hk/daedalus.git") {}).daedalus'
```

PR #2811, opened 2022-01-11
[DDW-893] Update the PR checklist

marcin-mazurek:
This PR ensures the PR checklist reflects the current delivery process.


.github/pull_request_template.md
```diff
@@ -21,19 +35,15 @@ This PR CHANGES.
 
 ### Basics
 
-- [ ] PR has been assigned and has appropriate labels (`feature`/`bug`/`chore`, `release-x.x.x`)
+- [ ] PR assigned to `input-output-hk/daedalus-dev` and `input-output-hk/daedalus-qa`
```
lucas-barros, line 38: Shouldn’t it be something like "PR assigned to `<developer>`" and "Add reviewers `input-output-hk/daedalus-dev` ...."? This way might cause confusion between PR assignee and PR reviewer

DominikGuzei (changes requested):
Looks great 👍 i share the same concern like @lucas-barros

.github/pull_request_template.md
marcin-mazurek: Well, so this actually is another new thing for me - are there any rules about who (or how many people) should be assigned to the PR using the _Asignees_ field? I thought we only use _Reviewers_ and thought this is what the creator of the original checklist had in mind.

.github/pull_request_template.md
```diff
@@ -1,17 +1,31 @@
...
 
-## Screenshots
+- [ ] TODO
 
-- Screenshot
+## Screenshots
 
----
+<!---
+Use the GitHub drag&drop feature to upload default-sized Daedalus window screenshots
```
renanvalentin, line 18: I think should be worth mentioning that screenshots cannot have drop shadow. On OSX this can be achieved by pressing `Command + Shift + 4 + Space bar` 

Screen Shot 2021-12-24 at 09 36 30

lucas-barros (approved):

renanvalentin (approved):

PR #2838, opened 2022-01-25
[DDW-881] A feature responsible for rebuilding app menu on changes from various stores

szymonmaslowski:
This PR:
 - fixes behaviour of the wallet settings item of the menu:
   - enables the option if the user is on the one of wallet pages
   - disables (greyed with deactivated hotkey) the option if the user is on the one of wallet pages but has some dialog opened
   - hides (and deactivates the hotkey) the option if the user:
     - is not on the one of the wallet pages
     - is already on the wallet settings page
 - unifies IPC communication regarding menu rebuild to a single channel
 - removes implicit IPC `channel.send()` calls from react components and stores methods
 - creates new feature responsible for watching for all data changes requiring menu rebuild and triggering it

## Screenshots

image
image
image
image

## Testing Checklist

- [Slack QA thread](https://input-output-rnd.slack.com/messages/GGKFXSKC6)
- [ ] Test

---

### Basics
- [X] PR assigned to the PR author(s) 
- [X] `input-output-hk/daedalus-dev` and `input-output-hk/daedalus-qa` assigned as PR reviewers
- [X] If there are UI changes, Alexander Rukin assigned as an additional reviewer
- [ ] All visual regression testing has been reviewed (Go to Github Actions tab -> Select Chromatic workflow -> Run on your working branch)
- [X] PR has appropriate labels (`release-vNext`, `feature`/`bug`/`chore`, `WIP`)
- [ ] PR link is added to a Jira ticket, ticket moved to In Review
- [X] PR is updated to the most recent version of the target branch (and there are no conflicts)
- [X] PR has a good description that summarizes all changes
- [X] PR contains screenshots (in case of UI changes)
- [ ] CHANGELOG entry has been added to the top of the appropriate section (*Features*, *Fixes*, *Chores*) and is linked to the correct PR on GitHub
- [X] There are no missing translations (running `yarn manage:translations` produces no changes)
- [ ] Text changes are proofread and approved (Jane Wild / Amy Reeve)
- [ ] Japanese text changes are proofread and approved (Junko Oda)
- [ ] Storybook works and no stories are broken (`yarn storybook`)

### Code Quality

### Testing

### After Review


source/renderer/app/features/menu-updater/makeReactionCallback.js
```diff
@@ -0,0 +1,42 @@
...
+
+  if (walletSettingsOptionVisible) {
+    walletSettingsState = uiDialogs.activeDialog
+      ? WalletSettingsStateEnum.disabled
+      : WalletSettingsStateEnum.enabled;
+  }
+
+  if (profile.currentLocale) {
+    // Just mentioning profile.currentLocale to make the Reaction watch its changes
+  }
```
szymonmaslowski, line 34: That is not an elegant solution at all, however I had no other idea probably because I am new in the mobx world.

alexander-rukin:
image
this dialog should have setting option disabledd too

szymonmaslowski:
> image
> 
> this dialog should have setting option disabledd too

This case is covered now.

gabriela-ponce:
@szymonmaslowski Here are a couple of scenarios where the "wallet settings" option is still visible and enabled:

- Transactions tab - Export CSV dialog -  Check [this](https://drive.google.com/file/d/14zZwXDoDp63Sr977Z9-p43FdBOx09RES/view?usp=sharing) screenshot.
- Send tab - Add tokens dialog-  Check [this](https://drive.google.com/file/d/1uH5i4qVe5rHEcYO-qdIFLpyPWiTKcbDN/view?usp=sharing) screenshot.

source/renderer/app/api/utils/localStorage.js
```diff
@@ -375,15 +375,6 @@ export default class LocalStorageApi {
...
 
-  getAssetSettingsDialogWasOpened = (): Promise<boolean> =>
-    LocalStorageApi.get(keys.ASSET_SETTINGS_DIALOG_WAS_OPENED, false);
-
-  setAssetSettingsDialogWasOpened = (): Promise<void> =>
-    LocalStorageApi.set(keys.ASSET_SETTINGS_DIALOG_WAS_OPENED, true);
-
-  unsetAssetSettingsDialogWasOpened = (): Promise<void> =>
-    LocalStorageApi.unset(keys.ASSET_SETTINGS_DIALOG_WAS_OPENED);
-
```
szymonmaslowski, line 386: I removed those three methods since they look like an overhead. We should not use persistent storage to store the information about whether the dialog is open since it is just a runtime information. Except the usage in the `AssetsStore` which was just checking whether the dialog was open I didn't found any other usages so they were safe to remove.

szymonmaslowski:
> @szymonmaslowski Here are a couple of scenarios where the "wallet settings" option is still visible and enabled:
> 
> * Transactions tab - Export CSV dialog -  Check [this](https://drive.google.com/file/d/14zZwXDoDp63Sr977Z9-p43FdBOx09RES/view?usp=sharing) screenshot.
> * Send tab - Add tokens dialog-  Check [this](https://drive.google.com/file/d/1uH5i4qVe5rHEcYO-qdIFLpyPWiTKcbDN/view?usp=sharing) screenshot.

Thanks @gabriela-ponce for rising those.
Regarding the Add tokens dialog - already working on a fix for it it.
Regarding the Export CSV I didn't consider it a dialog, but you are right. I have consulted with @alexander-rukin on that and we agreed it is worth to fix if it is relatively small amount of effort so I will check how is it implemented and if that will be easy to fix I will do that.

szymonmaslowski:
@gabriela-ponce The add tokens dialog is already fixed. The Export CSV native dialog unfortunately not - it requires more effort.
Apart from that I also found (and fixed) other three cases when wallet settings menu option supposed to be disabled
 - about dialog
 - diagnostics dialog
 - redeem rewards

source/renderer/app/components/wallet/WalletSendForm.js
```diff
@@ -1016,7 +1017,7 @@ export default class WalletSendForm extends Component<Props, State> {
                 label={intl.formatMessage(messages.addAssetButtonLabel)}
                 disabled={!this.hasAvailableAssets}
                 onClick={() => {
-                  this.setState({ isTokenPickerOpen: true });
+                  this.walletTokenPickerRef?.current?.open?.();
```
renanvalentin, line 1020: Why is it needed to use `ref` instead of `state`? This is creating a dependency between WalletSendForm and `WalletTokenPickerDialogContainer` as it requires the private method `open` to exist.

source/renderer/app/containers/tokens/WalletTokenPickerDialogContainer.js
```diff
@@ -0,0 +1,77 @@
...
+import type { ActionsMap } from '../../actions';
+import type { Props as PickerProps } from '../../components/wallet/tokens/wallet-token-picker/types';
+import WalletTokenPicker from '../../components/wallet/tokens/wallet-token-picker/WalletTokenPicker';
+import type { StoresMap } from '../../stores';
+
+export type Api = {
+  open: () => void,
+};
+
+type Props = PickerProps & {
```
renanvalentin, line 15: Would be possible to make this component stateless by providing `isOpen` and `onOpen` property. WalletSendForm.js could have a property `onOpenTokenPicker` and `WalletSendPage` container invokes `actions.dialogs.open.trigger` on `onOpenTokenPicker`

source/renderer/app/features/menu-updater/MenuUpdater.js
```diff
@@ -0,0 +1,35 @@
+// @flow
```
renanvalentin, line 1: I think this file should be renamed to `context.js` as it will be the feature provider if not mistaken.

source/renderer/app/features/menu-updater/MenuUpdater.js
```diff
@@ -0,0 +1,35 @@
...
+import { inject } from 'mobx-react';
+import type { StatelessFunctionalComponent } from 'react';
+import type { StoresMap } from '../../stores';
+import { useFeature } from '../../utils/mobx-features/hooks';
+import MenuUpdaterFeature from './feature';
+import type { MenuUpdaterProps } from './types';
+
+const MenuUpdater: StatelessFunctionalComponent<MenuUpdaterProps> = ({
```
renanvalentin, line 9: Maybe rename to `MenuUpdaterProvider` so we know this component is required by the app.

source/renderer/app/features/menu-updater/MenuUpdater.js
```diff
@@ -0,0 +1,35 @@
...
+  app,
+  children,
+  profile,
+  router,
+  staking,
+  uiDialogs,
+}) => {
+  useFeature(
```
renanvalentin, line 17: You can use this pattern to prevent `MenuUpdaterFeature` to be instantiated in every re-render:

```js
  const [menuUpdaterFeature] = useState(() => {
 new MenuUpdaterFeature({ app, profile, router, staking, uiDialogs })
  });
  useFeature(menuUpdaterFeature);
```

source/renderer/app/features/menu-updater/makeReactionCallback.js
renanvalentin: Maybe you could decorate [MenuUpdater](https://github.com/input-output-hk/daedalus/pull/2838/files#diff-cf32b35793444c25a614edfc40ddbf30f62d3a2181b84bc0f4b9bf954b1962d4R9)  with `observer`

source/renderer/app/features/menu-updater/makeReactionCallback.js
```diff
@@ -0,0 +1,43 @@
...
+import { ROUTES } from '../../routes-config';
+import type { MakeReactionCallbackArgs } from './types';
+
+const walletRoutes = Object.values(ROUTES.WALLETS);
+
+const makeReactionCallback = ({
+  stores: { app, profile, router, staking, uiDialogs },
+  rebuildApplicationMenu,
+}: MakeReactionCallbackArgs) => () => {
+  let walletSettingsState = WalletSettingsStateEnum.hidden;
```
renanvalentin, line 13: I'm not really sure how reaction works but another solution is to use `useEffect`. e.g:

```js
const ListenMenuChanges = observer(() => {

useEffect(() => {
 /* logic goes here ... */

rebuildApplicationMenu.send({
    isNavigationEnabled: profile.areTermsOfUseAccepted,
    walletSettingsState,
  });
},[])

return null
})
```

And use this on the app:

```
<MenuUpdaterProvider>
<App />
<ListenMenuChanges />
</MenuUpdaterProvider>
```

source/renderer/app/features/menu-updater/makeReactionCallback.spec.js
```diff
@@ -0,0 +1,212 @@
...
+      ...defaultArgs,
+      rebuildApplicationMenu: makeRebuildApplicationMenu({ send }),
+    })();
+
+    expect(send).toHaveBeenCalledTimes(1);
+    expect(send).toHaveBeenNthCalledWith(1, {
+      isNavigationEnabled: expect.any(Boolean),
+      walletSettingsState: expect.stringMatching(
+        new RegExp(
+          `(${WalletSettingsStateEnum.enabled}|${WalletSettingsStateEnum.disabled}|${WalletSettingsStateEnum.hidden})`
```
renanvalentin, line 65: Should we be more explicit on test assertion? Does it means that you are expecting any of these values? If so, you could refactor to use `test.each` and supply different params and outcomes.

renanvalentin:
Great job @szymonmaslowski ! I left a small suggestion related to usage of `ref` if is not complex to change. Thank you 🚀

alexander-rukin (approved):
Works well! 🎉

source/renderer/app/components/wallet/WalletSendForm.js
szymonmaslowski: The `open` method in this case is part of a public API of the `walletTokenPickerRef` component so I wouldn't call it private. It is just exposed through the ref API. The reason I resigned from local state is that I wanted to base the mechanic of opening/closing the dialog on the global `uiDialogs` state. Alternatively I could implement the open method on the closest parent container component and pass it down to here which is also fine for me. The reason I chose this approach is that it encapsulates all dialog related logic in the dedicated dialog component. Also I had the `WalletTokenPicker` component already imported there (it is needed for the uiDialogs' open function). However I understand it is not common to use ref apis. If this is problematic I am fine to refactor it to passing the `open` function from above.

source/renderer/app/containers/tokens/WalletTokenPickerDialogContainer.js
szymonmaslowski: Yep, it is possible, but if we chose the approach of passing down the `open`, `close` and `isOpen` functions then this component is redundant and we can use the `Picker` component directly. I just noticed in our codebase container components dedicated for dialogs and this is why I created this one.

source/renderer/app/features/menu-updater/MenuUpdater.js
szymonmaslowski: I didn't call it `context` since it does not provide anything through context so IMO it should not be called this way.

source/renderer/app/features/menu-updater/MenuUpdater.js
szymonmaslowski: I think it is common to call component a `Provider` if such component provides something through the react context. However I am happy to rename it if this is our rule for such components.

source/renderer/app/features/menu-updater/makeReactionCallback.js
szymonmaslowski: The problem is that I don't need the value of the `profile.currentLocale` (the main app obtains that value by itself), but I need to react on its changes. Event if this logic would be placed in the react component decorated with `observer` as I understand I also would need to do the same in order to tell the mobx that I want to react on changes of that property. But I could be wrong since I am new to mobx. I will play around :+1:

source/renderer/app/features/menu-updater/makeReactionCallback.js
szymonmaslowski: The `Reaction` class provides the api around the [mobx autorun feature](https://mobx.js.org/reactions.html) which is a way of making side effects on changes to the store so my case was a perfect use case for it. However you make a good point, using `useEffect` probably would simplify things in this case. Summarising: I don't have any feature to pass down the react tree through context and don't have to use the `autorun` and `Reaction` necessarily, so there is no point in calling this a Feature. I will use the `useEffect` instead of `Reaction` and move the implementation out of the features directory then.

source/renderer/app/features/menu-updater/makeReactionCallback.spec.js
szymonmaslowski: I was not aware of existence of `test.each`, it is cool :+1: However it is not the intension of this test case to test all of those values as all of them are tested bellow. My intention was to have a basic test case for ensuring the ipc channel communication is triggered by this functionality. I changed the `walletSettingsState` value to expect any string

source/renderer/app/containers/MenuUpdater/useMenuUpdater.spec.js
```diff
@@ -38,7 +40,15 @@ const makeRebuildApplicationMenu = ({
   send = () => {},
 } = {}): RebuildApplicationMenu => ({ send }: any);
 
-const defaultArgs: MakeReactionCallbackArgs = {
+const renderComponent = (args) => {
+  const Component = () => {
```
renanvalentin: Another suggestion is that you could test the `MenuUpdater` to have a better integration test. Currently you are only testing the internal implementation. 

```js
function renderComponent(spy) {
return <>
<MenuUpdaterProvider appStore={{}} profileStore={{}} router={{}} stakingStore={{}} uiDialogsStore={{}}>
 <MenuUpdater onUpdate={({isNavigationEnabled, walletSettingsState}) => spy({isNavigationEnabled, walletSettingsState})} />
</MenuUpdaterProvider>
}
```

gabriela-ponce:
@szymonmaslowski The scenarios were updated according to the new requirements you mentioned. And the latest fixes look good, great job 🚀 

I found this issue on build 20711 (doesn't reproduce on develop).

- Can't open the dialog "Native token settings". Check [this](https://drive.google.com/file/d/1j0w7449vEHEX6gLPPC_pQ-9SU5gMWO01/view) video.

source/renderer/app/containers/MenuUpdater/useMenuUpdater.spec.js
szymonmaslowski: The `MenuUpdater` component itself is not important as its role is only to provide required data using the mobx api. The sole logic is contained in the useMenuUpdater. Testing the MenuUpdater would require from me to involve mobx to testing which IMO is not worth for value it would give. I think we don't need to test things like connecting the component to mobx store

szymonmaslowski:
> @szymonmaslowski The scenarios were updated according to the new requirements you mentioned. And the latest fixes look good, great job 🚀
> 
> I found this issue on build 20711 (doesn't reproduce on develop).
> 
> * Can't open the dialog "Native token settings". Check [this](https://drive.google.com/file/d/1j0w7449vEHEX6gLPPC_pQ-9SU5gMWO01/view) video.

The change responsible for opening that dialog should have been missed during merging. It is reverted already and should be working fine now. Thanks @gabriela-ponce.

gabriela-ponce (approved):
Good job 🚀

renanvalentin (approved):
Great job! 🚀

PR #2884, opened 2022-02-18
[DDW-957] Fix random Buildkite failures and improve CI

marcin-mazurek:
This PR:

* Adds retry capability to Buildkite agent calls to ensure artifacts upload is not affected by intermittent network issues 
* Merges test and linter GH Actions pipelines into a single pipeline
* Removes Hydra linter and TSC checks (which are anyway covered by GitHub actions)
* Adds a `postinstall` hook that runs `yarn lockfile:fix` - only in local environment
* Introduces cache for node_modules in GitHub actions which should reduce the GH actions execution time by ~1.5 min

## Screenshots

Lockfile fixed automatically:

image

Lockfile fix skipped in CI:

image

Build fails when `lockfile:fix` is not run:

image

Build fails when `manage:translations` is not run:

Screenshot 2022-02-18 at 19 37 30

## TODO

- [ ] Daniel to remove `ci/hydra-build:tests.runTsc` GitHub check


.github/workflows/verify_pr.yml
```diff
@@ -0,0 +1,26 @@
+name: Verify Pull Request
```
marcin-mazurek, line 1: Here, I've merged the test pipeline with linter pipeline, which was introduced lately by @renanvalentin, as well as added a few more checks. I think this is as per our discussion to reduce the amount of work done by CI - when those two were running separately, we'd spend 75% of the time on installing `node_modules`. I'd like to start a discussion here and see which approach works better for us, separate, more granular pipelines, or one which checks everything?

.github/workflows/verify_pr.yml
```diff
@@ -0,0 +1,26 @@
...
+        uses: actions/checkout@v2
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: "14"
+      - name: Restore node_modules from cache
+        uses: actions/cache@v2
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
```
renanvalentin, line 18: Nicely done

scripts/utils.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+arg2nz() { test $# -ge 2 -a ! -z "$2" || usage "empty value for" "$1"; }
```
renanvalentin, line 3: @marcin-mazurek could you please add a comment or add func name because its not clear  what's this script is doing (plus i'm not expert on bash script 😅)

renanvalentin (changes requested):

scripts/utils.sh
DominikGuzei: This is transferring all your coins to him 😈

DominikGuzei (approved):

.github/workflows/verify_pr.yml
marcin-mazurek: Guys, at the moment unit tests take less than 30 seconds to execute. Any gain from parallelization will be negligible, especially that we're meant to run this locally before pushing. The only potential win would be by splitting `yarn check:all` but I feel this is too much outside of scope of this PR, and adds another layer of complexity as we would need to share files across the jobs (either by mounting a shared volume or by uploading and then downloading artifacts). I think such improvements definitely make sense but when the pipeline takes longer than 5 minutes to execute.

Anyway I feel this PR introduces a significant improvement in terms of reducing the build time by introducing cache for node_modules - installing packages now takes 15 seconds rather than 1.5 minutes:

image

scripts/utils.sh
marcin-mazurek: Good catch, although I just moved this function I've renamed it to something that makes more sense :) It essentially validates arguments in the following format `--build-id 20131` and ensures the second part is provided or calls the `usage` function that shows the script help.

package.json
```diff
@@ -62,7 +62,8 @@
     "yoroi:wallet:importer": "ts-node utils/api-importer/yoroi-wallet-importer.ts",
     "create-news-verification-hashes": "ts-node utils/create-news-verification-hashes/index.ts",
     "lockfile:check": "ts-node utils/lockfile-checker/index.ts --check",
-    "lockfile:fix": "ts-node utils/lockfile-checker/index.ts --fix"
+    "lockfile:fix": "ts-node utils/lockfile-checker/index.ts --fix",
+    "postinstall": "./scripts/postinstall.sh"
```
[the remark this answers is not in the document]
marcin-mazurek: Done

marcin-mazurek:
@DominikGuzei @szymonmaslowski @renanvalentin please re-review :)

renanvalentin (approved):

szymonmaslowski (approved):

PR #2939, opened 2022-03-30
[DDW-1062] Showcase filtering `Daedalus.json` logs with `jq`

michalrus:
*(internal PR)*

### Description

@danielmain suggested I create a simple Bash program to showcase how jq can be used for filtering logs

If you don’t have `jq` on `PATH`, one from Nixpkgs will be used.

Also, it will clean up and consume the development version of `Daedalus.json`, which is not compacted JSON entries separated by newlines (like the production version), but pretty-printed and with additional `[date] [context] ` before the object.

Please, look at the source of the filtering functions, to get an idea how to extend this to most common cases. Or use `jq` yourself, their query language is more/less intuitive.

### User Story

As a developer
I want to be able to filter Daedalus logs with jq
so that I only see the logs I’m interested in when developing.

### Usage examples

```
$ ./scripts/filter-logs.sh -h
Usage: ./filter-logs.sh [OPTION]...

  -h, --help                       Display help.
  -f, --follow                     Continue processing as the log file grows.
  -c, --compact                    Output compact instead of pretty-printed JSON.
  -i, --input <file>               Deadalus.json log location (default: /Users/mw/Library/Application Support/Daedalus Mainnet/Logs/pub/Daedalus.json).
  -m, --msg-contains <substring>   Filter down to entries ‘.msg’ of which contains a <substring>.
      --app-contains <string>      Filter down to entries ‘.app’ of which contains a <string> element.
  -d, --data-key <key>             Filter down to entries ‘.data’ of which contains a <key> key.
      --data-kv <key> <value>      Filter down to entries ‘.data’ of which contains a <key> with value <value>.

```

#### Continuously display only log entries that have `.data.state` defined:

```
$ ./scripts/filter-logs.sh -f --compact --data-key state
```

Screenshot-20220330-155839

#### Non-`--compact` version (pretty-printed):

Screenshot-20220330-155856


miorsufianiohk:
Note: No QA needed for this

danielmain (approved):
Great job 🤾‍♀ @michalrus !! This is extremely useful when troubleshooting. Thank you

michalrus:
Thank you!

PR #2960, opened 2022-04-14
Fix yarn.lock after release

michalrus:
_Internal PR to fix builds failing on `develop`._

/cc @danielmain


danielmain (approved):
Thanks @michalrus !

PR #2989, opened 2022-05-19
[DDW-809] Analytics - part 2 - Matomo client, router-level tracking and new Terms of Service

marcin-mazurek:
This PR introduces a Matomo client, implements router-level tracking with Matomo, example of component-level tracking for stake pool view type change, as well as updates the Terms of Service.

This PR should not be tested by QA team as it does not include full tracking - which is delivered in the third PR.

First part: https://github.com/input-output-hk/daedalus/pull/2927
Third part: https://github.com/input-output-hk/daedalus/pull/3003

## Screenshots

No UI changes except for TOS content.

## Testing Checklist

- [Slack QA thread](https://input-output-rnd.slack.com/archives/GGKFXSKC6/p1654839266665729)
- [ ] Test

---

### Basics

- [x] PR assigned to the PR author(s)
- [x] `input-output-hk/daedalus-dev` and `input-output-hk/daedalus-qa` assigned as PR reviewers
- [x] If there are UI changes, Alexander Rukin assigned as an additional reviewer
- [x] All visual regression testing has been reviewed (assign `run Chromatic` label to PR to trigger the run)
- [x] PR has appropriate labels (`release-vNext`, `feature`/`bug`/`chore`, `WIP`)
- [x] PR link is added to a Jira ticket, ticket moved to In Review
- [ ] PR is updated to the most recent version of the target branch (and there are no conflicts)
- [x] PR has a good description that summarizes all changes
- [x] PR contains screenshots (in case of UI changes)
- [x] CHANGELOG entry has been added to the top of the appropriate section (_Features_, _Fixes_, _Chores_) and is linked to the correct PR on GitHub
- [x] There are no missing translations (running `yarn manage:translations` produces no changes)
- [x] Text changes are proofread and approved (Jane Wild / Amy Reeve)
- [x] Japanese text changes are proofread and approved (Junko Oda)
- [x] Storybook works and no stories are broken (`yarn storybook`)
- [x] In case of dependency changes `yarn.lock` file is updated

### Code Quality

- [ ] Important parts of the code are properly commented and documented
- [ ] Code is properly typed with typescript types
- [ ] React components are split-up enough to avoid unnecessary re-renderings
- [ ] Any code that only works in main process is neatly separated from components

### Testing

- [ ] New feature/change is covered by acceptance tests
- [ ] New feature/change is manually tested and approved by QA team
- [ ] All existing acceptance tests are still up-to-date
- [ ] New feature/change is covered by Daedalus Testing scenario
[...]

PR #2927, opened 2022-03-21
[DDW-809] Analytics - part 1 - analytics opt-in, settings and unique user ID generation

marcin-mazurek:
This PR introduces Matomo as analytics platform. It does not cover tracking user actions.

This PR should not be tested by QA team as it does not include tracking itself - which is delivered in the second and third PR.

Second part: https://github.com/input-output-hk/daedalus/pull/2989
Third part: https://github.com/input-output-hk/daedalus/pull/3003

## Todos

- [ ] Set appropriate Matomo instance URL in analytics config
- [ ] Set appropriate Matomo instance domain pattern in CSP headers
- [x] Check feature parity between self-hosted and managed versions of Matomo
- [x] Consider rewriting storing analytics reference with Context Provider rather than MobX
- [x] Generate some traffic to get new dashboards
- [x] Separate dev instance
- [x] Send initial event immediately after user agrees to analytics from settings page
- [x] Final messages from legal team
- [x] Japanese translations

## Screenshots

Screenshot 2022-03-21 at 21 50 41
Screenshot 2022-06-20 at 23 24 56
Screenshot 2022-06-20 at 23 25 02
Screenshot 2022-06-20 at 23 25 09
Screenshot 2022-06-20 at 23 27 31
Screenshot 2022-06-20 at 23 28 46

## Testing Checklist

- [Slack QA thread](https://input-output-rnd.slack.com/messages/GGKFXSKC6)
- [ ] Test

---

### Basics

- [x] PR assigned to the PR author(s)
- [x] `input-output-hk/daedalus-dev` and `input-output-hk/daedalus-qa` assigned as PR reviewers
- [x] [N/A] If there are UI changes, Alexander Rukin assigned as an additional reviewer
- [x] [N/A] All visual regression testing has been reviewed (assign `run Chromatic` label to PR to trigger the run)
- [X] PR has appropriate labels (`release-vNext`, `feature`/`bug`/`chore`, `WIP`)
- [X] PR link is added to a Jira ticket, ticket moved to In Review
- [ ] PR is updated to the most recent version of the target branch (and there are no conflicts)
- [X] PR has a good description that summarizes all changes
- [X] PR contains screenshots (in case of UI changes)
- [x] CHANGELOG entry has been added to the top of the appropriate section (_Features_, _Fixes_, _Chores_) and is linked to the correct PR on GitHub
- [x] [N/A] There are no missing translations (running `yarn manage:translations` produces no changes)
[...]

PR #3003, opened 2022-06-15
[DDW-809] Analytics - part 3 - detailed tracking

marcin-mazurek:
This PR introduces detailed tracking across all Daedalus user journeys.

This PR should be tested by the QA team.

First part: https://github.com/input-output-hk/daedalus/pull/2927
Second part: https://github.com/input-output-hk/daedalus/pull/2989

## Screenshots

No UI changes

## Testing Checklist

- [Slack QA thread](https://input-output-rnd.slack.com/archives/GGKFXSKC6/p1654839266665729)
- [ ] Test

---

### Basics

- [x] PR assigned to the PR author(s)
- [x] `input-output-hk/daedalus-dev` and `input-output-hk/daedalus-qa` assigned as PR reviewers
- [x] If there are UI changes, Alexander Rukin assigned as an additional reviewer
- [x] All visual regression testing has been reviewed (assign `run Chromatic` label to PR to trigger the run)
- [x] PR has appropriate labels (`release-vNext`, `feature`/`bug`/`chore`, `WIP`)
- [x] PR link is added to a Jira ticket, ticket moved to In Review
- [ ] PR is updated to the most recent version of the target branch (and there are no conflicts)
- [x] PR has a good description that summarizes all changes
- [x] PR contains screenshots (in case of UI changes)
- [x] CHANGELOG entry has been added to the top of the appropriate section (_Features_, _Fixes_, _Chores_) and is linked to the correct PR on GitHub
- [x] There are no missing translations (running `yarn manage:translations` produces no changes)
- [x] Text changes are proofread and approved (Jane Wild / Amy Reeve)
- [x] Japanese text changes are proofread and approved (Junko Oda)
- [x] Storybook works and no stories are broken (`yarn storybook`)
- [x] In case of dependency changes `yarn.lock` file is updated

### Code Quality

### Testing

### After Review


source/main/environment.ts
```diff
@@ -43,7 +43,7 @@ const isShelleyQA = checkIsShelleyQA(NETWORK);
 const isSelfnode = checkIsSelfnode(NETWORK);
 const isDevelopment = checkIsDevelopment(NETWORK);
 const isWatchMode = process.env.IS_WATCH_MODE;
-const analyticsFeatureEnabled = isMainnet || isStaging || isTestnet;
+const analyticsFeatureEnabled = true;
```
marcin-mazurek, line 46: Set to `true` to include self-node support.

source/renderer/app/App.tsx
```diff
@@ -65,46 +70,53 @@ class App extends Component<{
...
-                  <RTSFlagsRecommendationOverlayContainer />
-                  <NotificationsContainer />
-                  {canShowNews && [
-                    <NewsFeedContainer key="newsFeedList" />,
-                    <NewsOverlayContainer key="newsFeedOverlay" />,
-                  ]}
-                </Fragment>
-              </IntlProvider>
-            </ThemeProvider>
+            <LocalStorageFeatureProvider localStorage={LocalStorageApi}>
```
marcin-mazurek, line 73: Relocated this from `index.tsx` file for consistency, and so DiscreetModeFeatureProvider can use AnalyticsProvider.

source/renderer/app/features/discreet-mode/context.tsx
```diff
@@ -18,17 +20,31 @@ interface Props {
...
     window.daedalus = merge(window.daedalus, {
       features: {
         discreetModeFeature: feature,
       },
     });
+
     return feature;
   });
+
+  useEffect(() => {
```
marcin-mazurek, line 40: Required because AnalyticsClient implementation can chance in the runtime.

source/renderer/app/stores/WalletsStore.ts
```diff
@@ -1508,6 +1541,11 @@ export default class WalletsStore extends Store {
       this.actions.wallets.generateAddressPDFSuccess.trigger({
         walletAddress,
       });
+      this.stores.analytics.analyticsClient.sendEvent(
+        'Wallets',
```
shawnbusuttil, line 1545: Yeah we'd benefit from const variables here. :)

marcin-mazurek:
@gabriela-ponce in terms of the issues you've raised:

> The actions involving switching on/off discreet mode are not showing up in matomo. Check [this video](https://drive.google.com/file/d/14N1FRhXH0JIEEy5P8ODXbEXpR0BpN3NM/view?usp=sharing).

Fixed, please check.

> When saving the address' QR from a Hardware Wallet, the action says "software wallet". Check [this](https://drive.google.com/file/d/1dkvcmxmqa6-Rxhc9aL2SonvyVgGsdjl6/view?usp=sharing) screenshot (you can see the previous step is what is required for HW to have these options enabled).

Fixed, please check.

> The displayed version for macOS is incorrect. Check [this](https://drive.google.com/file/d/16UZ0oFPqPaTp2KP7XaAhfH_GysL7yhhY/view?usp=sharing) screenshot.

That is a tricky one:

https://www.sindastra.de/p/2378/why-the-user-agent-string-is-wrong-on-mac
https://chromestatus.com/feature/5452592194781184

If we really want to fix it, I would suggest one of the following:

1) **My recommendation**: we can pass more detailed OS information to the custom dimension called OS. At the moment we only pass the operating system through it, which anyway feels redundant since Matomo handles OS detection on its own, by parsing the user agent header.
2) Alternative, more hacky, more complex, but more user friendly approach (using native Matomo capability to collect OS data): we can 'fix' user agent by explicitly setting it for macOS devices, so right OS version is sent to Matomo. Please note that Matomo Tracker JS library does not support this, so we would have to rewrite it in vanilla JS.

source/renderer/app/stores/WalletsStore.ts
shawnbusuttil: If they are unique, at least we can keep a reference of these actions somewhere.

But, the question is: is an action's name determined by the author and will it always be unique at all times? I'm guessing there's a reference on the other side that knows which actions to listen to, in which case it's better to keep a reference.

marcin-mazurek: Matomo doesn't work like PubSub, it doesn't need to know what are the correct action names. It accepts whatever values we push to its API. Here is an example dashboard it renders based on the data provided:

image

In order to render such a dashboard, the only thing you need to configure in Matomo is some basic site config (title, URL, etc.).

It acts pretty much as logger - the main difference is that it is not used for troubleshooting but to understand the user behaviour and improve the UX. And I've never seen any codebase where log messages would be extracted to an enum or constants, normally they are just plain inline strings. 

I personally believe it makes sense to extract event categories to an enum to minimise a risk of a typo which would then result in two categories which in fact should be one, but for action names which are unique I would keep them as is.

shawnbusuttil: Then let's go with just categories. 👍

marcin-mazurek: Cool, I extracted categories to an enum called `EventCategories` earlier.

gabriela-ponce:
> @gabriela-ponce in terms of the issues you've raised:
> 
> > The actions involving switching on/off discreet mode are not showing up in matomo. Check [this video](https://drive.google.com/file/d/14N1FRhXH0JIEEy5P8ODXbEXpR0BpN3NM/view?usp=sharing).
> 
> Fixed, please check.
> 
> > When saving the address' QR from a Hardware Wallet, the action says "software wallet". Check [this](https://drive.google.com/file/d/1dkvcmxmqa6-Rxhc9aL2SonvyVgGsdjl6/view?usp=sharing) screenshot (you can see the previous step is what is required for HW to have these options enabled).
> 
> Fixed, please check.
> 
> > The displayed version for macOS is incorrect. Check [this](https://drive.google.com/file/d/16UZ0oFPqPaTp2KP7XaAhfH_GysL7yhhY/view?usp=sharing) screenshot.
> 
> That is a tricky one:
> 
> https://www.sindastra.de/p/2378/why-the-user-agent-string-is-wrong-on-mac https://chromestatus.com/feature/5452592194781184
> 
> If we really want to fix it, I would suggest one of the following:
> 
> 1. **My recommendation**: we can pass more detailed OS information to the custom dimension called OS. At the moment we only pass the operating system name, which anyway feels redundant since Matomo handles OS detection on its own, by parsing the user agent header.
> 2. Alternative, more hacky, more complex, but more user friendly approach (using native Matomo capability to collect OS data): we can 'fix' user agent by explicitly setting it for macOS devices, so right OS version is sent to Matomo. Please note that Matomo Tracker JS library does not support this, so we would have to rewrite the MatomoClient in vanilla JS (shouldn't be a massive effort, but it's not 5 minutes either).

@marcin-mazurek the fixes for the first 2 items are ok. I'll recheck this PR if there are further changes.
Just FYI, the latest 2 builds failed, so I couldn't check the changes for the multi-signature key. 
Regarding the OS issue, not sure which one is the best approach, but we can ask @daniloprates if this data needs to be collected as well or we can ignore it for now.

marcin-mazurek:
@miorsufianiohk the label 'Wallet deleted' is in line with our user interface labels:

image

marcin-mazurek:
OK, I get what you're saying now. So we show 'Delete wallet' for software wallets and 'Unpair wallet' for hardware wallets. I think I would anyway recommend to keep the same event name, as this is the same action under the hood, and separating those would make the dashboards less readable (users will see Delete Wallet and Unpair Wallet separately, it might be interpreted as two different actions).

marcin-mazurek:
@miorsufianiohk Check out this screenshot from Matomo and look how we can nicely drill down the event by the wallet type:

image

dmitrii-gaico:
@marcin-mazurek Hello! Just founds another issue in logic. So the problem is If you open "analytics settings" from the app settings and close the app while dialog is active then on the next start-up the dialog would be displayed as a first screen you see. Please see the [record](https://drive.google.com/file/d/1_k7r7aknhzF0mn44i2D32lsPKk6R7UOH/view?usp=sharing) for details. So far confirmed at Windows but issue should not be platform specific.

marcin-mazurek:
@dmitrii-gaico with regards to the first issue - sure, I'll add appropriate margins/paddings to ensure this doesn't happen.

With regards to the second issue, we reset user analytics settings when the modal is opened, so if at this point the user quits the app, they will then see the analytics dialog again.

I didn't pay much attention it and didn't consider this a bug, as I thought it is very unlikely that the user behaves the way you described, and even if it happens - it is like reopening the browser and seeing the previously open page(s). Having said that this would be inconsistent with the rest of the app, where we don't have such a feature.

But it's not a problem - I can change it, do you think we need to confirm that with @daniloprates?

dmitrii-gaico:
> @dmitrii-gaico with regards to the first issue - sure, I'll add appropriate margins/paddings to ensure this doesn't happen.
> 
> With regards to the second issue, we reset user analytics settings when the modal is opened, so if at this point the user quits the app, they will then see the analytics dialog again.
> 
> I didn't pay much attention it and didn't consider this a bug, as I thought it is very unlikely that the user behaves the way you described, and even if it happens - it is like reopening the browser and seeing the previously open page(s). Having said that this would be inconsistent with the rest of the app, where we don't have such a feature.
> 
> But it's not a problem - I can change it, do you think we need to confirm that with @daniloprates?

Hey! Let me describe a little bit issue #2 and how I see it.
So from my point of view the friendly logic would be : 
1. If you already selected any option allow/not allow then we should not show this dialog again under any circumstances unless user navigated to the settings and wants to do some changes.
2. When we start the application we have a strict logic about what should be displayed initially and this is "Start up" screens (Starting node / Sync % Messages) and I feel those screen are more important to show rather then showing the Analytics Settings as they are representing the current status of the application. Currently for latest build Analytics Settings are displayed instantly and you would not see the Start Up screen until you do select Allow/Not allow.
3. Even if user is not doing the steps I did it might happen that Daedalus is crashed by any reason while Analytics Settings are opened and then we are returning to the point #2.

I would correct myself and say that this is more like a logic improvement rather than the issue. 
From my side I would keep the startup logic as is in PROD and would not add any other screens to be displayed before those. So I believe we don't need @daniloprates for confirmation here.

marcin-mazurek:
@dmitrii-gaico @gabriela-ponce @miorsufianiohk ready for another round of review

shawnbusuttil (approved):

dmitrii-gaico (approved):

source/renderer/app/analytics/types.ts
```diff
@@ -1,10 +1,26 @@
...
+}
+
+export enum EventCategories {
+  WALLETS = 'Wallets',
+  STAKE_POOLS = 'Stake Pools',
+  SETTINGS = 'Settings',
+  LAYOUT = 'Layout',
+  SYSTEM_MENU = 'System Menu',
+  VOTING = 'Voting',
+}
```
przemyslaw-wlodek, line 26: It's a little bit inconvenient to have string-based enum with different keys/values. Often there's a situation when you need to check property existence by it's value (e.g. coming from an API) and then you need to use `Object.values` for that, instead of simple `SomeEnum[value]`.

There were some other reasons why string-based enums with 1:1 mapping are preferable but I don't remember them tbh. Preferably we could have a constant that does the mapping (based on that enum).

source/renderer/app/features/discreet-mode/feature.ts
```diff
@@ -50,6 +51,12 @@ export class DiscreetMode extends Feature {
   @action
   toggleDiscreetMode = () => {
     this.isDiscreetMode = !this.isDiscreetMode;
+    this.analyticsTracker.sendEvent(
+      EventCategories.SETTINGS,
+      this.isDiscreetMode
+        ? 'Turned on discreet mode'
```
przemyslaw-wlodek, line 57: Should we create some constants for the event names? Sounds fair, easier to maintain and less prone to typos.

source/renderer/app/stores/CurrencyStore.ts
```diff
@@ -127,10 +128,23 @@ export default class CurrencyStore extends Store {
...
   };
   @action
   _toggleCurrencyIsActive = () => {
     this.isActive = !this.isActive;
     this.api.localStorage.setCurrencyIsActive(this.isActive);
+
+    this.analytics.sendEvent(
+      EventCategories.SETTINGS,
+      `Turned ${
+        this.isActive ? 'on' : 'off'
```
przemyslaw-wlodek, line 146: In this place you're doing ternary in string interpolation, in the other it's string or string (duplicating the text) 😅

I still recommend some constant/s for event names.

przemyslaw-wlodek (changes requested):

gabriela-ponce (approved):
Great job 🎉

source/renderer/app/analytics/MatomoClient.ts
```diff
@@ -40,18 +39,23 @@ export class MatomoClient implements AnalyticsClient {
     });
   };
 
-  sendEvent = async (category: string, action: string): Promise<void> => {
+  sendEvent = async (
+    category: string,
+    action: string,
+    name?: string
+  ): Promise<void> => {
```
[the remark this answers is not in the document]
marcin-mazurek: In general I'd agree, however in this case the third parameter is optional and is not used in 90% cases. I think it would make sense if we have more than 3 parameters or various combinations of parameters.

source/renderer/app/features/discreet-mode/feature.ts
marcin-mazurek: I've already discussed it with Shawn: https://github.com/input-output-hk/daedalus/pull/3003#discussion_r907583452

Let me know if you'd want me to introduce constants for event names anyway.

source/renderer/app/analytics/types.ts
marcin-mazurek: I see your point, however this is not data coming from an API, but human-readable labels. Personally, I feel that having an enum to define possible categories and object literal for mapping is an overkill. I checked TS documentation and there are no recommendations to keep the key and value of enum the same: https://www.typescriptlang.org/docs/handbook/enums.html 

I could maybe change it to object literal with `as const` if you think that makes things any clearer?

source/renderer/app/stores/CurrencyStore.ts
marcin-mazurek: Yes, that's something I overlooed. Fixed.

source/renderer/app/analytics/types.ts
marcin-mazurek: @szymonmaslowski found another blog post which shows plenty of examples with non 1:1 mapping of enum key/values: https://blog.logrocket.com/writing-readable-code-with-typescript-enums-a84864f340e9/

przemyslaw-wlodek: Yep, I've seen this article and I can't find a real reason not to use it if that's not coming from an API. Let's leave it :)
I just wanted to have this discussion. Thanks 😁

source/renderer/app/features/discreet-mode/feature.ts
przemyslaw-wlodek: As discussed offline 🙂 In future we might need event versioning or something else but for now let's keep it YAGNI way. Thanks

przemyslaw-wlodek (approved):
I'm happy to see it finally merged 😄 Good job @marcin-mazurek!

PR #3010, opened 2022-07-08
Update `cardano-node` to 1.35.1

michalrus:
Internal PR (also patching `cardano-wallet`, just from here)


danielmain (approved):
Thank you @michalrus

PR #3013, opened 2022-07-11
Update `cardano-node` to 1.35.1 via `cardano-wallet` PR `.patch`

michalrus:
… which also includes `cardano-ledger` update.

*Internal*


danielmain (approved):
Thank you @michalrus

PR #3014, opened 2022-07-11
Really cache the devshell setup in `$out` (and Hydra)

michalrus:
*Internal PR*

It often happens that developers have to rebuild the world when entering `nix-shell`.

It turns out we don’t cache the devshell environment in https://hydra.iohk.io/ 

This PR fixes that: instead of just `touch`ing `$out`, we need to `export` the environment into it.

Cf. e.g. https://github.com/NixOS/nixpkgs/blob/51ae0c2a8cdfc729fe0ed68f630d12d5e3f6594d/pkgs/build-support/mkshell/default.nix#L53


michalrus:
Closing, as it's already in `develop`

PR #3031, opened 2022-08-30
[DDW-1134] Fix Catalyst registration with hardware wallets

marcin-mazurek:
This PR fixes Catalyst registration with hardware wallets as well as prevents the application from crashing in dev mode, when cancelling saving Catalyst PDF in the native OS file window. 

## Testing Checklist

- [Slack QA thread](https://input-output-rnd.slack.com/archives/GGKFXSKC6/p1661865963222239)
- [ ] Test

---

### Basics

- [x] PR assigned to the PR author(s)
- [x] `input-output-hk/daedalus-dev` and `input-output-hk/daedalus-qa` assigned as PR reviewers
- [x] If there are UI changes, Alexander Rukin assigned as an additional reviewer
- [x] All visual regression testing has been reviewed (assign `run Chromatic` label to PR to trigger the run)
- [x] PR has appropriate labels (`release-vNext`, `feature`/`bug`/`chore`, `WIP`)
- [x] PR link is added to a Jira ticket, ticket moved to In Review
- [x] PR is updated to the most recent version of the target branch (and there are no conflicts)
- [x] PR has a good description that summarizes all changes
- [x] PR contains screenshots (in case of UI changes)
- [ ] CHANGELOG entry has been added to the top of the appropriate section (_Features_, _Fixes_, _Chores_) and is linked to the correct PR on GitHub
- [x] There are no missing translations (running `yarn manage:translations` produces no changes)
- [x] Text changes are proofread and approved (Jane Wild / Amy Reeve)
- [x] Japanese text changes are proofread and approved (Junko Oda)
- [x] Storybook works and no stories are broken (`yarn storybook`)
- [x] In case of dependency changes `yarn.lock` file is updated

### Code Quality

### Testing

### After Review


source/renderer/app/stores/VotingStore.ts
```diff
@@ -423,23 +423,21 @@ export default class VotingStore extends Store {
...
+      desktopDirectoryPath,
+      network,
+      isMainnet,
+      intl,
+    });
+
+    if (wasSaved) {
       this.actions.voting.saveAsPDFSuccess.trigger();
-    } catch (error) {
-      throw new Error(error);
```
DominikGuzei, line 442: Is removing the error handling safe here? How would the user know that an error occurred?

marcin-mazurek: This try-catch block doesn't do anything. If you're catching the error, not doing anything with it, and immediately rethrowing it, then the effect is exactly the same as not having the try-catch block at all :)

marcin-mazurek: It would make sense to keep this try-catch block if we were throwing a more concrete error instead or providing a different error message, but as it stands it's just adding noise :)

DominikGuzei: That's true indeed 👍

DominikGuzei: although you could also say the extra return values added noise 😋 you could just have thrown in the `votingPDFGenerator` instead of returning false … btw. I think `true` and `false` are not necessarily great return values for "did it work?"

source/renderer/app/containers/voting/dialogs/VotingRegistrationDialogContainer.tsx
```diff
@@ -269,30 +273,32 @@ class VotingRegistrationDialogContainer extends Component<Props, State> {
...
           address: address.id,
           amount,
           metadata: votingData.metadata,
-        }));
+        });
+
+        updateTxSignRequest(coinSelection);
       } else {
-        ({ fee } = await calculateTransactionFee({
+        coinSelection = await calculateTransactionFee({
```
renanvalentin, line 292: i don't think `coinSelection` is the appropriate name in this context.  TransactionStore.calculateTransactionFee returns a different data structure.

renanvalentin (approved):

source/renderer/app/stores/VotingStore.ts
DominikGuzei: or let's re-phrase my last comment: Boolean would be a good return value if the function just returned the information about success (eg. `wasPDFSaved()`). But since this function has side effects (is doing something, not just returning the status), I think throwing is also reasonable

marcin-mazurek: The booleans are actually not indicating whether it worked or failed. It indicates whether the user actually saved the file or cancelled the process using the "Cancel" button. But I can see it is confusing, see my latest commit - would enum be more readable?

DominikGuzei: Ah got it, so then throwing is less reasonable 👍  … yes I guess enums could improve the readability of return value

source/renderer/app/containers/voting/dialogs/VotingRegistrationDialogContainer.tsx
marcin-mazurek: Good catch, I got confused by destructuring which looked like we're dealing with the same model. Changed now.

ManusMcCole (approved):
Fix looks good 🚀

PR #3273, opened 2025-01-27
fix: vote delegation form

szymonmaslowski:
This PR
 - fixes the vote delegation form for wallets which had both: vote and pool delegation. With the recent Cardano Wallet update the form was not working properly for those wallets
 - enables HW based wallets to use cip129 drep format


hegaleon (approved):
Clear determination of script hash v id , using decoded hash to populate scriptHashHex and keyHashHex respectively

przemyslaw-wlodek (approved):

DominikGuzei (approved):
great catch 👏

PR #3310, opened 2026-04-10
Daedalus 7.3.0

AndrewWestberg:
This merges all of the maintenance work done for release 7.3.0. Main functionality added was the mithril phase 1 support for fast bootstrap.


michalrus:
Hey, have you checked that the auto-update still works? See [this Slack thread](https://input-output-rnd.slack.com/archives/C0AFL3H43EG/p1773313082262669) from **2026-03-12**:

- #3294 updated `cardano-shell`, because the old (ancient?) version was breaking CI;
- so a test of the whole auto-update flow needs to be performed before the next release
- How I did this before, I run `python -m http.server 12345` with new Daedalus installers for each operating systems, and a fake newsfeed that pointed `softwareUpdate` to these `localhost:12345` URLs,
- This is quite important because if anything is broken, and we **brick auto-update**, all users will have to manually download the next version from the website

The most important point is that we have to test the entire chain:
- that this version auto-updates correctly (to the fake one, say 7.98.0), 
- and the fake 7.98.0 one also updates to the next even faker one, 7.99.0

… on all three platforms (Linux, Windows, macOS)

Because when it comes to auto-update, in the _current_ release we influence the next 2 versions, unfortunately.

disassembler (approved):

PR #3312, opened 2026-04-13
Fix chain storage path handling and enhance validation

david-profrontsolutions:
Enhance the validation and handling of chain storage paths, including recovery fallbacks and notifications for existing data. Improve user experience by providing clearer guidance during storage location selection. Fix edge case where mithril snapshot failure leads to wiping data of the user's chosen external directory, which could include user data. Edge case fix creates a separate chain folder so user data is preserved at target location.


source/main/utils/chainStorageValidation.ts
```diff
@@ -119,40 +123,88 @@ export async function validateChainStorageDirectory(
       };
     }
 
+    const isDirectChainSelection =
```
AndrewWestberg, line 139: We now pass `currentCustomPath` into validation, but it is never used here. That means selecting a directory inside the currently managed chain, for example `/mnt/external/chain/db`, still validates successfully and `setDirectory()` later rewires the entry point to `/mnt/external/chain/db/chain`. The real chain data stays behind and we end up with a nested managed layout. Should this reject selections under the existing managed chain, except for the managed child itself being canonicalized back to its parent?

source/main/utils/chainStorageManagerLayout.ts
```diff
@@ -0,0 +1,597 @@
...
+
+export async function ensureManagedChainLayout(
+  ctx: ChainStorageManagerContext,
+  options: EnsureManagedLayoutOptions = {}
+): Promise<ManagedChainLayoutResult> {
+  await ctx._recoverInterruptedMigration(options);
+  await cleanupLegacyConfigIfPresent(ctx);
+  const entryPointState = await ctx._captureChainPathState();
+
+  if (entryPointState.type !== 'symlink') {
```
AndrewWestberg, line 31: This fallback looks risky now that `customPath` is no longer persisted anywhere. If `state/chain` is deleted or broken but the external `<custom>/chain` directory still exists, `getConfig()` can no longer recover the custom root and this branch immediately recreates the default chain directory instead. On `master` the config file still gave us a fallback source for the custom location, so this looks like a regression that can orphan valid external chain data and force an unnecessary resync.

source/main/utils/handleDiskSpace.ts
```diff
@@ -617,6 +793,140 @@ export const handleDiskSpace = (
...
+  };
+
+  const startDiskSpaceCheckRun = ({
+    hadNotEnoughSpaceLeft,
+    forceDiskSpaceRequired,
+  }: {
+    hadNotEnoughSpaceLeft?: boolean;
+    forceDiskSpaceRequired?: number;
+  }): Promise<CheckDiskSpaceResponse> => {
+    const diskSpaceCheckPromise = runHandleCheckDiskSpace(
```
AndrewWestberg, line 843: `runHandleCheckDiskSpace()` can now rethrow `MANAGED_CHAIN_LAYOUT_ERROR` from `ensureMithrilStartupGate()`, and this wrapper intentionally preserves that rejection for callers. The startup path in `index.ts` wants that, but the background `setInterval` caller still awaits `handleCheckDiskSpace()` without any containment. That means later layout / permission / migration failures during polling can surface as unhandled promise rejections in the main process. Can we make sure the interval-driven path catches and logs this while preserving the hard failure for the explicit startup call?

source/main/utils/chainStorageManagerLayout.ts
david-profrontsolutions: I changed this behavior on purpose. The comparison given is from the initial development code for the chain storage location view that was merged into master. 

The main reason I updated the validation UX was to reduce issues with having an additional config file that doesn't actually provide much value for the user or the app. It could also cause discrepencies and logic/context flow errors if future updates were made referencing this config file based on the users app state. This simplifies the logic to just default to the new chain/mithril flow. It is up to the user to understand/fix their external drive if disconnected or not available, not for the app to keep configs of their setups. 

This never existed in the current release of the app so it was additional file/config bloat that would have been added. If the users symlink is disconnected/broken they will come back to the initial chain storage location decision and decide if they need to reconnect their drive. A message is even provided to them stating that their symlink is broken.

source/main/utils/chainStorageValidation.ts
david-profrontsolutions: Everything lives in `chain` in the first place. The `db` reference would not even exist by the time the mithril process is complete, and is otherwise deleted if a failure occurs.

If the user does select a folder underneath an existing external `chain` directory like `immutable`, then yes in that edge case it will recreate and treat the `immutable` folder as the new source of truth. However, I find that case rather rare at this time, and more of a user error that will be easily identifiable. We can update if we need additional user defense logic, but otherwise I think it would create additional bloat to handle all possible user interaction decisions that don't significanlty reduce their UX.

I will clean up the unused typing and `options` param here.

source/main/utils/handleDiskSpace.ts
david-profrontsolutions: I will look into adding some hardening on this issue so it properly logs in event of error and doesn't result in failure issues for the user.

david-profrontsolutions:
## PR Update

The last commit update adds a final round of hardening on top of the chain-storage / Mithril path handling work for clean up and to address previous PR comments:

- Tightened destructive storage guards so chain-storage mutations only run when cardano-node is fully stopped, including rejecting the stopping state.
- Hardened validation and rollback around managed chain paths, including nested managed-child rejection, default-path alias handling, and broken symlink rollback.
- Improved Mithril IPC resilience by keeping listener setup idempotent, rebinding status delivery to the latest window after recreation, and making renderer send failures non-blocking for internal status listeners.
- Improved storage-picker safety by preserving the exact rejected path, disabling continue on invalid draft selections, surfacing apply/reset failures cleanly, and serializing the return-to-storage flow.
- Added an early selfnode config check so chain storage is not mutated when config creation cannot proceed.
- Updated the hardening summary and changelog to match the current runtime behavior.

### Verification

- Targeted Jest suites for chain storage, Mithril IPC/service flows, storage picker behavior, selfnode config handling, and disk-space coordination pass locally.

AndrewWestberg (approved):

PR #3333, opened 2026-05-28
Mithril partial sync

AndrewWestberg:
Adds a diagnostic window button to use mithril partial sync to allow for fast blockchain catch up.

After this merge, it will need a UI/UX pass to improve things.


nix/internal/launcher-config.nix
```diff
@@ -406,6 +406,7 @@ let
           legacySecretKey
           ;
         wipeChain = false;
+        mithrilPartialSyncEnabled = true;
```
david-profrontsolutions: Should this be disabled until we make the UI/UX cleanup pass? I don't think we want this released if we have this merged in and release with other fixes.

david-profrontsolutions (approved):
## Good backbone for Mithril Partial Sync

### Approving contingent on:
- The merge conflict being fixed
- Mithril Partial Sync flag set to `false`, to be disabled until UX clean-up

I only ran through a critical path working flow. Further testing will need to be done with the UI/UX clean-up.

---
### Potential bugs/issues

- Diagnostic Page: Once a Mithril Partial Sync was completed the Mithril sync option was disabled. Copy was present that stated Mithril partial sync couldn't be started because the current partial sync had not completed.

- After a successful Mithril Partial Sync, I closed the Daedalus Mainnet app, the cardano node took several minutes to stop. I then opened the app which led to the verifying on-disk blockchain state process. I thought this process gets skipped if the Cardano Node process is shut down gracefully? Will need to investigate further if connected to Mithril Partial Sync changes or existing Cardano Node issue.

---
### UI/UX items to add to a new issue ticket

There are a few initial UI/UX items to integrate before having this feature flag turned on. Most likely better to go through the UI/UX clean-up before going through any further manual testing path.

- English and Japanese copy review and polish.

- Diagnostic Page: Mithril Partial Sync copy needs to be cleaned up.

- Diagnostic Page: Partial Sync UI/UX clean up for user acknowledgment message.

- Diagnostic Page: Should look at simply disabling the partial sync option if sync status is closer than a certain percentage (< 5%).

- Mithril Partial Sync: Cardano Node stopping process needs refinement. Partial Sync Dialogue was not showing any updates, with elapsed time as 0:00 and initial greyed out steps. Cardano Node stopping is seen in the opaque background. This needs to be presented in the dialogue, with possibly more clarity on the cardano node stopping process. Process took a couple minutes to resolve, cancel button was clicked but didn't respond, continued to move onto the Mithril Partial Sync once the node stopped.

- Mithril Partial Sync: Elapsed Time doesn't update, during Finalizing -converting phase it was about 2:19, time only updates as each next step finishes, 6:46 was the final value even though more conversion and file activity was present. (seen on Windows Mainnet)

- Mithril Partial Sync: Progress circles for each step are not rotating, existing issue seen on initial Mithril Sync (seen on Windows Mainnet)

- Mithril Partial Sync: I did see an overlay error via the nix developer environment, could have been a bad state or other development environment issue. (seen on Linux Preprod)

nix/internal/launcher-config.nix
AndrewWestberg: fixed

PR #3337, opened 2026-07-02
Enhance Mithril Partial Sync UX

david-profrontsolutions:
Summary

Refines the UX of the Mithril partial-sync feature (follow-up to #3333) across a 29-commit sprint (task-ux-001 → 702). ~80 source files changed (+8.0k/−0.9k), plus Storybook stories, unit/e2e coverage, translations, and .agent planning docs.

Highlights

- Availability & behind-ness model — new partial-sync availability read model + IPC channel; cached certified-immutable-gap signal; UI gated on availability. User-facing "how far behind" is now expressed in epochs only, computed in the renderer from node tip.
- Decision & prompt flow — proactive Mithril fork prompt on the syncing screen with inline confirm, and a behind-ness decision modal replacing in-place confirmation.
- Progress overlay — download progress driven by file counts, live-activity affordances, a dedicated error-copy map with distinct cancelled copy, hardened recovery actions, and success-dismiss finalize wiring.
- Startup & recovery correctness — staging co-located on the chain volume with a disk-space preflight, node-start-verified marker on the success path, status always re-emitted on cancel (no stuck "stopping node" frame), and the native dialog as the single startup-interrupted recovery surface.
- Vocabulary & i18n — rolled out user-facing "Mithril Sync" naming, localized diagnostics copy, JA copy review.
- QA hardening — four rounds of manual-testing/code-review remediation (702a–d) closing the manual-QA deployment gate, with a smoke-test cheat sheet added.

Notes

- There was an issue identified with the Mithril cancel process which was fixed for the Mithril Partial Sync flow. This same robust cancel flow should be applied via another issue ticket for the Mithril Bootstrap flow.


source/main/mithril/MithrilPartialSyncService.ts
david-profrontsolutions, line 1209: UX: Mithril Sync progress rows show raw stage ids instead of translated labels

Observed: During Mithril Sync (partial-sync overlay), the verify and convert steps render as the bare words verifying / converting (lowercase), while the install step correctly shows "Moving snapshot to storage". The raw word installing also flashes briefly before finalizing.

Root cause: MithrilPartialSyncService emits progress items with the label set to the raw stage id (_upsertProgressItem(id, id, 'active') — source/main/mithril/MithrilPartialSyncService.ts:1202, :777). In the renderer, ITEM_ID_TO_MESSAGE (source/renderer/app/components/loading/mithril-bootstrap/MithrilStepIndicator.tsx:86-97) only maps the bootstrap flow's ids (step-1…step-7, install-snapshot, cleanup, conversion), so unmapped partial-sync ids fall back to the raw backend label (MithrilStepIndicator.tsx:563-566). "Moving snapshot to storage" only appears because the renderer synthesizes an install-snapshot item during finalizing (keepInstallingActiveDuringFinalizing, :258-288), which is in the map.

Fix: Give the partial-sync stage ids (verifying, converting, installing) proper i18n labels — either add entries to ITEM_ID_TO_MESSAGE (reusing/adding messages such as "Verifying snapshot" / "Converting snapshot format") or emit bootstrap-compatible item ids from the service. Existing messages in en-US.json ("Verifying database integrity", "Verifying snapshot digests", "Converting snapshot format") show the intended label style.

nix/internal/launcher-config.nix
```diff
@@ -406,7 +406,8 @@ let
           legacySecretKey
           ;
         wipeChain = false;
-        mithrilPartialSyncEnabled = false;
+        mithrilPartialSyncEnabled = true;
+        mithrilPartialSyncThresholdImmutables = 20;  # behind-ness threshold in immutable files; calibrate in QA
```
david-profrontsolutions: Is this still required with the behind-ness updates using the network tip and is there a better place/structure for this constant?

source/main/mithril/MithrilPartialSyncService.ts
```diff
@@ -540,8 +820,12 @@ export class MithrilPartialSyncService {
   }
 
   _resetToIdleStatus(): void {
```
david-profrontsolutions, line 790: **Cross-session recovery boundary diverges — the only recovery button always rejects**

**Severity: Critical · Effort: Medium**

When a `starting-node` failure happens across sessions (marker `installed-awaiting-node-start`, then `cardanoNode.start()` throws on the next launch), the `failed` snapshot with `allowedRecoveryActions: ['wipe-and-full-sync']` is emitted only through `MithrilController`, while `MithrilPartialSyncService._status` remains the fresh-session default (`idle`, `[]`). The overlay shows a single "wipe and full sync" button, but every click hits `_assertRecoveryActionAllowed`, which throws because the service's own boundary is still `idle` — the IPC request rejects and the user is permanently stuck on the failed overlay with a button that never works. In-session failures only pass by coincidence (the service happens to be left at `finalizing` with the same action list). The service status should be seeded from the emitted recovery snapshot (or the recovery assertions should consult the controller's broadcast state) so cross-session failures accept the action the UI offers.

*(Anchored to the nearest changed line — the recovery methods and `_assertRecoveryActionAllowed` sit just above, around line 810.)*

source/main/mithril/mithrilPartialSyncNodeStartup.ts
```diff
@@ -58,6 +60,12 @@ export class MithrilPartialSyncNodeStartup {
     }
 
     if (marker.state === 'node-start-verified') {
+      // Boundary C2: a prior run already proved one successful node start on the installed DB.
+      // Reclaim leftover staging on a close-without-dismiss / crash, then resume
+      // a normal boot. stagingRootPath is the durable colocated root persisted at cutover.
+      if (marker.stagingRootPath) {
+        await fs.remove(marker.stagingRootPath);
```
david-profrontsolutions: **Unguarded staging-dir removal can block node startup on every boot**

**Severity: Major · Effort: Low**

In the `node-start-verified` reclaim branch, `await fs.remove(marker.stagingRootPath)` is unguarded and runs before `clearMithrilPartialSyncMarker()`. If the remove rejects (e.g. Windows `EBUSY`/`EPERM` from antivirus scanning the leftover staging dir), the error propagates into `MithrilStartupGate`'s layout-verify try and is converted via `markManagedChainLayoutError` — Daedalus refuses to start the node even though the installed chain DB is verified-good, and because the marker is never cleared, the same failure repeats on every subsequent boot. The remove should be wrapped in try/catch (log and continue), and the marker cleared regardless: a leftover staging dir is a disk-space nuisance, not a reason to block startup.

david-profrontsolutions (commented):
**Multi-agent code review — Mithril Sync (partial-sync) UX branch**, diffed against the upstream `master` merge-base (`7405cc06`).

38 findings survived independent per-finding verification against the diff (37 confirmed at their exact file/line, 1 plausible — flagged on the comment). 26 are posted as inline comments: **1 critical, 8 major, 17 minor**. The 9 nit-level cleanups are listed below instead of as threads, and 3 further findings were duplicates folded into their inline comments.

Severity reflects user impact; effort is the estimated size of the fix.

## Nit-level cleanups (not posted as inline threads)

- `source/main/mithril/killProcessTree.ts:55` — second, divergent Windows tree-kill implementation; `CardanoNode.ts:1080` already contains one inline. A Windows kill bug fixed in one stays broken in the other — this helper was the natural place to unify them. *(effort: medium)*
- `source/main/mithril/mithrilSnapshotMetadata.ts:105` — `extractCertifiedEpoch` keeps lookup paths its own comment declares dead, and the bare top-level `['epoch']` fallback can latch onto an unrelated field if the aggregator payload shape shifts, skewing the user-visible behind-by-epochs figure. *(effort: low)*
- `source/main/utils/chainStorageCoordinator.ts:75` — `getPartialSyncAvailability` fabricates a full availability shape (`isSignificantlyBehind: false`) from the single bit it actually knows; any future caller that trusts it will permanently report not-behind. A boolean `isPartialSyncEnabled()` getter makes that misuse impossible. *(effort: low)*
- `source/renderer/app/components/loading/mithril-bootstrap/MithrilPartialSyncOverlay.tsx:187` — ~13 pre-formatted intl strings threaded as individual props into `MithrilProgressView`, which already has intl in context; a `'bootstrap' | 'partial-sync'` variant flag would collapse the 20+-entry prop list and stop the two fallback chains drifting. *(effort: medium)*
- `source/main/mithril/MithrilController.ts:54` — the idle status snapshot literal is hand-maintained in three places (also `MithrilPartialSyncService.ts:99` and `MithrilPartialSyncStore.ts:26`); a shared `makeIdlePartialSyncStatus()` factory in the common types module removes the drift. *(effort: low)*
- `source/main/mithril/MithrilPartialSyncService.ts:714` — `_runBinary` dynamically imports `./mithrilCommandRunner`, which is already statically imported at the top of the file. *(effort: low)*
- `source/renderer/app/components/status/DaedalusDiagnostics.tsx:723` — `shouldShowRecommendation` re-checks `isMithrilPartialSyncEnabled` inside a JSX block already guarded by it at line 718. *(effort: low)*
- `source/common/types/mithril-partial-sync.types.ts:95` — `MITHRIL_PARTIAL_SYNC_OVERLAY_STATUSES` filters `'idle'` out of a const literal (`['completed','failed','cancelled']`) that can never contain it; a plain spread states the actual invariant. *(effort: low)*
- `storybook/stories/loading/mithril/MithrilProgressView.stories.tsx:22` — `SNAPSHOT_FILES_TOTAL = 980` is declared identically here and in `MithrilBootstrap.stories.tsx` instead of living in the shared `_support/mithrilFixtures` module alongside `snapshotSize`. *(effort: low)*

## Duplicates folded into the inline comments

- `MithrilPartialSyncStore.ts:391` (`mithrilAttemptStartedThisSession` set before the IPC and never rolled back) — same root cause as, and covered by, the inline comment on `MithrilPartialSyncStore.ts:413`.
- `MithrilPartialSyncStore.ts:444` (`wipeAndFullSync`/`restartNormally` await with no catch and no resync) — covered by the inline comment anchored just above `restartNormally`.
- A second independent report of the missing plural form on `SyncingConnectingMithrilPrompt.tsx:22` — same defect as the inline comment there.

source/main/mithril/MithrilPartialSyncService.ts
```diff
@@ -686,6 +1095,37 @@ export class MithrilPartialSyncService {
     );
   }
 
+  async _assertSufficientDiskSpace(stagingRootPath: string): Promise<void> {
+    const snapshotSize = this._latestSnapshot?.size ?? 0;
+    const requiredBytes = Math.max(
+      snapshotSize * PARTIAL_SYNC_DISK_SAFETY_FACTOR,
```
david-profrontsolutions: **Disk preflight demands 1.2x the full DB size, false-blocking the feature's target users**

**Severity: Major · Effort: Medium**

`requiredBytes = max(snapshot.size * 1.2, 4GB)` uses the whole-chain certified snapshot size (`mithrilSnapshotMetadata.ts` falls back to `total_db_size_uncompressed` — hundreds of GB on mainnet) even when the partial range is a few hundred MB. A user one epoch behind with less free space than ~1.2x the entire chain DB gets "Not enough free disk space…" and can never use partial sync — precisely the users the feature targets, since their disk is already mostly consumed by the existing chain. The preflight should estimate the actual partial range (e.g. from the immutable-file count being downloaded, or per-file ancillary size metadata) instead of the full snapshot size.

source/main/mithril/mithrilPartialSyncNodeStartup.ts
```diff
@@ -66,17 +74,10 @@ export class MithrilPartialSyncNodeStartup {
...
-      allowedRecoveryActions: ['wipe-and-full-sync'],
-      transferProgress: {},
-      progressItems: [],
-      error: {
-        message:
-          'Daedalus detected an interrupted Mithril partial sync after live chain cutover. The installed chain data is not safe to start normally.',
-        stage: 'installing',
-      },
-    });
+    // The native dialog below is the SINGLE authoritative startup-interrupted recovery
```
david-profrontsolutions: **Interrupted-cutover "Quit" branch dead-ends with no recovery surface**

**Severity: Major · Effort: Low**

`handleInterruptedRecovery` no longer emits the pre-dialog `failed` status (with `wipe-and-full-sync` recovery), and the native dialog's "Quit" branch neither calls `app.quit()` nor emits any status. After choosing Quit, the gate transitions to `partialSyncCutoverUnsafe`, the node never starts, and the renderer sits on the connecting screen indefinitely with no explanation and no recovery option — at merge-base this path showed the failed Mithril overlay explaining the interrupted cutover. The branch should either actually quit the app or restore the `failed` status emission so the renderer can present recovery.

source/main/mithril/MithrilPartialSyncService.ts
```diff
@@ -252,53 +357,158 @@ export class MithrilPartialSyncService {
...
       this._markActiveProgressItemAs('error');
       this._updateStatus({
         status: 'failed',
-        allowedRecoveryActions: [
-          'retry',
-          'restart-normal',
-          'wipe-and-full-sync',
-        ],
-        error: this._buildError(error, this._getCurrentRecoveryStage()),
+        allowedRecoveryActions: ['retry', 'restart-normal'],
```
david-profrontsolutions: **Wipe escape hatch removed from failed-cancel-cleanup state**

**Severity: Major · Effort: Low**

When `finalizeCancel`'s cleanup fails, `allowedRecoveryActions` was narrowed from `['retry','restart-normal','wipe-and-full-sync']` to `['retry','restart-normal']` (the `cancelled` success path at line 421 has the same narrowing). If the staging directory can't be removed (e.g. a Windows file lock), Retry re-prepares the same undeletable directory and fails again, and Restart-normally throws on the same `fs.remove` — the wipe-and-full-sync option that previously let the user escape the loop is gone. `wipe-and-full-sync` should be restored to both paths.

source/main/mithril/MithrilPartialSyncService.ts
```diff
@@ -552,25 +836,48 @@ export class MithrilPartialSyncService {
...
     this._stagedDbPath = null;
     this._startedAt = null;
+    this._cancelFallbackErrorStage = null;
   }
 
   _getCurrentRecoveryStage(): MithrilPartialSyncErrorStage {
     return this._status.error?.stage ?? this._getFallbackErrorStage();
   }
 
   _getStagingRootPath(): string {
```
david-profrontsolutions, line 827: **Session-local staging-path resolver orphans multi-GB staging dirs on custom chain volumes**

**Severity: Major · Effort: Medium**

`_getStagingRootPath()` changed from a stable `stateDirectoryPath/mithril-partial-sync` to a session-local path derived from `_stagingChainDir` with a `stateDirectoryPath` fallback. The marker re-establishes cross-session resolvability only for `finalizeCompletedPartialSync` and the node-start-verified reclaim — `wipeAndFullSync`, `restartNormal`, and `_cleanupPartialSyncArtifacts` still use the session-local resolver, and the startup `cutover-in-progress` wipe path never removes `marker.stagingRootPath`. With a custom chain directory on a separate volume, a crash after the cutover marker is written leaves the multi-GB staged snapshot permanently orphaned: the startup wipe skips it, and every fresh-session cleanup targets the wrong `stateDirectoryPath` fallback. All cleanup/recovery paths should resolve the staging root from the marker when present.

source/renderer/app/components/status/MithrilPartialSyncSection.tsx
```diff
@@ -107,22 +120,25 @@ export default class MithrilPartialSyncSection extends Component<Props, State> {
...
           isActionBlocked={isActionBlocked}
           startError={startError}
+          behindByEpochs={this.props.behindByEpochs}
           onCancel={this.hideConfirmation}
           onConfirm={this.startFromConfirmation}
         />
       );
     }
 
+    if (!shouldShowRecommendation) {
```
david-profrontsolutions: **Diagnostics entry point disappears exactly when behind-ness can't be measured**

**Severity: Major · Effort: Low**

The Diagnostics "Mithril Sync" section, previously always rendered, now returns `null` unless `shouldShowRecommendation` (`enabled && isSignificantlyBehind`) — and `getPartialSyncBehindness` degrades to `isSignificantlyBehind: false` on any probe failure. A user weeks behind whose aggregator is unreachable (offline, firewall, DNS) or whose immutable-dir read fails loses the manual trigger entirely for at least the 5-minute cache TTL, removing a capability that existed unconditionally at merge-base. The section should render whenever the feature is enabled (or at least when the probe errored rather than measured not-behind).

source/renderer/app/stores/MithrilPartialSyncStore.ts
```diff
@@ -178,6 +411,13 @@ export default class MithrilPartialSyncStore extends Store {
     }
 
     if (this.status !== START_PENDING_STATUS) {
```
david-profrontsolutions: **Every pre-run start rejection is silently swallowed and the prompt never re-offers**

**Severity: Major · Effort: Low**

`startPartialSync` rethrows a start rejection only if status is still the renderer-optimistic `'stopping-node'` after the resync — but the main process never emits `'stopping-node'`, so every preflight rejection (feature disabled, recovery-fallback layout, sync already in progress, `assertStartAllowed`) resyncs to `idle` and is swallowed. This makes `MithrilPartialSyncSection.startFromConfirmation`'s error display and the Diagnostics confirmation error path unreachable: the overlay flashes and vanishes with no error anywhere, and `mithrilAttemptStartedThisSession = true` (set at line 391, never reset) suppresses the proactive prompt for the rest of the session. The rejection should be surfaced regardless of post-resync status, and the session flag reset on preflight failure.

source/renderer/app/components/loading/mithril-bootstrap/MithrilPartialSyncOverlay.tsx
```diff
@@ -62,13 +73,95 @@ function MithrilPartialSyncOverlay(props: Props, { intl }: Context) {
...
+    onQuit,
     onOpenExternalLink,
   } = props;
 
+  // On 'completed', auto-fire the finalize IPC (reset-to-idle +
+  // remove staging dir + clear marker + folder deletion) via the stable
+  // `onDismissCompleted` MobX action once the success frame has lingered — the
+  // explicit "Continue to Daedalus" click is gone. All cleanup semantics are
+  // preserved; only the trigger changes from a click to this timeout.
+  useEffect(() => {
```
david-profrontsolutions, line 79: **Completed overlay lost its button; one failed auto-finalize strands the user on a dead success screen**

**Severity: Major · Effort: Low**

The "Continue to Daedalus" dismiss button was removed (`'completed'` added to `hideAction`, the `loading.mithrilPartialSync.completed.continue` key deleted) and replaced by a one-shot 4-second auto-finalize `useEffect` that never retries. If the finalize request fails while status stays `completed`, the store's no-blind-flip logic correctly keeps the overlay up — but the effect's deps (`[status, onDismissCompleted]`) are unchanged, so it never re-fires, and the overlay has no interactive element: the user's only exit is force-quitting Daedalus. Previously the Continue button could be clicked repeatedly. Either retry the finalize or keep a manual dismiss affordance on the completed state.

*(Verification note: the buttonless overlay, one-shot timer, and no-blind-flip store logic are all confirmed in the code; the exact finalize-rejection trigger was judged plausible rather than reproduced.)*

source/main/mithril/mithrilCommandRunner.ts
```diff
@@ -152,11 +152,29 @@ export async function runBinary(
...
   return new Promise((resolve, reject) => {
-    const child = spawn(binaryPath, args, { cwd: workDir, env });
+    // Detach on POSIX so the child leads its own process group and
+    // killProcessTree's process.kill(-pid) reaps its whole tree. Gated off Windows
+    // (documented launcher breakage — see CardanoSelfnodeLauncher.ts); stdio stays
+    // piped and the child is deliberately NOT unref()'d.
+    const child = spawn(binaryPath, args, {
+      cwd: workDir,
+      env,
+      detached: !environment.isWindows,
```
david-profrontsolutions, line 171: **Detached POSIX spawn lets mithril-client outlive an uncleanly-killed Daedalus**

**Severity: Minor · Effort: Medium**

Children are now spawned `detached` on POSIX (own process group), removing the implicit guarantee that mithril-client died with Daedalus on group-delivered signals; the compensating reap (`reapPartialSyncOnShutdown`/`forceKillForShutdown`) only runs on clean `safeExit` paths. If the main process crashes, is OOM-killed, or its process group is killed by a supervisor/terminal, the detached child silently keeps downloading a multi-GB snapshot — bandwidth, CPU, and chain-volume disk with no visible owner — until it completes or the machine reboots. Consider not detaching on POSIX unless required, or persisting the child PID so startup can reap orphans.

source/renderer/app/stores/MithrilPartialSyncStore.ts
```diff
@@ -186,7 +426,13 @@ export default class MithrilPartialSyncStore extends Store {
...
+    try {
+      await mithrilPartialSyncCancelChannel.request();
+    } finally {
+      // Always resync so the UI never sticks on the optimistic frame —
+      // including the stopping-node no-op and the post-cutover rejection.
+      await this.syncStatus();
+    }
   };
 
   @action
```
david-profrontsolutions, line 365: **Overlay recovery actions surface IPC rejections as unhandled promise rejections**

**Severity: Minor · Effort: Low**

`cancelPartialSync` re-throws after its finally-resync, and `restartNormally`/`wipeAndFullSync` (lines 439/444, below) are bare awaits — and App.tsx wires them directly into Button `onClick`, so any main-process rejection (e.g. Cancel racing the `installing` transition: "cancellation is no longer allowed after live chain cutover has started") becomes an unhandled promise rejection and a silent button no-op. These store actions should catch, resync, and surface the error to the overlay instead of letting it escape the click handler.

source/main/mithril/MithrilPartialSyncService.ts
```diff
@@ -166,8 +247,10 @@ export class MithrilPartialSyncService {
...
       );
 
-      const latestSnapshotAtDownload =
-        await this.resolveLatestSnapshotMetadata();
+      // The latest-drift re-resolve is also part of the cancelable run.
+      const latestSnapshotAtDownload = await this.resolveLatestSnapshotMetadata(
+        { trackAsCancelable: true }
+      );
       if (
         latestSnapshotAtDownload.latestCertifiedImmutableNumber !==
```
david-profrontsolutions, line 241: **Service-thrown user-facing messages are hardcoded English and use the banned "partial sync" vocabulary**

**Severity: Minor · Effort: Medium**

Stage-error messages rendered verbatim in the UI (lines ~260, 671-673, 1123, 567) say "partial sync" — e.g. "The latest certified Mithril snapshot changed during partial sync preparation…", "Mithril partial sync download failed with exit code…" — contrary to the product vocabulary ("Mithril Sync", never "partial sync"), and they bypass i18n entirely, so ja-JP users see raw English. These need error codes mapped to intl messages on the renderer side rather than English prose flowing through `error.message` (renderer-side counterpart: the comment on `MithrilErrorView.tsx:130`).

source/renderer/app/components/loading/mithril-bootstrap/MithrilStepIndicator.tsx
```diff
@@ -40,9 +41,11 @@ interface Context {
...
-const STATUS_TO_STEP: Partial<
-  Record<MithrilBootstrapStatus | MithrilPartialSyncStatus, StepId>
-> = {
+const STATUS_TO_STEP: Partial<Record<
+  MithrilBootstrapStatus | MithrilPartialSyncStatus,
+  StepId
+>> = {
+  'stopping-node': 'preparing',
   preparing: 'preparing',
   downloading: 'downloading',
```
david-profrontsolutions, line 50: **Partial-sync stage ids fall through to raw machine strings in the step list**

**Severity: Minor · Effort: Medium**

The service emits progress items with ids/labels `verifying`, `converting`, `installing` (label === id), but `ITEM_ID_TO_MESSAGE` / `DOWNLOAD_SUB_IDS` / `FINALIZE_SUB_IDS` (around line 85) only know the bootstrap vocabulary (`step-1`..`step-7`, `install-snapshot`, `conversion`, `cleanup`). During a real run the overlay shows raw lowercase ids as sub-items — untranslated in ja-JP, next to the properly translated "Moving snapshot to storage" item. The specs mask this by feeding bootstrap-style ids the partial-sync service never produces; the indicator needs mappings (and spec fixtures) for the actual partial-sync stage ids.

source/common/types/mithril-partial-sync.types.ts
```diff
@@ -29,6 +30,13 @@ export type MithrilPartialSyncErrorStage =
   | 'finalizing'
   | 'starting-node';
 
+export type MithrilPartialSyncErrorCode =
```
david-profrontsolutions, line 33: **Insufficient-disk-space error code can't reach the copy layer, so users get generic failure copy**

**Severity: Minor · Effort: Low**

The service throws `PARTIAL_SYNC_INSUFFICIENT_DISK_SPACE` from the disk preflight, but the shared `MithrilPartialSyncErrorCode` union omits it, so the renderer's `COPY_BY_CODE` map can't key on it (and `COPY_BY_STAGE` deliberately omits `'preparing'`). A user with a near-full volume sees the generic "Mithril Sync failed" title/hint instead of disk-space guidance — the required/available GB figures survive only in the raw-message details line. Add the code to the union and give it dedicated copy.

source/renderer/app/components/loading/mithril-bootstrap/MithrilErrorView.tsx
```diff
@@ -114,7 +128,12 @@ function MithrilErrorView(props: Props, { intl }: Context) {
           {title || intl.formatMessage(messages[copy.title])}
         </h1>
         {error?.message && <p>{error.message}</p>}
```
david-profrontsolutions: **Raw backend error.message rendered verbatim in the error overlay body**

**Severity: Minor · Effort: Medium**

The overlay body renders `error?.message` directly (also lines 105/159 in the details section), so service-thrown English strings — "Not enough free disk space to stage the Mithril partial sync…", "Mithril partial sync cancellation is no longer allowed after live chain cutover has started." — reach users untranslated and off-vocabulary. Raw messages belong in the collapsed details/logs area at most; the body should come from code-keyed intl copy. (Renderer-side counterpart of the comment on `MithrilPartialSyncService.ts` service-thrown messages.)

source/renderer/app/components/loading/syncing-connecting/SyncingConnectingMithrilPrompt.tsx
```diff
@@ -0,0 +1,245 @@
...
+const messages = defineMessages({
+  promptTitle: {
+    id: 'daedalus.diagnostics.dialog.mithrilProactivePromptTitle',
+    defaultMessage: '!!!Mithril Sync',
+    description:
+      'Choice-view title for the proactive Mithril prompt on the syncing screen',
+  },
+  promptBody: {
+    id: 'daedalus.diagnostics.dialog.mithrilProactivePromptBody',
+    defaultMessage: '!!!Your node is about {epochs} epochs behind.',
```
david-profrontsolutions: **"about 1 epochs behind" — bare {epochs} interpolation with no plural form**

**Severity: Minor · Effort: Low**

`'Your node is about {epochs} epochs behind.'` has no ICU plural argument, and `computeBehindByEpochs` returns exactly 1 in the most common near-tip case — so both the proactive prompt and the Diagnostics confirmation (`MithrilPartialSyncConfirmation.tsx:18`) render "about 1 epochs behind". `{epochs, plural, one {# epoch} other {# epochs}}` fixes both.

source/main/mithril/MithrilPartialSyncService.ts
```diff
@@ -600,6 +907,105 @@ export class MithrilPartialSyncService {
...
+      value: localImmutableNumber,
+      fetchedAt: now,
+    };
+    return localImmutableNumber;
+  }
+
+  // Drop both behind-ness input caches so the next probe re-resolves fresh. Called at
+  // start()/cancel() entry and inside _resetToIdleStatus() (restart-normal/wipe/finalize paths).
+  _invalidateBehindnessCaches(): void {
+    this._latestCertifiedImmutableCache = null;
```
david-profrontsolutions, line 941: **Behind-ness caches survive a chain-directory change**

**Severity: Minor · Effort: Low**

`_invalidateBehindnessCaches` is called only at `start()`/`cancel()` entry and `_resetToIdleStatus` — `chainStorageCoordinator._notifyDirectoryChanged` only reaches `MithrilController.resetStartupGateOnDirectoryChange`. After the user relocates chain storage, up to 5 minutes of probes serve the old directory's immutable count: the proactive prompt fails to appear for a now-empty directory, or reports "significantly behind" for one at tip. Invalidate the caches on directory change.

source/renderer/app/components/status/MithrilPartialSyncSection.tsx
```diff
@@ -98,7 +111,7 @@ export default class MithrilPartialSyncSection extends Component<Props, State> {
   };
 
   render() {
-    const { formattedSyncPercentage, isActionBlocked, isSynced } = this.props;
+    const { isActionBlocked, shouldShowRecommendation } = this.props;
     const { isShowingConfirmation, startError } = this.state;
```
david-profrontsolutions, line 117: **Hardcoded English fallback "Unable to start Mithril partial sync."**

**Severity: Minor · Effort: Low**

The non-`Error` rejection fallback (line 107) is a raw string (not react-intl) using the banned "partial sync" vocabulary — and it has already drifted from the sibling fallback in `SyncingConnectingMithrilPrompt` ("Unable to start Mithril sync."). Move it to an intl message with the approved wording.

source/renderer/app/components/loading/syncing-connecting/SyncingConnectingMithrilPrompt.tsx
```diff
@@ -0,0 +1,245 @@
...
+        'SyncingConnectingMithrilPrompt: Mithril sync start rejected after confirmation',
+        { error }
+      );
+      if (!this._isMounted) {
+        return;
+      }
+
+      this.setState({
+        isStarting: false,
+        startError:
```
david-profrontsolutions: **Start-rejection message extraction implemented three times with divergent fallbacks**

**Severity: Minor · Effort: Low**

The `error instanceof Error ? error.message : fallback` extraction exists here, in `MithrilPartialSyncSection.startFromConfirmation`, and in `MithrilPartialSyncStore.toStartError` — with already-divergent fallback copy. Any future fix (e.g. handling serialized IPC error objects like `toStartError` does) must be re-applied in each place or the dialogs behave differently for the same failure. Extract one shared helper.

source/common/types/mithril-partial-sync.types.ts
```diff
@@ -102,6 +122,7 @@ export const isMithrilPartialSyncBlockingNodeStart = (
 ): boolean =>
   [
```
david-profrontsolutions, line 157: **isMithrilPartialSyncBlockingNodeStart re-lists all nine working statuses inline**

**Severity: Minor · Effort: Low**

The inline array is byte-identical to `MITHRIL_PARTIAL_SYNC_WORKING_STATUSES` defined in the same file. If a new working stage is added to the constant but the inline copy is missed, the node-start gate silently stops blocking that stage and cardano-node can start mid-sync — during cutover that risks chain-DB corruption. Reuse `isMithrilPartialSyncWorkingStatus`.

source/renderer/app/components/loading/mithril-bootstrap/MithrilPartialSyncOverlay.tsx
```diff
@@ -38,6 +41,7 @@ interface Context {
 
 const PROGRESS_STATUSES: MithrilPartialSyncStatus[] = [
```
david-profrontsolutions: **PROGRESS_STATUSES hand-copies the working-status set**

**Severity: Minor · Effort: Low**

The local array duplicates the nine working statuses plus `'completed'` instead of `isMithrilPartialSyncWorkingStatus(status) || status === 'completed'`. A status added to the common set but not here makes the overlay fall through to `MithrilErrorView` for a healthy state — a failure dialog during a working sync.

source/main/mithril/mithrilCommandRunner.ts
```diff
@@ -229,14 +248,29 @@ export async function runCommand(
...
 
   return new Promise((resolve, reject) => {
+    // Detach on POSIX so the child leads its own process group and
+    // killProcessTree's process.kill(-pid) reaps its whole tree. Gated off Windows
+    // (documented launcher breakage — see CardanoSelfnodeLauncher.ts); stdio stays
+    // piped and the child is deliberately NOT unref()'d.
     const child = spawn(binaryPath, commandArgs, {
       cwd: workDir,
       env,
+      detached: !environment.isWindows,
```
david-profrontsolutions: **Spawn block duplicated verbatim between runBinary and runCommand**

**Severity: Minor · Effort: Medium**

The ~15-line block (detached spawn options, spawn info log, exit/close handlers) is pasted into both paths. A one-sided edit to kill semantics or the detach gate silently gives download children and conversion children different process-group behavior, breaking `killProcessTree` for one of them. Extract a shared spawn helper.

source/main/mithril/MithrilPartialSyncService.ts
```diff
@@ -600,6 +907,105 @@ export class MithrilPartialSyncService {
...
+      logger.warn(
+        'MithrilPartialSyncService: behind-ness probe failed; treating as not significantly behind',
+        { error }
+      );
+      return { isSignificantlyBehind: false };
+    }
+  }
+
   async listSnapshots(): Promise<Array<MithrilSnapshotItem>> {
     const { stdout } = await this._runCommand(
```
david-profrontsolutions: **listSnapshots/showSnapshot duplicate the raw CLI pipelines**

**Severity: Minor · Effort: Low**

Four methods spawn the same two CLI commands with copy-pasted argument arrays and JSON parsing, differing only in the normalizer. A future CLI flag change must be made twice or the bootstrap picker and partial-sync preflight see different snapshot sets; `listSnapshots` could map `_listSnapshotsRaw` results directly.

source/main/mithril/MithrilPartialSyncService.ts
```diff
@@ -600,6 +907,105 @@ export class MithrilPartialSyncService {
...
+  }
+
+  async getPartialSyncBehindness(): Promise<{
+    isSignificantlyBehind: boolean;
+    behindByImmutables?: number;
+    certifiedEpoch?: number | null;
+  }> {
+    try {
+      const latest = await this._getCachedLatestCertifiedImmutableNumber();
+      // Cached local read dedupes the checkDiskSpace fork + immutable/ readdir.
```
david-profrontsolutions: **getPartialSyncBehindness serializes two independent reads**

**Severity: Minor · Effort: Low**

The aggregator metadata fetch and local immutable readdir are awaited sequentially though independent. On a cold cache with a slow aggregator plus slow disk, the serial sum can exceed the renderer's 10s availability timeout, logging spurious failures and delaying the proactive prompt. `Promise.all` halves the cold-path latency.

source/renderer/app/stores/MithrilPartialSyncStore.ts
```diff
@@ -52,13 +69,38 @@ export default class MithrilPartialSyncStore extends Store {
...
+  // Separate session-scoped re-pop guard, set true when a Mithril
+  // attempt begins (`startPartialSync`). AND-ed into the proactive-prompt gate so
+  // the prompt never re-offers after an attempt regardless of the terminal
+  // outcome (completed/cancelled/failed/restart-normal). In-memory, session-scoped,
+  // never reset (no idle reset). Distinct from `proactivePromptDismissedThisSession`,
+  // which stays single-purpose ("user clicked Standard Sync on the prompt").
+  @observable mithrilAttemptStartedThisSession = false;
+  @observable isPartialSyncEnabled = false;
+  @observable isSignificantlyBehind = false;
+  @observable behindByImmutables: number | undefined = undefined;
```
david-profrontsolutions: **behindByImmutables observable has no consumer**

**Severity: Minor · Effort: Low**

Written on every 30s availability refresh but never read — and per the epochs-only display rule it must never be rendered, so it exists only as an invitation for a future component to display immutable-file counts (a documented copy violation). Drop the field or keep it main-side.

source/renderer/app/stores/MithrilPartialSyncStore.ts
```diff
@@ -52,13 +69,38 @@ export default class MithrilPartialSyncStore extends Store {
   @observable filesDownloaded: number | undefined = undefined;
   @observable filesTotal: number | undefined = undefined;
   @observable elapsedSeconds: number | undefined = undefined;
```
david-profrontsolutions: **elapsedSeconds observable is dead since the startedAt refactor**

**Severity: Minor · Effort: Low**

Assigned on every status frame (line 239) but nothing reads it since the overlay switched to the `startedAt` anchor. Two elapsed-time representations invite the next contributor to wire the stale one, reintroducing the jumpy backend-driven timer the refactor removed.

nix/internal/launcher-config.nix
AndrewWestberg: I don't know how to exactly make a number for this. This seems to be wanting to do some type of percentage behind the tip. However, that's not a useful metric since the chain is always adding blocks to the denominator. 2% behind today might be very close to tip. 2% five years from now might take a massive time to sync. I think we need to just go with X blocks from tip means we allow them to do fast partial mithril sync. I don't know how fast blocks sync, but if we're more than 30 minutes from being on tip, I think we should allow them to do partial sync.

source/renderer/app/stores/MithrilPartialSyncStore.ts
AndrewWestberg: Can we avoid swallowing start rejections when the status resync returns to a non-error state? This currently returns for any status other than the optimistic `stopping-node`. If main rejects before `MithrilPartialSyncService.start()` publishes a `failed` status, `syncStatus()` can restore `idle`; because `mithrilAttemptStartedThisSession` was already set, the proactive prompt disappears and the user gets no visible error or retry path. Consider only swallowing once the backend has published an error/terminal state, and rethrow/reset the session attempt guard when the resynced status is `idle`.

source/renderer/app/components/loading/syncing-connecting/SyncingConnectingMithrilPrompt.tsx
AndrewWestberg: `computeBehindByEpochs` can return `1`, but this string always renders the plural noun, so users can see "1 epochs behind". The diagnostics confirmation copy has the same pattern. Can we switch these to ICU plural formatting, e.g. `{epochs, plural, one {# epoch} other {# epochs}}`, and update the locale/message snapshots and tests for both the proactive prompt and diagnostics confirmation copy?

nix/internal/launcher-config.nix
david-profrontsolutions: The `mithrilPartialSyncThresholdImmutables = 20;` represents 20 Immutable chunk Files which, from my research, is equivalent to 1 Epoch. Percentage was initially implied/proposed, but that was removed. If there are some references to percentage still I will make a sweep. 

I chose 1 epoch for now, but will do further testing to identify how that matches up on mainnet as a rough indicator.

source/main/mithril/MithrilPartialSyncService.ts
david-profrontsolutions: Fixed in `7c9f7de20` — preflight now estimates the missing range (snapshot − chain dir) plus a 20 %-of-snapshot margin, floored at 4 GB; whole-snapshot bound only as measurement-failure fallback.

nix/internal/launcher-config.nix
david-profrontsolutions: The constant is already absolute (immutable chunks ≈ 6 h each; 20 ≈ 5 days), not a percentage. In `7c9f7de20` the manual trigger in Diagnostics becomes available whenever the feature is enabled (no behind-ness gate), and the constant only gates the proactive prompt — keeping that conservative while letting mildly-behind users opt in manually. Number stays QA-calibratable in launcher config.

source/renderer/app/stores/MithrilPartialSyncStore.ts
david-profrontsolutions: Fixed in `7c9f7de20` per your suggestion — rejections rethrow unless the backend published an error/terminal status, and the session attempt guard resets when the resync lands on `idle`, so the prompt re-offers and errors are visible.

source/renderer/app/components/loading/syncing-connecting/SyncingConnectingMithrilPrompt.tsx
david-profrontsolutions: Fixed in `7c9f7de20` — ICU plural on both surfaces, locales and specs updated.

source/main/mithril/MithrilPartialSyncService.ts
david-profrontsolutions: Fixed in `7c9f7de20` — recovery actions seed service status from the broadcast snapshot, so the wipe button validates against what the user sees. Repro spec added.

source/main/mithril/mithrilPartialSyncNodeStartup.ts
david-profrontsolutions: Fixed in `7c9f7de20` — reclaim removal is best-effort; marker always clears; EBUSY spec added.

source/main/mithril/mithrilPartialSyncNodeStartup.ts
david-profrontsolutions: Fixed in `7c9f7de20` — Quit actually quits (`safeExitWithCode`); the continue branch was verified already correct and is unchanged.

source/main/mithril/MithrilPartialSyncService.ts
david-profrontsolutions: Wipe stays off by design here — a recorded decision (pre-cutover DB is intact; wipe would hit the same lock). The trap itself is fixed in `7c9f7de20` — cleanup is best-effort, restart-normal always proceeds, leftover staging is reclaimed at the next partial-sync start.

source/main/mithril/MithrilPartialSyncService.ts
david-profrontsolutions: Fixed in `7c9f7de20` — marker-first staging resolution across all cleanup/recovery paths; custom-volume spec added.

source/main/mithril/mithrilCommandRunner.ts
david-profrontsolutions: Intentionally detached — `killProcessTree`'s group kill (`kill(-pid)`) depends on the child being a group leader, and `detached` doesn't cause crash-orphaning on POSIX (children reparent to init either way). No change; persisted-PID reaper recorded as a future option.

source/renderer/app/components/status/MithrilPartialSyncSection.tsx
david-profrontsolutions: Fixed in `7c9f7de20` — section always renders when the feature is enabled, with state-adaptive copy; the manual trigger is never hidden by behind-ness or probe failures.

source/renderer/app/stores/MithrilPartialSyncStore.ts
david-profrontsolutions: Fixed in `7c9f7de20` (with the duplicate thread) — rejections surface, session flag resets on idle resync, prompt re-offers.

source/renderer/app/stores/MithrilPartialSyncStore.ts
david-profrontsolutions: Fixed in `7c9f7de20` — recovery actions catch, resync, and surface errors in the overlay; unhandled-rejection specs added.

source/main/mithril/MithrilPartialSyncService.ts
david-profrontsolutions: Fixed in `7c9f7de20` — behind-ness caches invalidate on chain-directory change.

source/main/mithril/MithrilPartialSyncService.ts
david-profrontsolutions: Fixed in `7c9f7de20` — parallelized with `Promise.all`.

source/renderer/app/components/loading/mithril-bootstrap/MithrilPartialSyncOverlay.tsx
david-profrontsolutions: Auto-finalize stays by design (a recorded decision); fixed the dead-end in `7c9f7de20` — one retry, then the error view with a working retry action. No more swallowed rejection.

source/common/types/mithril-partial-sync.types.ts
david-profrontsolutions: Fixed in `7c9f7de20` — code added to the union with dedicated disk-space copy.

source/main/mithril/MithrilPartialSyncService.ts
david-profrontsolutions: Fixed in `7c9f7de20` — renderer now maps `verifying`/`converting`/`installing` to translated labels; fixtures use real service ids. (Same fix as the step-indicator thread.)

source/renderer/app/components/loading/mithril-bootstrap/MithrilStepIndicator.tsx
david-profrontsolutions: Fixed in `7c9f7de20` together with the progress-label thread (`MithrilPartialSyncService.ts`) — mappings and honest fixtures added.

source/renderer/app/components/status/MithrilSyncProcessSummary.messages.ts
```diff
@@ -0,0 +1,23 @@
...
+ * view (`SyncingConnectingMithrilPrompt.tsx`) so the byte-identical sentence
+ * lives behind a single i18n key (`daedalus.diagnostics.dialog.mithrilSyncProcessSummary`).
+ * Do NOT redeclare this id in either consumer — import this module instead.
+ */
+const messages: Record<string, ReactIntlMessage> = defineMessages({
+  processSummary: {
+    id: 'daedalus.diagnostics.dialog.mithrilSyncProcessSummary',
+    defaultMessage:
+      '!!!For this process to begin your Cardano node will need to be shutdown. Mithril will then be used to sync the verified chain data. On Mithril Sync completion, the node will be restarted to sync the remaining blocks.',
+    description:
```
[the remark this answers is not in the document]
david-profrontsolutions: Fixed in `7c9f7de20` — "shut down", locales and specs regenerated.

source/main/mithril/MithrilPartialSyncService.ts
david-profrontsolutions: Fixed in `7c9f7de20` — error codes over prose at all user-reachable throw sites; renderer owns the words via i18n.

source/renderer/app/components/loading/mithril-bootstrap/MithrilErrorView.tsx
david-profrontsolutions: Fixed in `7c9f7de20` — intl-keyed body copy; raw message demoted to details.

source/renderer/app/components/loading/syncing-connecting/SyncingConnectingMithrilPrompt.tsx
david-profrontsolutions: Same fix as the duplicate report on this line — ICU plural (`{epochs, plural, one {# epoch} other {# epochs}}`) on both the prompt and the Diagnostics confirmation; locales and specs updated. Fixed in `7c9f7de20`.

source/renderer/app/components/status/MithrilPartialSyncSection.tsx
david-profrontsolutions: Fixed in `7c9f7de20` — intl fallback, approved vocabulary, single shared helper.

source/renderer/app/components/loading/mithril-bootstrap/MithrilBootstrap.scss
```diff
@@ -8,9 +10,10 @@
 }
 
 .backdrop {
+  @include overlay-backrop;
   background: var(--theme-mithril-overlay-backdrop-start);
   height: 100%;
-  opacity: 0.97;
+  opacity: 1;
```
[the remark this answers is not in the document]
david-profrontsolutions: Fixed in `7c9f7de20` — dead blur removed.

source/renderer/app/components/loading/syncing-connecting/SyncingConnectingMithrilPrompt.tsx
david-profrontsolutions: Fixed in `7c9f7de20` — one shared helper: known codes resolve their mapped intl copy, everything else gets one intl fallback; raw messages/codes never render.

source/common/types/mithril-partial-sync.types.ts
david-profrontsolutions: Fixed in `7c9f7de20` — single source of truth for working statuses.

source/renderer/app/components/loading/mithril-bootstrap/MithrilPartialSyncOverlay.tsx
david-profrontsolutions: Fixed in `7c9f7de20` — derived from the shared predicate.

source/main/mithril/mithrilCommandRunner.ts
david-profrontsolutions: Fixed in `7c9f7de20` — shared spawn helper; both children get identical lifecycle semantics.

source/main/mithril/MithrilPartialSyncService.ts
david-profrontsolutions: Fixed in `7c9f7de20` — CLI pipelines deduplicated.

source/renderer/app/stores/MithrilPartialSyncStore.ts
david-profrontsolutions: Fixed in `7c9f7de20` — dead observable removed.

source/renderer/app/stores/MithrilPartialSyncStore.ts
david-profrontsolutions: Fixed in `7c9f7de20` — dead observable removed.

AndrewWestberg (approved):

source/renderer/app/components/loading/syncing-connecting/SyncingConnectingStatus.tsx
```diff
@@ -80,6 +82,12 @@ const messages = defineMessages({
     defaultMessage: '!!!TLS certificate is not valid, please restart Daedalus.',
     description: 'The TLS cert is not valid and Daedalus should be restarted',
   },
+  mithrilSyncInterrupt: {
+    id: 'loading.screen.mithrilSyncInterrupt',
+    defaultMessage: '!!!Use Mithril Sync instead',
+    description:
+      'Button label shown during ledger replay to interrupt and switch to Mithril Sync',
+  },
 });
```
david-profrontsolutions, line 91: We need to integrate this into the message structure, I will add it in as part of the copy adjustment before merging the PR.

source/main/mithril/MithrilPartialSyncService.ts
david-profrontsolutions, line 1: There is a broken unit test off this file, I will fix it with the copy adjustment work.

source/common/types/mithril-partial-sync.types.ts
david-profrontsolutions: As this error code logic was adjusted, this might be dead code now. Also includes associated i18n message copy.

david-profrontsolutions (commented):
There are a few clean up items that need to be addressed before merging in the PR. I don't see any major issues. I will be doing a copy adjustment pass that can fix the higher priority ones.

1. Mithril Button on the Cardano Node page:  When clicked, while the Cardano Node was stopping, the Mithril Sync seemed to already start downloading the snapshot before the node was done with cleanup. This needs to be looked at closer to make sure there isn't a conflict between any immutable files or other processes. (medium priority)

2. When only downloading the ancillary files, the progress bar jumps around because it thinks there are only 2 snapshot files. Will have to adjust the way the progress bar works in this scenario. (low priority)

3. It already had to be updated, but the Daedalus Diagnostic page allows for the Mithril Sync to activate even when close to the tip. This now only downloads the ancillary state for the user, but is unnecessary either way for a user to click. The copy will be adjusted to make sure the user is aware of proceeding with the Mithril Sync.

The remaining are minor issues addressed via the comments below.

disassembler:
@david-profrontsolutions review by claude:

Code Review: PR #3337 — Enhance Mithril Partial Sync UX

  Author: David Kirshon | Prior approval: Andrew Westberg | Your review:
  Requested

  ---
  Overview

  This follow-up to #3333 promotes the Mithril Partial Sync from a "fresh
  install only" tool into a general catch-up mechanism for nodes that have
  fallen significantly behind. The main additions are: a new
  availability/behind-ness model and IPC channel, a two-phase cancel that
  eliminates a cleanup race, process-group kill for mithril-client children, a
  disk-space preflight, new UX surfaces (proactive fork prompt + inline
  ledger-replay button), and startup/recovery hardening. The code is high
  quality and the architecture is consistent with #3333.

  ---
  Blocking Issues

  [B1] derivePartialSyncRange silently proceeds when node is already caught up
  source/main/mithril/mithrilPartialSyncPreflight.ts

  Previously, when localImmutableNumber >= latestCertifiedImmutableNumber the
  function threw PARTIAL_SYNC_NO_CERTIFIED_RANGE. Now it returns { start:
  latest, end: latest } and proceeds — the rationale being that the ledger
  state may be missing even when immutable files are up to date.

  The scenario is real, but the side effects are bad: disk-space preflight runs
   unnecessarily, mithril-client is invoked as a no-op, the informative error
  is gone, and the Diagnostics "Force Mithril Sync" button can silently trigger
   this zero-range path near tip. At minimum, a log line explaining why a
  zero-range sync was kicked off is needed. Ideally this should be a separate
  code path rather than a silent side-effect of removing the early-out.

  [B2] mithrilPartialSyncEnabled = true in production Nix config
  nix/internal/launcher-config.nix

  This enables partial sync for all networks including mainnet in production
  build artifacts. That may be intentional for QA, but it must not land on
  develop/main without a clear plan. Needs either a revert to false before
  merging, or an explicit comment + merge-blocking note documenting the
  intended revert.

  [B3] abandonCancel() emits allowedRecoveryActions: [] — escape hatch
  unverified
  source/main/mithril/MithrilPartialSyncService.ts

  When cleanup after a failed cancel is itself abandoned, the overlay receives
  allowedRecoveryActions: [] with status failed. If the overlay doesn't
  explicitly render a Quit button in this case, the user is stuck with no
  escape short of force-quitting. The defensive branch appears to exist in the
  component, but there's no test proving it. A unit test or Storybook story for
   this exact state is needed.

  ---
  Non-Blocking Issues

  [N1] Disk-space preflight fails open when snapshot size is unknown
  When _latestSnapshot?.size is 0 or null, requiredBytes falls back to the
  DISK_SPACE_REQUIRED floor constant — which may be very small. Change ?? 0 to
  ?? Infinity so an unknown snapshot size fails safe (preflight rejects) rather
   than silently passing.

  [N2] "Use Mithril Sync instead" button during ledger replay has no minimum
  threshold
  source/renderer/app/components/loading/syncing-connecting/SyncingConnectingSt
  atus.tsx
  A user at 95% ledger replay can click this and throw away nearly-complete
  work. Consider gating the button behind a lower threshold (e.g.
  replayProgress < 50) or adding a confirmation step. (Note: We already added a
   2-second debounce on our branch to fix the 99.99% flash — this is a related
  but separate concern about the threshold itself.)

  [N3] Fresh install boots into proactive Mithril prompt immediately
  An empty immutable directory returns position 0, making the node appear
  maximally behind, so the proactive prompt fires on the very first boot. If
  this is intentional, the interaction with the existing bootstrap path (fresh
  install → bootstrap vs. fresh install → partial sync) should be documented.

  [N4] computeBehindByEpochs fallback can understate how far behind the node is
  source/renderer/app/utils/mithrilBehindness.ts
  The certifiedEpoch fallback lags the actual network tip by up to one epoch,
  so the displayed "N epochs behind" figure is slightly understated. The
  comment acknowledges this, but the UI label should say "approximately" to set
   correct expectations.

  [N5] _handleDecline guard ordering changed subtly — no action needed,
  behavior is identical, worth keeping in mind during future refactors.

  [N6] stdinInput/onLogStream removed from mithrilCommandRunner.ts without
  deprecation — confirm yarn compile passes clean (it should, given CI).

  ---
  What's Done Well

  - makeIdlePartialSyncStatus() factory correctly prevents shared nested-object
   references across state resets — an important correctness fix over the
  previous spread-constant pattern.
  - Two-phase cancel (cancel() → cancelling, then finalizeCancel()) is
  architecturally sound and eliminates the previous cleanup race.
  - detached: true + process group kill is the correct POSIX pattern, and the
  Windows taskkill /pid N /t /f fallback is right. The spawn option comment
  explaining why detached is needed is excellent.
  - startedAt timestamp replacing elapsedSeconds correctly fixes the
  frozen-elapsed display bug.
  - Availability polling backoff (30s → 5min after 2 stable reads) is a
  thoughtful network-efficiency improvement.
  - _throwIfCancelled() at each await seam is the right
  cooperative-cancellation pattern.
  - Disk-space formula max(snapshot − chainDir, 0) + 0.2 × snapshot is more
  accurate than requiring the full snapshot size for a partial sync.

  ---
  Test Coverage Gaps

  - No unit tests for finalizeCancel / abandonCancel state machine paths.
  - No story or unit test for allowedRecoveryActions: [] overlay state (see
  B3).
  - Confirm mithrilBehindness.ts utility functions have unit coverage.

  ---
  Verdict

  Well-implemented, meaningfully safer than #3333. The three blocking issues
  are targeted fixes — none are architectural. Ready to approve once B1, B2, B3
   are addressed.

disassembler:
I don't think I agree with B1/B2.

disassembler:
I also think our approach is better than N2

PR #3345, opened 2026-07-24
Fix portable Linux XKB and GBM paths

Liberty-Chris:
## Summary

Fixes a Linux-specific issue where Daedalus can fail to launch because the portable installer references XKB keyboard data at an unavailable Nix store path and does not include Mesa's `dri_gbm.so` backend.

The Linux package now includes the XKB configuration inside the portable bundle and sets `XKB_CONFIG_ROOT` relative to the installed application.

It also bundles `dri_gbm.so` and its dependencies, then sets `GBM_BACKENDS_PATH` to the bundled Electron library directory.

## Tested

- Linux installer build passed
- Confirmed bundled XKB rules are present
- Confirmed bundled `dri_gbm.so` is present
- Confirmed `dri_gbm.so` uses an `$ORIGIN` RPATH
- Confirmed GBM dependencies resolve from the portable bundle
- Installer checksum verified after transfer
- Installer completed successfully on native Ubuntu
- Daedalus launched without the original XKB, GBM, or segmentation-fault errors
- Mithril bootstrap began successfully
- Full blockchain synchronization was not completed because the test partition lacked sufficient storage

### DIFF
--- a/nix/internal/x86_64-linux.nix
+++ b/nix/internal/x86_64-linux.nix
@@ -194,6 +194,7 @@ in rec {
         ${pkgs.libva.out}/lib/*.so.2
         ${pkgs.atk}/lib/libatk-bridge-2.0.so
         ${pkgs.libgbm}/lib/libgbm.so.1
+        ${pkgs.mesa}/lib/gbm/dri_gbm.so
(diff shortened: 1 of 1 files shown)


disassembler:
superceded by #3347

PR #3348, opened 2026-07-27
Fix Windows mapped-drive chain target fallback

disassembler:
## Summary

Fixes a Win specific issue where Daedalus can lose track of a chain directory located on a mapped NAS/network drive.

Chain junction points to a path such as `Z:\DaedalusChain\chain`, `fs.readlink` still returns the correct target, but `fs.realpath` can fail with `ENOENT`. In that case, Daedalus was falling back to the local `<stateDir>\chain` path instead of keeping the configured mapped drive location.

Resolver now reads the junction target first, uses the resolved path when available, and falls back to the original junction target when `realpath` fails.

Added a regression test for this case.

### Tested

- TypeScript compilation passed
- Main-process build passed
- Existing Windows junction test passed
- New mapped-drive regression test passed
- Reproduced and verified using a local SMB share mapped as `Z:`
- Confirmed the development build launched and shut down normally

### DIFF
--- a/source/main/utils/chainStoragePathResolver.ts
+++ b/source/main/utils/chainStoragePathResolver.ts
@@ -43,22 +43,32 @@ const resolveManagedChainPathFromEntryPoint = async (
     const chainStats = await fs.lstat(chainPath);
 
     if (chainStats.isSymbolicLink()) {
+      const rawLinkTarget = await fs.readlink(chainPath);
+      const linkTargetPath = path.isAbsolute(rawLinkTarget)
+        ? rawLinkTarget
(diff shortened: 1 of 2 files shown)


Liberty-Chris (commented):
Windows NAS/UNC testing

Liberty-Chris (approved):
Windows NAS/UNC test passed.

Tested on Windows 11 64-Bit using an actual network attached storage device accessed over SMB through a UNC path and mapped as Z:. Daedalus correctly retained the configured chain storage target on the NAS and did not fall back to the local state directory. The application launched and shut down normally, and the fix behaved as expected.

Approved.

PR #3355, opened 2026-08-04
DRep Discovery

disassembler:
Replaces the GovernanceQueryService/CLI two-phase DRep query with the `GET /v2/dreps*` wallet API endpoints. Delivers the full DRep discovery and delegation flow end-to-end.

**What's new:**

- **DRep directory** (`/governance/dreps`) — suggested cohort of 20 DReps with reroll, show-all, search (name substring + ID prefix + exact), sort by voting power / activity / name, and pagination
- **DRep detail page** (`/governance/dreps/:drepId`) — full off-chain metadata (objectives, motivations, qualifications, references, payment address), on-chain stats, verified-name anchor source label
- **Favorites** (`/governance/favorites`) — persistent favorites stored in localStorage; favorites view with empty state and stale-favorite indicators for `doNotList` DReps
- **Auto-favorite** — DReps are automatically added to favorites when selected for delegation (directory, detail page, and existing delegations via `onEnsureFavorited`)
- **Sentinel cards** — Abstain and No Confidence rendered separately below the cohort; excluded from search and favorites
- **Delegation handoff** — `GovernanceStore.delegationNavState` carries form state across route boundaries (workaround for hash-history v4 dropping `location.state`); full round-trip Form → Directory → Detail → Form
- **GovernanceStore** — new MobX store: `suggestedDReps`, `allDReps` (lazy-loaded on search/show-all), `favoriteDRepIds`, `delegationNavState`, `refresh`/`loadAllDReps`/`fetchDRep`
- **Verified names** — wallet backend fetches and hash-verifies off-chain metadata; names clamped to 80 chars in the store; anchor host shown in confirmation dialog
- **CIP-129 canonical form** — all internal state uses CIP-129 bech32; CIP-105 derived on demand; `isSameVoteTarget` compares by credential hex + type so both forms match

## Test plan

- [ ] DRep directory loads suggested cohort on mount; reroll fetches a new cohort
- [ ] Show all triggers `GET /v2/dreps` (once, guarded by no-op on repeat calls)
- [ ] Name search filters by verified name substring (min 2 chars); ID prefix search filters by bech32 prefix (min 8 data chars); exact valid ID resolves directly
- [ ] Sentinels (Abstain, No Confidence) appear below cohort in default view; hidden during search and show-all
- [ ] Selecting a DRep for delegation auto-adds it to favorites; selecting when already favorited does not toggle it off
- [ ] Wallet with existing DRep delegation: opening the governance form auto-favorites the delegated DRep
- [ ] Favorites view shows favorited DReps; stale badge appears for `doNotList` DReps
- [ ] Delegation round-trip: Form → Browse DReps → Select → returns to Form with wallet + DRep pre-filled
- [ ] Two-hop: Form → Browse DReps → View details → Select → returns to Form with wallet + DRep pre-filled
- [ ] DRep detail page shows full metadata; verified name shows anchor host; favorite toggle works independently of delegation
- [ ] `doNotList` DReps excluded from directory and search (except exact ID lookup)
- [ ] Jest specs pass: `yarn test:jest`


Crypto2099:
First comment here, I notice that we have a filter for "Expiry" with the options of "All" or "Expiring in 7-12 epochs". What is the value of showing this to users who are attempting to select a DRep? I understand the premise of warning them that a DRep may be about to become inactive via the badge, but I'm not sure that's something that a user would actually want to filter for?

Crypto2099:
Second comment, when at the default directory page, there is a toggle to show "Favorited" DReps, but given that we're querying a random sampling of 20 DReps, there is likely to be none of my favorites in the results. Toggling "Show All" + "Favorited" does show the DRep I am currently delegated to but this is an awkward UX.

Crypto2099:
Third comment, on the DRep Directory, there seems to be abundant whitespace in the current list layout, it may may sense to swap to a card base layout (or provide some options like card, list, table)... Also, the "Threshold" and "Primary" badges seem to cancel one another out although they represent different things (one indicates whether off-chain metadata is verified, the other whether or not the DRep is close to expiry)

Crypto2099:
Fourth comment: DRep Detail page...

If a section is empty because a DRep did not provide it in their metadata (i.e., References/Links) then these should be hidden from view.

Crypto2099:
Fifth comment: Drep Detail Page...

It's good that we are already forward-thinking with the "Current Votes" section, however, this will rapidly outgrow the small area that it currently has in the upper context box, this would likely be better served as a separate table below the profile information and probably "hidden by default" unless the user requests to see the DRep's voting history.

Crypto2099:
Open question, when showing the list of all DReps, I am given the ability to sort them based on delegated power, high to low. Is it possible to get a number on "total stake delegated to DReps" based on the  most recent snapshot or something else that could provide an indicator when a DRep is above a certain "power threshold" (i.e. if a DRep is responsible for >5% of voting power within the system, show a warning badge)?

source/renderer/app/stores/GovernanceStore.ts
```diff
@@ -0,0 +1,389 @@
...
+
+  @computed get hasError(): boolean {
+    return this.refreshState === GovernanceRefreshState.Failed;
+  }
+
+  @computed get isGovernancePage(): boolean {
+    return (
+      this.stores.router.location.pathname.indexOf(ROUTES.GOVERNANCE.ROOT) === 0
+    );
+  }
```
Crypto2099, line 225: This is currently throwing an error in the console on the startup screen:
mobx.module.js:1838 [mobx] Encountered an uncaught exception that was thrown by a reaction or observer component, in: 'Reaction[Reaction@337]' TypeError: Cannot read properties of null (reading 'pathname')
    at get isGovernancePage (GovernanceStore.ts:165:35)
    at trackDerivedFunction (mobx.module.js:775:1)
    at ComputedValue.computeValue (mobx.module.js:1272:1)
    at ComputedValue.trackAndCompute (mobx.module.js:1257:1)
    at ComputedValue.get (mobx.module.js:1217:1)
    at ObservableObjectAdministration.read (mobx.module.js:4136:1)
    at GovernanceStore.get (mobx.module.js:4400:1)
    at GovernanceStore.ts:365:18

Crypto2099:
Couple of points about delegation certificates being issued:
1. In the overall transaction list, a certificate that includes delegation to an SPO but auto-delegates to the auto-abstain DRep currently appears as a "Voting Power Delegation" which is not entirely accurate, may need to rephrase this to simply being a "Delegation Transaction"?

image

https://preprod.cexplorer.io/tx/3876d07d02774012ed5abab7fdee35149d6486dd505d20a98e1b4d0b66efe754?tab=delegations

2. Wallet is continuing to show undelegated in the "Delegation Center" even after submitting an SPO Delegation certificate in the previous epoch:
image

3. The current governance page upon submitting delegation to a DRep seems to only be submitting a DRep delegation certificate but not any SPO information via the combined certificate. Not sure if this is problematic or not, but one of my wallets that was previously delegated to an SPO but not a DRep switched to "undelegated" in the interface after I submitted a DRep-only delegation certificate...
image

https://preprod.cexplorer.io/tx/470c10c5abcb0b685fbed229aa25f8985cffbc8081c5bd79129a75d6c4344494?tab=delegations

image

Crypto2099:
Multiple console errors related to DRep discovery window in the console logs:

```
[daedalus.*preprod*:error:ipcRenderer] {
  "msg": "AdaApi::getDRepSummary error",
  "data": {
    "error": {
      "code": "bad_request",
      "message": "Invalid DRep key/script hash: expecting a Bech32 encoded value with human readable part of 'drep', 'drep_vkh' or 'drep_script'."
    }
  }
}
```

```
[daedalus.*preprod*:error:ipcRenderer] {
  "msg": "GovernanceStore: fetchDRepSummary failed",
  "data": {
    "errorType": "UNKNOWN"
  }
}
```

Crypto2099:
> Multiple console errors related to DRep discovery window in the console logs:
> 
> ```
> [daedalus.*preprod*:error:ipcRenderer] {
>   "msg": "AdaApi::getDRepSummary error",
>   "data": {
>     "error": {
>       "code": "bad_request",
>       "message": "Invalid DRep key/script hash: expecting a Bech32 encoded value with human readable part of 'drep', 'drep_vkh' or 'drep_script'."
>     }
>   }
> }
> ```
> 
> ```
> [daedalus.*preprod*:error:ipcRenderer] {
>   "msg": "GovernanceStore: fetchDRepSummary failed",
>   "data": {
>     "errorType": "UNKNOWN"
>   }
> }
> ```

Note that these errors seem to be triggered when the list of candidate DReps on the page is refreshed specifically so may be a race condition against an empty DRep list?

disassembler:
## DRep Discovery — Code Review

### Bug 1: Anchor URL color (dark blue, should be cyan)

**`DRepDetailAnchorSection.tsx:126`, `DRepDetail.scss:108–123`**

The anchor URL `<dd>` has class `anchorValue`, which sets `color: var(--theme-gov-font-color-accent)`. `DRepExternalLink` renders a bare `<a>` with no color rule of its own, so it inherits that gray and the browser applies its default dark blue on top.

There is already a dead `.anchorLink` class at `DRepDetail.scss:121–123` with `color: var(--theme-link-main-color)` (the correct cyan) — nothing in the codebase ever applies it. The reference links in the references section work correctly because `.referenceItem a` uses a CSS descendant selector that catches the `<a>` inside the list item.

**Fix:** Wrap the `<DRepExternalLink>` at `DRepDetailAnchorSection.tsx:126` with `<span className={styles.anchorLink}>`, or apply `anchorLink` directly to the `<dd>`. Either way, the orphaned `.anchorLink` rule in `DRepDetail.scss` should become a proper selector on the element rather than dead code.

---

### Bug 2: "Highest voting power" sort doesn't show the large DReps

**`DRepDirectory.tsx:363`, `drepCohort.ts:63`**

When `!isShowAll`, `visibleEntries` sources from `suggestedDReps` — 20 items drawn from a pool already filtered by `maxVotingPowerShare: 0.015`. Any DRep above 1.5% is structurally absent from that pool. Selecting "Voting power (high → low)" just re-orders those 20; it cannot surface the large DReps. The sort dropdown appears regardless of `isShowAll` state with no indication it is operating on a pre-filtered list. `onSortChange` does not call `setIsShowAll(true)`, so there is no automatic expansion.

**Fix options (pick one):**
- Auto-switch to show-all when `votingPowerDesc`/`votingPowerAsc` is selected in `onSortChange`
- Hide/disable power sort options when `!isShowAll`
- Add a banner when power sort is active in cohort mode explaining it operates within the eligible pool only

---

### Bug 3: Filter bar dropdowns don't match app control language

**`DRepDirectoryFilters.tsx:137–148`, `DRepDirectoryFilters.scss:53–78`**

The filter/sort selects are native `<select>` elements stripped of browser chrome (`appearance: none`, `background: transparent`, `border: 0`). The value text uses the same gray as the caption label above it (`--theme-gov-font-color-accent`), the chevron is 50% opacity, and there is no per-control border or background. The app's standard dropdown (settings page, `ProfileSettingsForm.tsx:110–119`) is Polymorph `Select` + `SelectSkin` — a labeled, boxed, 50px-tall control. The filter strip reads as unfinished next to the bordered search field immediately above it and the 50px controls in the row above that.

Two fixes, pick one:

**Option A — Switch to Polymorph Select (full match to app standard)**
Replace each native `<select>` with `<Select skin={SelectSkin} optionHeight={50} label={…}>` matching `ProfileSettingsForm.tsx:110–119`. Break the horizontal strip into individual labeled controls. More vertical space, visually consistent with the rest of the app.

**Option B — Fix the strip's visual affordances (keep compact layout)**
- Raise `min-height` from 46px → 50px to match the row above
- Change value text color from `--theme-gov-font-color-accent` to `--theme-gov-font-color-regular` so the selected value is visually distinct from the caption label
- Change chevron color from `--theme-gov-font-color-lighter` (50% opacity) to `--theme-gov-font-color-accent`
- Add subtle background and padding to `.facetControl` so each control has a visible hit target

Option A removes the custom pattern and reuses the existing styled component. Option B is lower-risk and may be more appropriate if the filter bar is intentionally a compact toolbar rather than a form.

---

### Note: top-35 exclusion replaced by 1.5% cap — plan doc needs update

**`governance-drep-discovery-plan.md:62,120`**

The main plan still says "exclude top 35 DReps by voting power rank." The implementation uses a 1.5% share cap (`drepCohort.ts:63`) — this is a deliberate design change, not a regression. The threshold comes directly from the **Target15 campaign** by Japanese SPO hix (@hix_coffeepool): to bring Cardano's CR10 (top-10 concentration ratio) from ~47.8% down to 15%, each of the top 10 DReps must hold ≤87M ADA, which is ≤1.5% of total voting power. The 1.5% cap embeds that goal into the cohort filter: any DRep above the threshold is one a user should be encouraged to move away from, not toward. The rank-based approach excluded exactly 35 regardless of how stake was actually distributed and gave no signal about why — the share cap is directly tied to a measurable decentralization target and self-adjusts as the network shifts.

`cohort-selection.md` already documents the rationale. `governance-drep-discovery-plan.md:62,120` just needs the stale "top 35" language replaced to avoid confusing future reviewers.

---

### Other issues

**Dead `host` prop on `VerifiedFieldRow`** (`DRepDetailAnchorContent.tsx:206–221`): Accepted and passed at every call site, never used in render. Remove it.

**`isRankingAvailable` is hardcoded `true`** (`DRepDirectory.tsx:305`): This prop gates the power sort options in `DRepDirectoryFilters`. Since it is always true, the conditional at `DRepDirectoryFilters.tsx:209` is dead and power sort options always appear. Either wire it to something real (e.g. hide when voting power data is unavailable) or remove the prop and conditional entirely.

**Stale dependency in `visibleEntries` memo** (`DRepDirectory.tsx:384`): `favoriteDRepIds` is in the dep array but the memo reads `favoriteEntries`/`pinnedFavoriteIds` instead. Not a correctness bug, but misleading.

---

### What looks solid

- **Cohort relaxation logic** (`drepCohort.ts:209–234`): Progressive criteria-dropping in the right priority order (metadata first, activeOnly last) is well-designed and well-commented.
- **Seeded shuffle** (`drepCohort.ts:244–266`): mulberry32 + Fisher-Yates, per-band seeding avoids the same permutation for equal-length bands.
- **`doNotList` is pre-filter and non-relaxable** (`drepCohort.ts:99–101`): Correctly enforced outside the relaxation loop — a DRep's own opt-out is never overridden when the pool is thin.
- **URL safety in `DRepExternalLink`**: `preventDefault()` + routing through `onOpenExternalLink`, real `href` preserved for browser status bar. Explicitly noted in comment.
- **Anchor verification layering** (`GovernanceStore.ts:399–445`): "Our fetcher leads, wallet is fallback, cache keyed by hash" is sound. The comment explaining why hash is the cache key is accurate.
- **`blankToNull` normalization** (`GovernanceStore.ts:101–105`): Consistently applied to all string fields from the API, preventing empty-string fields from rendering as labelled rows with no content.
- **`CurrentDRepSummary` badge-state derivation** is clean and the comment noting it unifies the six-epoch threshold with the directory (retiring the old twelve-epoch window) shows the change was intentional.

Crypto2099 (approved):

PR #3357, opened 2026-08-05
ci(watchdog): run the cargo test suite as a flake check

Crypto2099:
The `watchdog` crate carries 94 tests and none of them ran in CI.
`perSystem/watchdog.nix` defined a single check, `watchdog-clippy`, so the Rust
code that now supervises `cardano-node` and `cardano-wallet` was linted but never
tested.

This adds `checks.watchdog-test` via `craneLib.cargoTest`. `flake.nix` aggregates
`self.checks.x86_64-linux` into the `hydraJobs.required` job, so it becomes
merge-blocking with no further configuration.

## The source filter also needed widening

The crane fileset covered only `Cargo.toml`, `Cargo.lock` and `src`. Two
consequences:

- `cargoTest` would have compiled and run the 67 unit tests, silently skipped
  `tests/integration.rs` and `tests/ipc.rs`, and reported success.
- `watchdog-clippy` is configured with `--all-targets`, but the integration tests
  were never in its source closure, so it has been linting less than it claims
  since the crate landed.

Adding `watchdog/tests` to the fileset fixes both.

## Cargo.lock guard

The module wrapped its outputs in `lib.optionalAttrs` on a `pathExists` check of
`watchdog/Cargo.lock`. A missing lockfile evaluated to an empty attrset, removing
the package and both Rust checks from the flake with no error reported: CI would
report green with the watchdog neither built nor tested. `Cargo.lock` is tracked
in git, so its absence is a broken tree rather than a supported configuration.
This asserts on it instead.

## Verification

| Command | Result |
| --- | --- |
| `nix build .#checks.x86_64-linux.watchdog-test` | 94 passing: 67 unit, 25 integration, 2 ipc |
| `nix eval .#checks.x86_64-linux` | `watchdog-test` present |
| Evaluation with `Cargo.lock` moved aside | fails with the intended message rather than silently dropping both checks |
| `nix fmt -- --ci` | 0 changed |

The sandbox run is the result that matters: the 25 integration tests spawn 13
mock binaries and bind loopback ports, and all of that works under Nix isolation.


disassembler (approved):

disassembler:
closing, appended to watchdog PR

PR #3361, opened 2026-08-05
fix(chain-storage): report a read-only directory as not-writable

Crypto2099:
Selecting a read-only directory for chain storage tells the user only
"Unable to validate selected directory", which does not say what is wrong or
what to do about it.

## The defect

`fs.access(resolvedValidationPath, W_OK)` raises `EACCES`, which fell through to
the generic `catch` and returned `reason: 'unknown'`. The `'not-writable'` reason
was only ever produced for a path that **is not a directory**, so
`chainStorageUtils.ts:41` — which already has `validationNotWritable` copy — was
unreachable for the case it was written for.

`EACCES`, `EPERM` and `EROFS` are now caught at the access call and mapped to
`'not-writable'`. Anything else is rethrown, so genuinely unexpected failures
still surface as `'unknown'` rather than being mislabelled.

## A real-filesystem spec

`chainStorageValidation.spec.ts` mocks `fs-extra` wholesale. That is the right
tool for branch logic, but it cannot express this defect: the bug is *which errno
the real syscall raises*, and a mock only ever returns what the author already
believed.

`chainStorageValidation.realfs.spec.ts` is new and does not mock `fs-extra`. It
uses real temp directories, real permission bits and real symlinks. The existing
mocked spec is untouched and still passes — Jest's module registry is per test
file, so the two coexist.

| Case | Covers |
| --- | --- |
| read-only directory | `EACCES` from `fs.access(W_OK)` — the defect above |
| symlinked target | `realpath` resolution to the real path |
| dangling symlink | link exists, target deleted |
| file selected as target | not-a-directory |
| spaces and non-ASCII in the path | `Ünïcödé 目録 dir` |

The read-only test **asserts its own precondition** — that the `0555` mode really
denies writes — because a process with privileges that bypass mode bits, such as
root or `CAP_DAC_OVERRIDE` in a container, would otherwise pass it vacuously. It
fails loudly in that situation rather than reporting a false green.

## Verification

The spec was written first and run against unpatched code, where it failed with
`Expected: "not-writable"` / `Received: "unknown"`. It passes with the fix.

`lint`, `compile`, `stylelint`, `treefmt` and `jest` all pass — 67 suites, 783
tests.

## Context

This is the first of the real-filesystem cases. It follows the pattern
established by `ensureDirectoryExists.spec.ts` in #3352, which fixed a closely
related defect: a state or logs directory that is a symlink. Both are the same
class — code reasoning about the filesystem from assumptions no test ever
checked against a real one.

### DIFF
--- a/source/main/utils/chainStorageValidation.realfs.spec.ts
+++ b/source/main/utils/chainStorageValidation.realfs.spec.ts
@@ -1,0 +1,157 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import { validateChainStorageDirectory } from './chainStorageValidation';
+
+// This spec deliberately does NOT mock `fs-extra`. `chainStorageValidation.spec.ts`
+// covers the branch logic against a mocked filesystem; the cases here are ones a
+// mock cannot express, because they depend on how the real filesystem behaves:
+// permission bits, symlink resolution, and the errno a syscall actually raises.
+//
+// Jest's module registry is per test file, so mocking `fs-extra` there and using
+// the real module here does not conflict.
+
+// `../config` refuses to load outside the Nix shell, and `./logging` writes to
+// the Electron app data directory. Both are mocked; neither is the filesystem
+// behaviour under test. `check-disk-space` is left real, so the free-space read
+// happens against the actual temp volume.
+jest.mock('../config', () => ({
+  DISK_SPACE_REQUIRED: 1024,
+}));
+
+jest.mock('./logging', () => ({
+  logger: {
+    warn: jest.fn(),
+    error: jest.fn(),
+    info: jest.fn(),
+    debug: jest.fn(),
+  },
+}));
+
+const REQUIRED_SPACE = 1024;
+
+const makeGetDefaultConfig = (defaultPath: string) =>
+  jest.fn().mockResolvedValue({
+    defaultPath,
+    availableSpaceBytes: Number.MAX_SAFE_INTEGER,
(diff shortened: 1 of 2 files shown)


disassembler (approved):
Correct fix, right errno set, good tests. Approve.

`fs.access(path, W_OK)` raises `EACCES` on a read-only directory, which fell through to the generic catch and returned `reason: 'unknown'` — showing the user "Unable to validate selected directory" with no actionable information. The `'not-writable'` reason and its UI copy already existed but were only reachable for a non-directory path, making them dead code for the case they were written for.

**`EACCES`, `EPERM`, and `EROFS` is the right set.** Mode bits, capabilities/LSM, and read-only filesystem mount respectively. Rethrow on anything else so unexpected failures still surface as `'unknown'` rather than being mislabelled.

**`expectWritesDenied` precondition is correct** — root and `CAP_DAC_OVERRIDE` bypass mode bits; without this assertion the test passes vacuously in a privileged container. Same pattern as the EXDEV preconditions in #3363.

**`afterEach` restores write permissions before `rmSync`** — necessary, otherwise recursive removal fails on the read-only directories the spec creates.

**Test written before the fix**, failing with `Expected: "not-writable"` / `Received: "unknown"` — confirms it exercises the actual defect.

Minor: the `'file selected as target'` test only asserts `result.isValid === false` without checking `result.reason`. Since the renderer has specific copy per reason, asserting it too would be more complete. Not blocking.

PR #3362, opened 2026-08-05
fix(chain-storage): resolve relative symlink targets to absolute paths

Crypto2099:
Two resolvers disagreed about the same symlink.

## The defect

When the chain entry point is a symlink whose target `realpath` cannot resolve,
`resolveManagedChainPathFromEntryPoint` falls back to `readlink`, which returns
the target **exactly as recorded**. For a relative link that is a relative path.

`captureChainPathState` in `chainStorageManagerShared.ts` already handles this:

```ts
linkTargetPath = path.isAbsolute(rawLinkTarget)
  ? rawLinkTarget
  : path.resolve(path.dirname(ctx._chainPath), rawLinkTarget);
```

`chainStoragePathResolver.ts` did not. This makes it match.

## Why it matters

`resolveMithrilWorkDir` feeds `MithrilBootstrapService.setWorkDir()` and the
partial sync cutover in `chainStorageCoordinator.ts`. A relative path there
resolves against `process.cwd()`, not the state directory — so Mithril would
work against a location unrelated to the configured chain storage.

## The Windows interaction, which an existing test caught

My first attempt applied `path.isAbsolute` and broke
`chainStorageWindowsNetwork.spec.ts`:

```
Expected: "Z:\DaedalusChain\chain"
Received: "/home/adam/daedalus/C:\Users\test\...\Daedalus Mainnet/Z:\DaedalusChain\chain"
```

`path` is the POSIX implementation whenever this runs on Linux — including in
tests that simulate Windows by overriding `process.platform`, since `path` does
not follow that override. `path.isAbsolute('Z:\\...')` is therefore `false`, and
the target was resolved into nonsense.

Absoluteness is now tested against **both** flavours:

```ts
path.isAbsolute(rawTarget) || path.win32.isAbsolute(rawTarget)
```

A drive-letter or UNC target stays intact; a genuinely relative target is
absolute under neither, so it still gets resolved.

The Windows **junction** branch is deliberately left returning the raw readlink
target. Junctions record absolute paths, so there is nothing to resolve.

## Verification

`chainStoragePathResolver.realfs.spec.ts` is new and uses a real filesystem.
Written before the fix; the two relevant cases failed with `Received:
"../elsewhere"` and `path.isAbsolute(result)` false, and pass after.

| Case | Covers |
| --- | --- |
| plain directory | no link involved |
| absolute symlink | `realpath` succeeds |
| relative symlink | `realpath` succeeds, result must be absolute |
| relative symlink, target removed | `realpath` fails — **the defect** |
| absolute symlink, target removed | fallback stays absolute |
| `Z:\DaedalusChain\chain` target | the win32 interaction above |
| missing entry point | falls back to the chain path |

All 127 chain-storage tests pass, including the mapped-drive regression test.
Full set green: 67 suites, 785 tests, plus lint, compile, stylelint and treefmt.

### DIFF
--- a/source/main/utils/chainStoragePathResolver.realfs.spec.ts
+++ b/source/main/utils/chainStoragePathResolver.realfs.spec.ts
@@ -1,0 +1,107 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import { resolveMithrilWorkDir } from './chainStoragePathResolver';
+
+// Real filesystem, no `fs-extra` mock. `chainStoragePathResolver.spec.ts` covers
+// the branch logic against mocks; these cases depend on what the filesystem
+// actually does — in particular that `realpath` fails on a dangling link while
+// `readlink` still returns its raw target.
+
+jest.mock('./logging', () => ({
+  logger: {
+    warn: jest.fn(),
+    error: jest.fn(),
+    info: jest.fn(),
+    debug: jest.fn(),
+  },
+}));
+
+describe('resolveMithrilWorkDir against a real filesystem', () => {
+  let tmpRoot: string;
+  let stateDir: string;
+  let chainPath: string;
+
+  beforeEach(() => {
+    tmpRoot = fs.realpathSync(
+      fs.mkdtempSync(path.join(os.tmpdir(), 'chain-path-resolver-'))
+    );
+    stateDir = path.join(tmpRoot, 'state');
+    fs.mkdirSync(stateDir, { recursive: true });
+    chainPath = path.join(stateDir, 'chain');
+  });
+
+  afterEach(() => {
+    fs.rmSync(tmpRoot, { recursive: true, force: true });
+  });
+
+  it('returns the chain directory itself when it is a plain directory', async () => {
+    fs.mkdirSync(chainPath);
+
+    await expect(resolveMithrilWorkDir(stateDir)).resolves.toBe(chainPath);
+  });
+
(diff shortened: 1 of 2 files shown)


disassembler (approved):
Correct fix, well-tested, clearly documented. Approve.

When `realpath` fails on the chain entry point, the fallback to `readlink` returns the target as-recorded. For a relative symlink that's a relative path — which then resolves against `process.cwd()` rather than the state directory in `setWorkDir()` and the partial sync cutover. `captureChainPathState` already handled this correctly; this makes the resolver match.

**The `path.win32.isAbsolute` check is the key insight.** `path` is the POSIX implementation on Linux even in tests that override `process.platform`. Without the win32 check, `Z:\DaedalusChain\chain` fails `path.isAbsolute` and gets resolved into nonsense. Checking both flavours is correct.

**The junction branch is correctly left alone** — junctions always record absolute paths, nothing to resolve.

**Tests written before the fix**, with the two key cases failing on `Received: "../elsewhere"` — confirms the fix exercises the actual defect path.

**The Windows drive-letter test is valid** — Linux creates symlinks with arbitrary target strings; `realpath` fails, `readlink` returns it, `path.win32.isAbsolute` catches it correctly.

Minor: UNC paths (`\\server\share`) are mentioned in the docstring as handled but have no test case. The logic covers them (`path.win32.isAbsolute` returns true for UNC) — just worth adding a test if the pattern comes up in the field.

PR #3363, opened 2026-08-05
test(chain-storage): exercise the EXDEV fallback in movePath for real

Crypto2099:
`movePath` falls back to copy-then-remove when `fs.move` fails with `EXDEV` —
what happens when a user relocates chain storage to another disk.

Every migration spec stubs it out with `jest.spyOn(manager, '_movePath')`, so
**that branch had never executed in a test**, despite being the code that moves a
user's entire chain directory.

## The obstacle was not what it looked like

Testing `EXDEV` was assumed to need loopback mounts and therefore root.
`/dev/shm` is a tmpfs on a different device, present and writable on a developer
machine *and inside the Nix build sandbox*:

```
build dir: dev=2096
/dev/shm:  dev=84, writable: yes
```

So a genuine cross-device move needs no privileges and no setup.

## Coverage

| Case | Covers |
| --- | --- |
| same-device move | `fs.move` succeeds by rename |
| cross-device move | **the EXDEV fallback** |
| non-EXDEV error | rethrow, target not created |

## Two deliberate choices

**The cross-device test asserts its own preconditions** — that the two roots
differ by `st_dev`, and that a bare `renameSync` between them really raises
`EXDEV`. Without those, `fs.move` could succeed by rename and the test would pass
green while exercising nothing.

**It fails loudly rather than skipping** when no cross-device pair is available.
A silent skip on the only test covering this path would report coverage that does
not exist.

## Verification

Run inside the sandbox with `nix build .#checks.x86_64-linux.jest`, not only on
the host, because the whole question was whether `/dev/shm` is available there:
**67 suites, 781 tests**. Lint, compile, stylelint and treefmt also pass.

### DIFF
--- a/source/main/utils/chainStorageManagerShared.realfs.spec.ts
+++ b/source/main/utils/chainStorageManagerShared.realfs.spec.ts
@@ -1,0 +1,126 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import { movePath } from './chainStorageManagerShared';
+
+// Real filesystem, no `fs-extra` mock.
+//
+// `movePath` exists to survive a move across a device boundary: `fs.move` fails
+// with EXDEV, and it falls back to copy-then-remove. Every migration spec stubs
+// this function out with `jest.spyOn(manager, '_movePath')`, so until now the
+// EXDEV branch had never executed in a test — despite being the code that moves
+// a user's entire chain directory, potentially tens of gigabytes, to another
+// disk.
+
+jest.mock('./logging', () => ({
+  logger: {
+    warn: jest.fn(),
+    error: jest.fn(),
+    info: jest.fn(),
+    debug: jest.fn(),
+  },
+}));
+
+/**
+ * Returns a directory on a different device from `reference`, or null.
+ *
+ * `/dev/shm` is a tmpfs on Linux and is present and writable inside the Nix
+ * build sandbox, which makes a genuine cross-device move testable without any
+ * privileges or mounting.
+ */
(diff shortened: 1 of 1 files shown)


disassembler (approved):
Approve. Correct, well-reasoned, and the sandbox verification removes the last excuse for not having this test.

**The `/dev/shm` insight is the whole contribution** — the assumption that cross-device testing requires loopback mounts was wrong, and demonstrating that `/dev/shm` is present and writable inside the Nix sandbox is exactly the verification that matters.

**The precondition assertions within the test are correct and necessary:**
```ts
expect(fs.statSync(source).dev).not.toBe(fs.statSync(otherDeviceDir).dev);
expect(() => fs.renameSync(source, target)).toThrow(
  expect.objectContaining({ code: 'EXDEV' })
);
```
Without these, `fs.move` could succeed by rename and the test passes green while exercising nothing. `renameSync` on Linux is atomic — calling it here doesn't corrupt the source.

**Fails loudly rather than skipping** when no cross-device pair is available — correct, same reasoning as #3369's `subst` test.

**`findCrossDeviceRoot`** tries multiple candidates and confirms write access before claiming one. Degrades gracefully if `/dev/shm` isn't available.

Three cases cleanly cover the function: same-device (rename path), cross-device (EXDEV fallback), non-EXDEV error (rethrow, no target created).

**Minor:** `crossDeviceRoot as string` after the Jest null assertion — TypeScript can't narrow through Jest matchers, so the cast is necessary. Fine as-is.

PR #3366, opened 2026-08-06
fix(chain-storage): roll back a failed migration even if the journal cannot

Crypto2099:
Investigating the field report that a state or logs directory may be a symlink —
the same class as #3352 — turned up a defect in the migration rollback path.

## The rollback was gated behind bookkeeping

```ts
} catch (error) {
  journal.state = 'rollback';
  await ctx._writeMigrationJournal(journal);    // unguarded
  await ctx._rollbackMigrationJournal(journal); // never reached if it throws
  throw error;
}
```

Whatever broke the migration can equally break that write. **A full disk is the
obvious case**, because the migration is moving the entire chain directory: the
move fails with `ENOSPC`, the rollback should restore the previous layout, and
instead the journal write fails on the same full disk. The user is left with
chain data half moved between two locations, and the original error is replaced
by the journal one, so the logs do not even say what actually went wrong.

Recording the rollback state is bookkeeping. The rollback is what protects the
data. It must not depend on the bookkeeping succeeding.

The first journal write, at the top of the migration, is deliberately left
unguarded — it happens **before** anything moves, so failing there aborts safely.

## Verification

The new spec fails against unpatched code with exactly the symptom described:

```
Expected: [Error: move failed partway]
Received: [Error: no space left on device]
```

The original error was replaced and the rollback never ran. With the fix, the
rollback runs once and the original error propagates.

## How the Logs directory actually behaves

`chainStorageManagerLayout.realfs.spec.ts` is new and records, against a real
filesystem, what journal I/O does for each shape that directory can take:

| `Logs` is | `writeMigrationJournal` | `readMigrationJournal` |
| --- | --- | --- |
| a plain directory | writes | reads |
| a symlink to a directory | writes **into the target** | reads |
| a relative symlink | writes into the target | reads |
| a **dangling** symlink | throws `ENOENT` | returns `null` |
| a corrupt file | — | returns `null` |

The symlink cases are the field-reported scenario and they work. The dangling
case is what makes a journal write able to fail at all, which is what the fix
above addresses. `fs.ensureDir` raises `ENOENT` on a dangling symlink rather than
creating the directory — worth knowing, since the obvious assumption is that
`ensureDir` makes a directory appear no matter what was there.

Full set green: 67 suites, 785 tests, plus lint, compile, stylelint and treefmt.

### DIFF
--- a/source/main/utils/chainStorageManager.spec.ts
+++ b/source/main/utils/chainStorageManager.spec.ts
@@ -670,6 +670,54 @@ describe('ChainStorageManager', () => {
     );
   });
 
+  it('_migrateLegacyCustomLayout rolls back even when the journal cannot record it', async () => {
+    const manager = new ChainStorageManager('/tmp/state');
+    const migrationFailure = new Error('move failed partway');
+
+    jest
+      .spyOn(manager, '_preflightLegacyMigration')
+      .mockResolvedValue(undefined);
+    // Whatever broke the migration can equally break the journal write: a full
+    // disk fails both. Recording the rollback must not gate the rollback.
+    jest
+      .spyOn(manager, '_writeMigrationJournal')
+      .mockImplementation(async (journal) => {
+        if (journal.state === 'rollback') {
+          throw Object.assign(new Error('no space left on device'), {
+            code: 'ENOSPC',
+          });
+        }
+      });
+    jest.spyOn(manager, '_movePath').mockRejectedValue(migrationFailure);
+    jest.spyOn(manager, '_createSymlink').mockResolvedValue(undefined);
+    jest.spyOn(manager, '_pathExistsViaLstat').mockResolvedValue(false);
+    const rollback = jest
+      .spyOn(manager, '_rollbackMigrationJournal')
+      .mockResolvedValue(undefined);
+
+    await expect(
+      manager._migrateLegacyCustomLayout({
+        kind: 'legacy-custom-root',
+        customPath: '/mnt/custom-parent',
+        resolvedCustomPath: '/mnt/custom-parent',
+        managedChainPath: '/mnt/custom-parent/chain',
+        resolvedManagedChainPath: undefined,
+        currentChainSource: '/mnt/custom-parent',
+        entryPointState: {
+          type: 'symlink',
+          resolvedPath: '/mnt/custom-parent',
+        },
+        managedChainExists: false,
+        managedChainIsDirectory: false,
+        managedLegacyEntries: ['immutable'],
+        ignoredLegacyEntries: [],
+      })
+      // The original failure must survive, not be replaced by the journal error.
+    ).rejects.toBe(migrationFailure);
+
+    expect(rollback).toHaveBeenCalledTimes(1);
+  });
+
   it('_recoverInterruptedMigration cleans up a completed cutover on restart', async () => {
     const manager = new ChainStorageManager('/tmp/state');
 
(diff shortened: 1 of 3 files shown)


disassembler:
## Review: PR #3366 — `fix(chain-storage): roll back a failed migration even if the journal cannot`

### Overview

In the migration catch block, `_writeMigrationJournal` was called unguarded before `_rollbackMigrationJournal`. A full-disk migration failure (`ENOSPC`) would equally fail the journal write, leaving chain data half-moved, the rollback never run, and the original error replaced by the journal error in the logs. The fix wraps the rollback journal write in a best-effort try-catch so the rollback always runs and the original error always propagates.

---

### What's good

**The fix is exactly right.** The reasoning — recording state is bookkeeping, rolling back is data protection, one must not gate the other — is correct. The first journal write (state = `start`) is correctly left unguarded, because failing before anything has moved is safe.

**The test uses referential equality** (`.rejects.toBe(migrationFailure)`) rather than `.rejects.toThrow()`. That verifies the *exact original error object* propagates, not just any error — which is the precise guarantee the fix is trying to provide.

**The error log in the catch includes both `journalError` and `_migrationJournalPath`** — enough for a support engineer to understand what happened without needing to reproduce.

**`chainStorageManagerLayout.realfs.spec.ts`** is documentation-as-tests for every shape the `Logs` directory can take. The dangling symlink case is the important one — it establishes that `fs.ensureDir` raises `ENOENT` rather than creating through a dangling link, which is what makes journal writes able to fail at all. Worth having permanently.

**`beforeEach` uses `fs.realpathSync` on `os.tmpdir()`** — correct, resolves the `/tmp` → `/private/tmp` symlink on macOS so path comparisons don't silently fail.

---

### Minor

The `makeCtx()` in the realfs spec returns `{} as never` with only the two fields journal I/O actually reads. It works, but `as never` is a blunt cast. `as Pick<typeof ctx, '_logsDirectoryPath' | '_migrationJournalPath'>` or a typed partial would be clearer about what's intentionally omitted.

The journal fixture in the realfs spec uses hardcoded POSIX paths (`/mnt/custom-parent` etc.) as JSON values. Since those strings are never resolved against the real filesystem in these tests, they round-trip correctly on Windows too — fine as-is.

---

### Summary

Correct fix, good tests, clear reasoning. Approve as-is; the `as never` cast is the only thing worth mentioning and it's not blocking.

disassembler (approved):
Correct fix, good tests, clear reasoning. Approve as-is.

In the migration catch block, `_writeMigrationJournal` was called unguarded before `_rollbackMigrationJournal`. A full-disk migration failure (`ENOSPC`) would equally fail the journal write, leaving chain data half-moved, the rollback never run, and the original error replaced by the journal error in the logs. The fix wraps the rollback journal write in a best-effort try-catch so the rollback always runs and the original error always propagates.

**The fix is exactly right.** Recording state is bookkeeping, rolling back is data protection — one must not gate the other. The first journal write (state = `start`) is correctly left unguarded since failing before anything has moved is safe.

**The test uses referential equality** (`.rejects.toBe(migrationFailure)`) rather than `.rejects.toThrow()` — verifies the exact original error object propagates, not just any error.

**`chainStorageManagerLayout.realfs.spec.ts`** is documentation-as-tests for every shape the `Logs` directory can take. The dangling symlink case establishes that `fs.ensureDir` raises `ENOENT` rather than creating through a dangling link — which is what makes journal writes able to fail at all.

**Minor:** `makeCtx()` returns `{} as never` — `as Pick<...>` would be clearer but not blocking.

PR #3367, opened 2026-08-06
ci(checks): assert the portable bundle is structurally sound

Crypto2099:
The bundle declares things about itself — the paths its launcher exports, the shared libraries its objects need, the links it contains — and nothing verified those declarations against the directory actually produced. A build can succeed while producing a bundle that cannot start, and no existing check can see that.

## What it asserts

Three properties of a built bundle:

1. **Launcher environment paths resolve.** Every bundle-relative path the launcher exports must exist inside the bundle. The variables are parsed out of the *built* launcher rather than hardcoded, so one added later is covered with no edit here. Values rooted elsewhere are correctly not matched — `XCURSOR_PATH` points at `/usr/share/icons` and is `-d` guarded.
2. **No dangling symlinks.** A broken link is a shipped defect: the macOS signing pass rejects one, and on Linux it surfaces as a missing file at runtime.
3. **The shared library closure is complete.** Every `DT_NEEDED` entry must name a file present in the bundle. A library that resolves only because the build machine happens to have it installed fails only on a machine with no populated Nix store.

## Why there is a self-test

A checker that only ever runs against a bundle that passes is untested in the direction that matters. The self-test builds a minimal fixture bundle, breaks it one way at a time, and requires the checker to reject each break *for the expected reason*:

- a launcher export pointing at something absent
- a dangling symlink
- an unresolvable `DT_NEEDED`
- and the checker refusing to report success when a tool it depends on is missing, rather than skipping every file it was meant to inspect

That last case is the one worth having. An absent `readelf` turns every assertion into a silent no-op, and the check then reports a clean pass over a set it never inspected — worse than no check at all, because it is trusted.

The fixture links statically, so its library closure is genuinely complete rather than dependent on what the builder provides.

## Verification

```
$ nix build .#checks.x86_64-linux.bundle-integrity
SELF-TEST PASSED
Bundle: /nix/store/...-relocatable-electron
ELF objects: 189
1. Launcher environment paths
  ok: XKB_CONFIG_ROOT -> share/X11/xkb
  ok: GBM_BACKENDS_PATH -> lib/electron/lib
2. Dangling symlinks
  ok: none
3. Shared library closure
  ok: every DT_NEEDED resolves inside the bundle
PASSED
```

`shellcheck` and `treefmt` re-run and pass — both see the new files.

Adding the derivation to `checks.<system>` makes it merge-blocking automatically via the `required` aggregate in `flake.nix`, so there is no workflow or matrix to maintain.

## Where the check lives

It is defined in `perSystem/bundle-integrity.nix` rather than in `perSystem/checks.nix`. It is the only check tied to one system's build products — it inspects the Linux relocatable bundle and needs ELF tooling — so keeping it in the shared file meant every other change to the check set collided with it.

flake-parts merges `checks` across perSystem modules, so it lands in the same set, shows up in `nix eval .#checks.x86_64-linux`, and is collected into the required aggregate exactly as before. This branch does not touch `checks.nix` at all.

### DIFF
--- a/perSystem/bundle-integrity.nix
+++ b/perSystem/bundle-integrity.nix
@@ -1,0 +1,42 @@
+# The portable bundle declares things about itself — paths its launcher
+# exports, libraries it needs, links it contains — and nothing verified those
+# declarations against the directory actually produced. Both recent release
+# failures were "build succeeds, application does not start", which no other
+# check in this set can see.
+#
+# This lives in its own module rather than in checks.nix because it is the only
+# check tied to a specific system's build products: it inspects the Linux
+# relocatable bundle and needs ELF tooling. flake-parts merges `checks` across
+# perSystem modules, so it joins the same set and becomes merge-blocking the
+# same way.
+{inputs, ...}: {
+  perSystem = {
+    system,
+    lib,
+    pkgs,
+    ...
+  }: {
+    checks = lib.optionalAttrs (system == "x86_64-linux") {
+      # The self-test runs first: it breaks a fixture bundle one way at a time
+      # and requires the checker to reject each break. A checker that has
+      # silently stopped asserting anything then fails here, instead of passing
+      # every bundle put in front of it.
+      bundle-integrity =
+        pkgs.runCommand "daedalus-bundle-integrity" {
+          # glibc.static: the self-test's fixture binary is linked statically so
+          # the fixture bundle is genuinely self-contained and its library
+          # closure does not depend on what the builder happens to provide.
+          nativeBuildInputs = [pkgs.binutils pkgs.stdenv.cc pkgs.glibc.static];
+        } ''
+          # Invoked through bash rather than executed directly: the sandbox has
+          # no /usr/bin/env for the scripts' shebangs to resolve.
+          bash ${inputs.self}/scripts/check-bundle-integrity-selftest.sh \
+            ${inputs.self}/scripts/check-bundle-integrity.sh
+          echo
+          bash ${inputs.self}/scripts/check-bundle-integrity.sh \
+            ${inputs.self.internal.x86_64-linux.relocatableElectron}
+          touch $out
+        '';
+    };
+  };
+}
--- a/perSystem/checks.nix
+++ b/perSystem/checks.nix
@@ -59,33 +59,6 @@
       jest = mkJsCheck "daedalus-jest" "yarn test:jest --maxWorkers=4";
       cucumber-unit = mkJsCheck "daedalus-cucumber-unit" "yarn test:unit";
       shellcheck = pkgs.callPackage ../tests/shellcheck.nix {src = inputs.self;};
-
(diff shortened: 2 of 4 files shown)


disassembler (changes requested):
Solid PR, one fix needed before merge.

### Issue: dangling symlinks can mask DT_NEEDED failures

`have_lib` is built by indexing basenames of all ELF files *and* all symlinks:

```bash
while IFS= read -r link; do
  have_lib["$(basename "$link")"]=1
done < <(find "$BUNDLE" -type l)
```

`find -type l` returns all symlinks, including dangling ones. So if a dangling symlink happens to be named `libfoo.so.1` and an ELF object has `DT_NEEDED: libfoo.so.1`, check 3 reports it resolved even though check 2 would flag the same link as broken. The two checks can mask each other in that case.

Fix: use `-type l ! -xtype l` to include only non-dangling symlinks, or build `have_lib` from `-xtype f` to include only real files and working symlinks.

---

### What's good

**The self-test is the best part of the PR.** Case 4 in particular — requiring the checker to refuse rather than silently pass when `readelf` is absent — is exactly the failure mode worth preventing.

**Launcher path parsing is generic** — extracts `$BUNDLE_DIR`/`$LIB_DIR`-rooted exports from the built launcher rather than a hardcoded list. If the parsing stops matching it fails with a clear error rather than passing silently.

**`is_elf` via magic bytes** is the right approach. **`recursiveImports`** means no explicit wiring needed. **Correctly `x86_64-linux` only.**

**Minor:** `find "$BUNDLE" -xtype l` for dangling symlinks is Linux-specific — fine since this check is Linux-only.

Crypto2099:
Good catch on the masking — fixed in 3c982c793.

`have_lib` now indexes only links that resolve (`-type l ! -xtype l`). You are right that the bundle would still have failed via assertion 2, but the reported reason would have been the wrong one, and the moment anyone allowlisted a broken link the closure assertion would have gone quietly green over a missing library.

Rather than just fix it, I added it to the self-test, since the whole point of that file is pinning down how the checker fails:

```
ok   does not let a dangling symlink satisfy a DT_NEEDED entry
```

The fixture compiles a dynamically linked binary, then creates a dangling symlink named after each of its `DT_NEEDED` entries. Verified the case genuinely regresses — with the one-line fix reverted:

```
FAIL does not let a dangling symlink satisfy a DT_NEEDED entry — rejected, but not for the expected reason
SELF-TEST FAILED: 1 case(s)
```

`bundle-integrity`, `shellcheck` and `treefmt` pass.

One structural change since you reviewed: the check moved out of `perSystem/checks.nix` into its own `perSystem/bundle-integrity.nix`. It was the only check tied to one system's build products, so keeping it in the shared file made it collide with every other change to the check set. flake-parts merges `checks` across modules, so it lands in the same set and is collected the same way.

disassembler (approved):
Changes look good — the self-test for dangling symlinks is a nice addition, and moving the check into its own module makes sense. Approved.

PR #3368, opened 2026-08-06
ci(checks): run the test suites on macOS, not only x86_64-linux

Crypto2099:
The Jest and Cucumber suites ran on `x86_64-linux` alone. Every platform branch in the code under test — path separators, case sensitivity, symlink and junction handling — was therefore exercised by overriding `process.platform` in-process, which asserts what the code does when *told* it is somewhere else rather than what it does when it is there.

That gap lines up with where defects actually surface: the recent fix history is dominated by path, symlink and drive-mapping behaviour that a `process.platform` override cannot reproduce.

## Two commits, because the first one alone did nothing

**1. Resolve `internal` per-system** instead of pinning it to `x86_64-linux`, so the suites can run wherever the installer pipeline already builds `node_modules` — which is every system here, since the darwin installers depend on it. The prerequisite is already built and cached rather than being new work.

**2. Collect every system's checks into the required aggregate.** `hydraJobs.checks` in `flake.nix` was pinned to `x86_64-linux`:

```nix
checks.x86_64-linux = self.checks.x86_64-linux;
```

so a derivation added to `checks.aarch64-darwin` became a flake output that Hydra never evaluated and `required` never collected. It existed locally, was absent from CI, and nothing reported the discrepancy — a local `nix flake check` and a green PR would disagree with no way to notice.

That is what made commit 1 a no-op on its own: it added `jest` and `cucumber-unit` to both darwin systems, and the resulting Hydra eval contained only the same ten `x86_64-linux` jobs as before.

## Effect

```
required aggregate:   10 constituents -> 18

x86_64-linux:         10 checks, unchanged
aarch64-darwin:        2 -> 4   (+jest, +cucumber-unit)
x86_64-darwin:         2 -> 4   (+jest, +cucumber-unit)
```

Two of the eight additions — `drt-clippy` and `treefmt` on each darwin — are pre-existing derivations that had never been built by CI either. They are green.

## What stays on one platform

`lint`, `compile`, `stylelint`, `i18n`, `storybook` and `shellcheck` read the same source and produce the same answer wherever they run, so repeating them per system would spend darwin builder time for no additional signal. Only the checks that *execute* the code under test gain from a second platform.

## Verification

All 18 check jobs are green on this PR's evaluation, including the four new darwin ones:

```
checks.aarch64-darwin.jest            SUCCESS
checks.aarch64-darwin.cucumber-unit   SUCCESS
checks.x86_64-darwin.jest             SUCCESS
checks.x86_64-darwin.cucumber-unit    SUCCESS
```

The linux side is provably untouched — the `jest` derivation hashes identically before and after:

```
$ nix eval --raw .#checks.x86_64-linux.jest.drvPath
86ljrk4zz4629zyr50iv0kisrwqhyrz5-daedalus-jest.drv   # both
```

Merge latency now follows the slower of the two platforms rather than linux alone. The darwin installers are `nonrequired` and remain so; only the checks are gated.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -73,18 +73,24 @@
           # Hydra never evaluated and `required` never collected — present
           # locally, absent from CI, and silently so.
           checks = lib.genAttrs supportedSystems (system: self.checks.${system});
+          # Only x86_64-linux gates a merge. The darwin checks are built and
+          # reported on every commit, so a failure is visible on the pull
+          # request, but a scarce or flaky mac builder does not hold up work —
+          # the same posture the darwin installers already have.
           required = inputs.nixpkgs.legacyPackages.x86_64-linux.releaseTools.aggregate {
             name = "github-required";
             meta.description = "All jobs required to pass CI";
             constituents =
-              lib.collect lib.isDerivation self.hydraJobs.checks;
+              lib.collect lib.isDerivation self.hydraJobs.checks.x86_64-linux;
           };
           nonrequired = inputs.nixpkgs.legacyPackages.x86_64-linux.releaseTools.aggregate {
             name = "github-nonrequired";
             meta.description = "Jobs built by Hydra but not required to pass CI";
             constituents =
               lib.collect lib.isDerivation self.hydraJobs.installer
-              ++ lib.collect lib.isDerivation self.hydraJobs.devshell;
+              ++ lib.collect lib.isDerivation self.hydraJobs.devshell
+              ++ lib.collect lib.isDerivation
+              (removeAttrs self.hydraJobs.checks ["x86_64-linux"]);
           };
         };
       };
--- a/flake.nix
+++ b/flake.nix
@@ -68,7 +68,11 @@
           devshell = lib.genAttrs supportedSystems (system: self.devShells.${system}.default);
           # Exposing these DLLs for easier development/debugging on Windows:
           nativeModules.x86_64-windows = self.internal.x86_64-windows.nativeModulesZip;
-          checks.x86_64-linux = self.checks.x86_64-linux;
+          # Every system's checks, not only x86_64-linux. Pinned to one system,
+          # a derivation added to `checks.aarch64-darwin` was a flake output
+          # Hydra never evaluated and `required` never collected — present
+          # locally, absent from CI, and silently so.
+          checks = lib.genAttrs supportedSystems (system: self.checks.${system});
           required = inputs.nixpkgs.legacyPackages.x86_64-linux.releaseTools.aggregate {
             name = "github-required";
             meta.description = "All jobs required to pass CI";
--- a/perSystem/checks.nix
+++ b/perSystem/checks.nix
@@ -5,10 +5,11 @@
     pkgs,
     ...
   }: let
-    # JS/TS checks only need to run on one platform — x86_64-linux is cheapest.
     # We reuse the pre-built node_modules from the installer pipeline so there's
-    # no redundant yarn install.
-    internal = inputs.self.internal.x86_64-linux;
+    # no redundant yarn install. Resolved per-system rather than pinned to
+    # x86_64-linux, so a check can run anywhere that pipeline already builds —
+    # which is every system here, since the darwin installers depend on it.
+    internal = inputs.self.internal.${system};
     inherit (internal) nodejs yarn srcWithoutNix node_modules;
 
     mkJsCheck = name: command:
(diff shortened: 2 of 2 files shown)


disassembler:
## Review: PR #3368 — `ci(checks): run the test suites on macOS, not only x86_64-linux`

### Overview

Two-part fix: resolves `internal` per-system so darwin runners can build the checks without a redundant `yarn install`, then widens `hydraJobs.checks` from `x86_64-linux` only to all supported systems so the resulting derivations are actually evaluated by Hydra and collected by `required`. The PR description correctly identifies that commit 1 alone was a no-op — the `hydraJobs` pinning was the real bug.

---

### Issue 1: jest and cucumber-unit run on `x86_64-darwin`

`supportedSystems` is `["x86_64-linux" "x86_64-darwin" "aarch64-darwin"]`, so `lib.genAttrs supportedSystems` runs jest and cucumber-unit on all three — including the slow `x86_64-darwin` runners. The signal from `x86_64-darwin` is redundant with `aarch64-darwin` (same OS, same filesystem semantics, same path behaviour) and the build time cost is real.

Fix:

```nix
checks =
  lib.optionalAttrs (system != "x86_64-darwin") {
    jest = mkJsCheck "daedalus-jest" "yarn test:jest --maxWorkers=4";
    cucumber-unit = mkJsCheck "daedalus-cucumber-unit" "yarn test:unit";
  }
  // lib.optionalAttrs (system == "x86_64-linux") {
    ...
  };
```

### Issue 2: darwin checks block merges

`required` collects via `lib.collect lib.isDerivation self.hydraJobs.checks` — after this PR that includes all systems, so a flaky darwin runner blocks a merge. Mac CI blocking work is a real operational cost.

Darwin checks should be built and reported by Hydra (same model as the installers) but not gate merges. Change `required` to only collect linux, and add darwin to `nonrequired`:

```nix
required = inputs.nixpkgs.legacyPackages.x86_64-linux.releaseTools.aggregate {
  name = "github-required";
  meta.description = "All jobs required to pass CI";
  constituents =
    lib.collect lib.isDerivation self.hydraJobs.checks.x86_64-linux;
};
nonrequired = inputs.nixpkgs.legacyPackages.x86_64-linux.releaseTools.aggregate {
  name = "github-nonrequired";
  meta.description = "Jobs built by Hydra but not required to pass CI";
  constituents =
    lib.collect lib.isDerivation self.hydraJobs.installer
    ++ lib.collect lib.isDerivation self.hydraJobs.devshell
    ++ lib.collect lib.isDerivation self.hydraJobs.checks.aarch64-darwin;
};
```

---

### What's good

The split between "runs per system" and "runs on one system" is exactly right: `jest` and `cucumber-unit` gain real signal from each native platform, while `lint`, `compile`, `stylelint`, `i18n`, `storybook`, `shellcheck` produce identical output everywhere and staying on `x86_64-linux` only is correct.

The drv-hash proof that the linux `jest` derivation is unchanged rules out accidental regression on the existing platform.

The silent-absence bug (`checks.aarch64-darwin` existed locally, was never evaluated by Hydra, nothing reported the discrepancy) is exactly the kind of thing that erodes trust in CI slowly. Worth fixing.

---

### Minor observations

**`--maxWorkers=4` is hardcoded** — consistent with the existing linux check, but darwin builder core counts may differ. Not a blocker.

**`drt-clippy` and `treefmt` on darwin becoming required** as a side effect — worth a note so the next person knows why a darwin clippy failure blocks PRs.

---

### Summary

The intent is right, two changes needed before merge: drop `x86_64-darwin` from the per-system jest/cucumber-unit checks (`aarch64-darwin` gives you the macOS signal at a fraction of the cost), and move darwin checks to `nonrequired` so a flaky mac runner doesn't block merges.

disassembler (changes requested):
The intent is right, two changes needed before merge.

### Issue 1: jest and cucumber-unit run on `x86_64-darwin`

`supportedSystems` is `["x86_64-linux" "x86_64-darwin" "aarch64-darwin"]`, so `lib.genAttrs supportedSystems` runs jest and cucumber-unit on all three — including the slow `x86_64-darwin` runners. The signal from `x86_64-darwin` is redundant with `aarch64-darwin` (same OS, same filesystem semantics, same path behaviour) and the build time cost is real.

Fix:

```nix
checks =
  lib.optionalAttrs (system != "x86_64-darwin") {
    jest = mkJsCheck "daedalus-jest" "yarn test:jest --maxWorkers=4";
    cucumber-unit = mkJsCheck "daedalus-cucumber-unit" "yarn test:unit";
  }
  // lib.optionalAttrs (system == "x86_64-linux") {
    ...
  };
```

### Issue 2: darwin checks block merges

`required` collects via `lib.collect lib.isDerivation self.hydraJobs.checks` — after this PR that includes all systems, so a flaky darwin runner blocks a merge. Mac CI blocking work is a real operational cost.

Darwin checks should be built and reported by Hydra (same model as the installers) but not gate merges. Change `required` to only collect linux, and add darwin to `nonrequired`:

```nix
required = inputs.nixpkgs.legacyPackages.x86_64-linux.releaseTools.aggregate {
  name = "github-required";
  meta.description = "All jobs required to pass CI";
  constituents =
    lib.collect lib.isDerivation self.hydraJobs.checks.x86_64-linux;
};
nonrequired = inputs.nixpkgs.legacyPackages.x86_64-linux.releaseTools.aggregate {
  name = "github-nonrequired";
  meta.description = "Jobs built by Hydra but not required to pass CI";
  constituents =
    lib.collect lib.isDerivation self.hydraJobs.installer
    ++ lib.collect lib.isDerivation self.hydraJobs.devshell
    ++ lib.collect lib.isDerivation self.hydraJobs.checks.aarch64-darwin;
};
```

---

### What's good

The split between "runs per system" and "runs on one system" is exactly right. The drv-hash proof that the linux `jest` derivation is unchanged rules out accidental regression. The silent-absence bug (`checks.aarch64-darwin` present locally, absent from CI) is worth fixing.

**Minor:** `--maxWorkers=4` is hardcoded — consistent with linux, but darwin builder core counts may differ. `drt-clippy` and `treefmt` on darwin becoming required as a side effect is worth a note so the next person knows why a darwin clippy failure blocks PRs.

Crypto2099:
Both points taken, applied in 47634ac09.

**One darwin, not two.** jest and cucumber-unit now run on `x86_64-linux` and `aarch64-darwin` only. Agreed the second darwin run buys nothing — same kernel, same filesystem, and nothing in these suites is architecture dependent.

**Darwin no longer gates.** `required` collects `hydraJobs.checks.x86_64-linux`; everything else in `checks` joins the installers and devshells in `nonrequired`. Built and reported on every commit, visible on the PR, but not holding work behind a scarce mac builder.

I used `removeAttrs self.hydraJobs.checks ["x86_64-linux"]` rather than naming `aarch64-darwin` explicitly, so a system added later lands in `nonrequired` by default. Given this PR exists because a derivation was silently absent from CI, defaulting to "reported" seemed better than defaulting to "invisible".

That also resolves your minor point about `drt-clippy` and `treefmt` on darwin becoming merge-blocking as a side effect — they no longer do, so there is nothing to document.

Measured after the change:

```
required     10 constituents  (x86_64-linux only, as before this branch)
nonrequired  25               (16 installers, 3 devshells, 6 darwin checks)

checks.x86_64-linux     compile cucumber-unit drt-clippy i18n jest lint shellcheck storybook stylelint treefmt
checks.aarch64-darwin   cucumber-unit drt-clippy jest treefmt
checks.x86_64-darwin    drt-clippy treefmt
```

The x86_64-linux `jest` derivation still hashes identically to master, so the existing platform is provably untouched.

On `--maxWorkers=4`: leaving it. It matches the existing linux check, and picking a different number per system is a guess until there is a reason to think the darwin builders are differently sized.

disassembler (approved):
Looks great. The `removeAttrs` approach for defaulting future systems to nonrequired is the right call. Approved.

PR #3369, opened 2026-08-06
test(chain-storage): exercise Windows reparse points on a real Windows runner

Crypto2099:
Every existing test of Windows behaviour in this repository overrides `process.platform` and mocks `fs-extra`. That makes the mock the thing under test: it encodes what the author believed Windows does, and it passes whether or not that belief holds.

## Two beliefs in the tree contradict each other

Daedalus creates a **junction** on Windows:

```ts
// chainStorageManagerShared.ts
await fs.symlink(targetPath, symlinkPath,
  process.platform === 'win32' ? 'junction' : 'dir');
```

Both `captureChainPathState` and `resolveMithrilWorkDir` then handle a junction **twice** — once under `stats.isSymbolicLink()`, and again under `process.platform === 'win32' && stats.isDirectory()`. Only one of those can be the live path for a junction.

Meanwhile `chainStorageWindowsNetwork.spec.ts` mocks it as:

```ts
(fs.lstat as jest.Mock).mockResolvedValue({
  isSymbolicLink: () => true,
  isDirectory: () => false,
});
```

If that mock is right, the `win32 && isDirectory()` branch in each resolver is unreachable for junctions and only ever runs for a plain directory. If it is wrong, the mocked expectations are built on a false premise. Nothing currently distinguishes the two, because the only test of it supplies its own answer.

## What this adds

A Windows-gated spec against real NTFS reparse points:

| Area | Covered |
|---|---|
| `lstat` | how a junction is classified, versus a plain directory |
| `createSymlink` | junction creation, writing through it, and a UNC target — which cannot be a junction and should fail rather than dangle later |
| `resolveMithrilWorkDir` | a live junction, a junction whose target was deleted, and no junction at all |
| drive letters | a junction onto a `subst` drive — a genuine drive-letter path with no network involved, the closest reproduction of the reported mapped-drive case that needs no share |
| `isSamePath` / `isSubPath` | case-insensitive comparison on NTFS, including a sibling with a shared prefix |

### The user-created symbolic link

Chain storage on a network drive is a supported advanced setup that the application cannot set up itself. A junction cannot target a mapped drive, and creating a true symbolic link needs `SeCreateSymbolicLinkPrivilege`, which Daedalus does not have and should not request. The answer is for the user to create the link out of band, elevated:

```
mklink /D "%APPDATA%\Daedalus Mainnet\chain" "Z:\DaedalusChain\chain"
```

Once it exists, the application works against it — so the resolvers must follow a real symbolic link and not only the junction they create themselves, and nothing asserted that. The network target cannot be automated without a share, but the link *type* can: one test creates exactly what `mklink /D` produces and requires `resolveMithrilWorkDir` to follow it to the same place.

`createSymlink` should **not** be changed to attempt a symbolic link — it would fail without elevation, and the junction is correct for the ordinary local case.

The dangling-junction case asserts the result is absolute, because it is handed to `MithrilBootstrapService.setWorkDir()` where a relative value would resolve against `process.cwd()`.

The `subst` test asserts a free drive letter was found rather than skipping if none is — a silent skip there would report coverage of the drive-letter case that does not exist.

## And a way to run it

Windows is the one platform the Nix check set cannot reach: there is no Windows Nix builder, and the Windows installer is cross-built from Linux with wine. So this is a workflow rather than a derivation, and it is the first one in the repository.

It is **deliberately outside the branch protection contexts**. A red run reports rather than blocks. The point is to see what Windows actually does with this code; gating on it is a separate decision.

`--ignore-scripts` on install, for the same reason `nix/internal/x86_64-linux.nix` uses it — the install scripts build native modules and download binaries. No spec imports `node-hid` or `usb`, so nothing under test needs them.

## Also

`ensureDirectoryExists.spec.ts` now creates the link type the product creates for the platform. A `'dir'` symlink needs `SeCreateSymbolicLinkPrivilege` on Windows while a junction needs none, so testing with `'dir'` asserted against a link Daedalus never makes.

## Verification

On Linux the new spec skips in full and the existing suite is unchanged:

```
Test Suites: 1 skipped, 1 passed, 1 of 2 total
Tests:       12 skipped, 4 passed, 16 total
```

`compile`, `lint`, `jest` and `treefmt` all pass. What Windows does is answered by this PR's own run.

### DIFF
--- a/.github/workflows/windows-tests.yml
+++ b/.github/workflows/windows-tests.yml
@@ -1,0 +1,51 @@
+# Windows is the one platform the Nix check set cannot reach. There is no
+# Windows Nix builder here — the Windows installer is cross-built from Linux
+# with wine — so the suites that run on x86_64-linux and both darwins through
+# `checks.<system>` have no equivalent path onto Windows.
+#
+# This is reporting-only by design. It is not in the branch protection contexts,
+# so a red run does not block a merge; the value is seeing what Windows actually
+# does with the reparse-point code, which no amount of overriding
+# `process.platform` on another platform can tell us.
+
+name: Windows tests
+
+on:
+  pull_request:
+  push:
+    branches: [master]
+  workflow_dispatch:
+
+concurrency:
+  group: windows-tests-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  jest:
+    name: Jest on Windows
+    runs-on: windows-latest
+    timeout-minutes: 30
+
+    steps:
+      - uses: actions/checkout@v4
+
+      # Matches nodejs_24 from nix/internal/common.nix, so the Windows run is
+      # on the same major version as every other platform.
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 24
+          cache: yarn
+
+      # --ignore-scripts for the same reason the Nix build uses it
+      # (nix/internal/x86_64-linux.nix): the install scripts build native
+      # modules and download binaries. No spec imports node-hid or usb, so
+      # nothing under test needs them, and building them here would make the
+      # run depend on a working MSVC toolchain for no gain.
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile --ignore-scripts
+
+      # Coverage is on by default and roughly doubles the runtime without
+      # adding signal here — the question is which assertions hold on Windows,
+      # not which lines they touch.
+      - name: Run Jest
+        run: yarn test:jest --coverage=false --ci
--- a/source/main/cardano/utils.spec.ts
+++ b/source/main/cardano/utils.spec.ts
@@ -1,6 +1,12 @@
+import path from 'path';
 import fs from 'fs-extra';
 import { spawnSync } from 'child_process';
 import { createSelfnodeConfig } from './utils';
+
+// The selfnode state paths are built with `path.join` in the code under test,
+// so the expectations are built the same way rather than as POSIX literals.
+const SELFNODE_STATE = '/tmp/selfnode/state';
+const stateFile = (name: string) => path.join(SELFNODE_STATE, name);
 
 const chainStorageManagerMock = {
   unlinkChainEntryPoint: jest.fn(),
@@ -85,10 +91,10 @@ describe('cardano utils', () => {
     ).toBeLessThan(
       chainStorageManagerMock.resetToDefault.mock.invocationCallOrder[0]
     );
-    expect(fs.remove).toHaveBeenCalledWith('/tmp/selfnode/state/wallets');
+    expect(fs.remove).toHaveBeenCalledWith(stateFile('wallets'));
     expect(result).toEqual({
-      configPath: '/tmp/selfnode/state/config.yaml',
-      genesisPath: '/tmp/selfnode/state/genesis.json',
+      configPath: stateFile('config.yaml'),
+      genesisPath: stateFile('genesis.json'),
       genesisHash: 'mock-genesis-hash',
     });
   });
@@ -112,6 +118,6 @@ describe('cardano utils', () => {
       chainStorageManagerMock.unlinkChainEntryPoint
     ).not.toHaveBeenCalled();
     expect(chainStorageManagerMock.resetToDefault).not.toHaveBeenCalled();
-    expect(fs.remove).not.toHaveBeenCalledWith('/tmp/selfnode/state/wallets');
+    expect(fs.remove).not.toHaveBeenCalledWith(stateFile('wallets'));
   });
 });
--- a/source/main/mithril/MithrilBootstrapService.spec.ts
+++ b/source/main/mithril/MithrilBootstrapService.spec.ts
@@ -1,3 +1,4 @@
+import path from 'path';
 import { parseMithrilProgressUpdate } from './mithrilProgress';
 import { MithrilBootstrapService } from './MithrilBootstrapService';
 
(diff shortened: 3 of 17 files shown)


disassembler (changes requested):
### The new spec (`chainStorageWindows.realfs.spec.ts`)

The central diagnostic assertion is exactly right:

```ts
it('reports a junction as a symbolic link, not as a directory', () => {
  const stats = fs.lstatSync(chainPath);
  expect(stats.isSymbolicLink()).toBe(true);
  expect(stats.isDirectory()).toBe(false);
});
```

If this passes, the `win32 && isDirectory()` branches in `captureChainPathState` and `resolveMithrilWorkDir` are dead code for junctions and can be removed. If it fails, the existing mocks are built on a false premise. Either way, the question gets answered.

The network drive test is particularly well-designed — it reproduces the exact reported field defect ("created but corrupt") via a real loopback share, and the assertion captures the precise failure mode:

```ts
expect(linkExists && !linkWorks).toBe(false);
```

Succeeding with a working link *or* failing outright are both acceptable; existing-but-dangling is the one thing that must not happen.

**One issue:** `net share` and `net use` require elevation on Windows. Standard GitHub-hosted `windows-latest` runners run as a non-admin user — the network drive test will likely fail or error with access denied. Either skip it with a note when not elevated, or document that it needs a self-hosted elevated runner.

**Minor:** `findFreeDriveLetter` is defined identically in both the `drive letters` and `mapped network drive` describes. Hoist it to the top of the `onWindows` block.

---

### Existing spec fixes

**`atPath`/`atResolvedPath` helpers are the right pattern** — building path expectations with `path.sep` the same way production code does keeps assertions about logic rather than platform.

**`path.basename(targetPath) === 'foo'` instead of `targetPath.endsWith('/foo')`** is a correct fix — `endsWith` silently passes for any path ending in that suffix.

**`atPath` is duplicated across four spec files.** Extract it to a shared test utility.

**Some `endsWith(atPath(...))` patterns remain** in `MithrilPartialSyncService.spec.ts` for mock dispatch. Where the full path is known, prefer an exact match; `endsWith` is fragile if two paths share a suffix.

Crypto2099:
Two of these are applied in 88cca1035; two I want to push back on, because a run that has already happened contradicts them.

**Applied.** `findFreeDriveLetter` is hoisted. Worth noting the duplication was accidental but the *difference* was not — the two copies used disjoint candidate sets (`YXWVUT` and `NMLKJ`) so the tests cannot select the same letter and interfere. The hoisted helper takes the candidates as a parameter, so that property is now expressed rather than implied.

**Applied.** `atPath` / `atResolvedPath` moved to `source/main/utils/pathAssertions.ts`. `testMatch` is `*.spec.ts` / `*.test.ts`, so a plain module is imported and never run as a suite.

---

**`net share` / `net use` already work on the hosted runner.** GitHub-hosted `windows-latest` runs as an administrator account. The mapped-network-drive test has been executing on every run of this PR, and the current run is fully green:

```
Test Suites: 67 passed, 67 total
Tests:       795 passed, 795 total
PASS source/main/utils/chainStorageWindows.realfs.spec.ts
```

If elevation were unavailable the test would fail rather than skip — `execFileSync('net', ['share', …])` throws, and the letter assertion is `expect(letter).not.toBeNull()` rather than a skip. So no self-hosted runner is needed. Worth keeping in mind for the reverse reason: if a future runner image drops that privilege, this goes red loudly instead of quietly reporting coverage it does not have.

**The remaining `endsWith(atPath(...))` are deliberate, and exact matching is what broke.** This is the one I would not change. The staging root is derived from the managed chain's parent in some fixtures and from the state directory in others, and it reaches the assertion through `path.resolve` — which on Windows *also* qualifies the path with the current drive. So the "full path" is not knowable from the fixture:

```
- Expected  "\tmp\mithril-partial-sync\download"
+ Received  "D:\tmp\mithril-partial-sync\download"
```

Exact matching is what the first three CI rounds used, and it failed precisely this way. Worse, when a mock branch missed, it fell through to a default returning plain strings and the converter then called `entry.isDirectory()` on one — so the symptom was a `TypeError` rather than a path mismatch, which is what made it take three rounds to find.

The tails that remain (`/download/db/ledger`, `/download/db`, `/chain/immutable`) are unique within each fixture, which is the property the branch actually depends on. Happy to add a comment saying so if that reads better than the current one.

Crypto2099:
Rebased on master now that the four chain-storage PRs have landed, and this turned up two things worth separating from the test plumbing.

**Green again: 72 suites, 818 passed, 2 skipped.**

The merged `.realfs` specs are written against POSIX and failed loudly on Windows, which is the correct behaviour and the reason this branch exists. Getting them green produced two findings:

**`checkDiskSpace` shells out on Windows and is slow enough to blow the default five-second test timeout.** Every assertion reaching it timed out — the symlink-resolution case *and* the non-ASCII-path case — which is why it first looked like a symlink problem and then like a unicode problem before the shared cause was obvious. Raising the allowance fixes the tests. What it does not fix is that `validateChainStorageDirectory` is a user-facing path paying a subprocess cost on every call, on the platform where most reported storage defects originate. Flagging rather than fixing here.

**A dangling chain symlink is reported as `unknown` on Windows, not `path-not-found`.** That is the same shape as the `EACCES` mislabelling #3361 just fixed: a real, specific condition surfacing as "Unable to validate selected directory" instead of the message that exists for it. I gated it with the observed behaviour recorded next to the gate rather than adapting the assertion, because rewriting it would hide the question instead of answering it. Unexplained, and a candidate defect.

**`EXDEV` keeps its coverage on Windows** rather than being skipped. No `/dev/shm`, but a rename across drive letters raises `EXDEV` too, so the candidate roots are per-platform and the fallback branch is exercised on both.

Only two assertions remain POSIX-gated: the dangling-symlink case above, and the read-only case, which cannot be set up without an ACL since mode bits do not deny writes on Windows.

One correction to my own previous commit: I had gated the symlink-resolution case as a platform difference. It was not — it was the same timeout. It is back out of the gate and running on Windows.

disassembler (approved):
Pushbacks both accepted — CI evidence on the `net share`/`net use` point is conclusive, and the drive-letter prefix issue explains exactly why exact matching failed. The `endsWith` tails being unique within each fixture is a reasonable property to rely on.

On `checkDiskSpace`: agreed it's out of scope here, but the subprocess cost on every user-facing validation call on Windows is worth tracking. Longer term I'd like to see that logic move into the watchdog process so it runs off the hot path — flagging for now.

Approved.

PR #3374, opened 2026-08-07
feat(drt): verify release provenance before uploading anything

Crypto2099:
Publishing a release asserts that a version corresponds to a commit. Nothing established that assertion — `drt release` uploaded whatever was in the directory it was pointed at, and `release-cli` had no release validation of any kind.

## Why the check is at publish time and not at tag time

A tag is reachable from its branch when it is created, and can stop being reachable afterwards: a history rewrite orphans one silently. The artifacts are built after that point. So "is the tag an ancestor of the branch" answers the question at a moment when it is not yet possible to get the wrong answer.

Comparing the revision recorded in the artifacts against the commit the tag resolves to settles it at the only moment that matters. It is indifferent to when the tag was created and unaffected by a rewrite.

## What it refuses

Before anything is hashed, signed or uploaded:

| Condition | Why |
|---|---|
| An installer name yields no revision | The artifact's provenance cannot be established |
| Installers do not all carry the same revision | A directory holding two builds cannot be released as one |
| The revision is the `dirty` marker | The source tree was not clean, so the commit is unknown and the build unreproducible |
| `meta.json` records a different `gitrev` | The directory holds artifacts from a different build than its metadata describes |
| The tag resolves to a different commit | Publishing would claim a version corresponds to a commit it was not built from |
| The tag does not exist | There is nothing to check the claim against |

## Resolving the tag

`git ls-remote`, so no checkout is required. `refs/tags/<v>^{}` is requested alongside `refs/tags/<v>` so an annotated tag yields the commit rather than the tag object.

- `--repo-url` overrides the remote (defaults to this repository)
- `--skip-tag-check` disables that half while leaving the artifact-consistency checks in force. It is deliberately awkward to reach, and prints that it was used.

## Reading the revision out of a filename

Names are `daedalus-<version>-<build>-<cluster>-<revShort>-<system>.<ext>` (`nix/internal/any-darwin.nix:26`). Two traps, both covered by tests:

- **The build counter is decimal, so it is also valid hex.** `daedalus-11.2.0-86185-…` — searching for "the field that looks like a hash" can return `86185`. Position disambiguates; pattern matching does not.
- **The system component contains a dash**, and a prerelease version may too, so the name cannot be parsed by counting fields from the left. It is parsed from the right against the known target systems.

## Verification

Unit tests, run under the pinned toolchain — 9 pass, 7 new:

```
$ nix develop .#release-cli --command cargo test --locked
test provenance::tests::build_counter_is_not_mistaken_for_a_revision ... ok
test provenance::tests::version_containing_a_dash_is_handled ... ok
test provenance::tests::rev_from_every_target_system ... ok
test provenance::tests::short_and_full_revisions_compare_equal ... ok
test provenance::tests::unrecognised_names_yield_nothing ... ok
test provenance::tests::dirty_tree_marker_is_read_back ... ok
test provenance::tests::rev_from_linux_installer_name ... ok
test result: ok. 9 passed; 0 failed
```

End to end against a built `drt` and a real tag:

```
=== matching artifacts and tag ===
  provenance: 8f9737937 matches tag 11.2.0 (8f9737937d346e847c9c29c965fba7aa44136612)

=== artifacts from a different commit than the tag ===
Error: tag 11.2.0 points at 8f9737937…, but the installers were built from fa8258757.
Publishing would claim a version corresponds to a commit it was not built from.

=== two commits in one directory ===
Error: installers were built from more than one commit, so they cannot be released together:
  8f9737937: daedalus-…-x86_64-linux.bin
  fa8258757: daedalus-…-x86_64-darwin.pkg

=== dirty source tree ===
Error: installers were built from a dirty source tree, so the commit they correspond to
is unknown and unreproducible
```

`drt-clippy`, `treefmt` and `shellcheck` all pass. Note that local `cargo` cannot build this crate — the AWS SDK requires rustc 1.91.1 and the system toolchain here is 1.88.0 — so both clippy and the tests must be run through Nix, which supplies the pinned 1.95.0 toolchain.

`cmd_release`'s parameters moved into a struct: nine positional arguments trip `clippy::too_many_arguments`, and grouping them reads better than an `allow`.

### DIFF
--- a/release-cli/src/cli.rs
+++ b/release-cli/src/cli.rs
@@ -50,18 +50,15 @@ pub enum Commands {
         #[arg(long)]
         skip_version_json: bool,
 
-        /// Repository to resolve the release tag against, for checking that the
-        /// installers were built from the commit the tag names.
-        #[arg(long, default_value = "https://github.com/input-output-hk/daedalus")]
-        repo_url: String,
+        /// Repository, as owner/name, whose tags name the release commits.
+        /// Resolved over the GitHub API; GITHUB_TOKEN is used when set.
+        #[arg(long, default_value = "input-output-hk/daedalus")]
+        repo: String,
 
-        /// Skip resolving the release tag and comparing it against the commit
-        /// the installers were built from.
+        /// Do not resolve the release tag; check only that the installers are
+        /// internally consistent.
         ///
-        /// The installers are still checked for internal consistency. Use only
-        /// when the tag legitimately does not exist yet — a release whose tag
-        /// names a different commit than its artifacts is the failure this
-        /// check exists to prevent.
+        /// Useful when the tag has not been pushed yet.
         #[arg(long)]
         skip_tag_check: bool,
     },
--- a/release-cli/src/cli.rs
+++ b/release-cli/src/cli.rs
@@ -49,6 +49,21 @@ pub enum Commands {
         /// Useful for staging binaries before making them the latest version.
         #[arg(long)]
         skip_version_json: bool,
+
+        /// Repository to resolve the release tag against, for checking that the
+        /// installers were built from the commit the tag names.
+        #[arg(long, default_value = "https://github.com/input-output-hk/daedalus")]
+        repo_url: String,
+
+        /// Skip resolving the release tag and comparing it against the commit
+        /// the installers were built from.
+        ///
+        /// The installers are still checked for internal consistency. Use only
+        /// when the tag legitimately does not exist yet — a release whose tag
+        /// names a different commit than its artifacts is the failure this
+        /// check exists to prevent.
+        #[arg(long)]
+        skip_tag_check: bool,
     },
 
     /// Code-sign macOS (.pkg) and/or Windows (.exe) installers via SSH
--- a/release-cli/src/main.rs
+++ b/release-cli/src/main.rs
@@ -71,7 +71,7 @@ async fn main() -> Result<()> {
             dry_run,
             test,
             skip_version_json,
-            repo_url,
+            repo,
             skip_tag_check,
         } => {
             cmd_release(ReleaseOptions {
@@ -82,7 +82,7 @@ async fn main() -> Result<()> {
                 dry_run,
                 test,
                 skip_version_json,
-                repo_url: &repo_url,
+                repo: &repo,
                 skip_tag_check,
             })
             .await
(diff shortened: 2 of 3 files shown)


disassembler:
## Review: PR #3374 — `feat(drt): verify release provenance before uploading anything`

### Overview

This adds a pre-upload provenance gate to `drt release`. Before hashing, signing, or uploading, it verifies that all installers in the directory came from the same commit, and that commit matches the release tag. The new `provenance.rs` module handles filename parsing and consistency checks. The concept is sound and the tests are solid.

---

### Critical: `resolve_tag_commit` is architecturally incompatible with how `drt` works

`drt` is explicitly designed to run without a local git repository. It fetches artifacts from Hydra, signs them on remote hosts (macOS via SSH to `OSX_SIGN_HOST`; Windows via a remote HSM), and publishes to S3. No git context exists or is assumed on the machine running `drt`. The existing codebase has zero git dependency.

`resolve_tag_commit` (`main.rs:369–416`) breaks that design:

```rust
let output = std::process::Command::new("git")
    .args(["ls-remote", repo_url, ...])
    .output()
```

Replacing the shell-out with `gix` or `git2` doesn't fix this — the problem isn't how git is invoked, it's that git is involved at all.

**The right approach** uses two HTTP calls `drt` can already make, requiring no auth token for a public repo:

**1. GitHub REST API to resolve the tag:**
```
GET https://api.github.com/repos/input-output-hk/daedalus/git/ref/refs/tags/{version}
```
Returns `object.sha` and `object.type`. If `type == "tag"` (annotated), peel it:
```
GET https://api.github.com/repos/input-output-hk/daedalus/git/tags/{sha}
```
No token required under normal rate limits for a public repo.

**2. Hydra already recorded the gitrev.** `drt fetch-installers` writes `meta.json` from Hydra's eval metadata — that `gitrev` field is the authoritative provenance record for the build. The consistency check against `meta.json` in `provenance.rs` is already there; the tag resolution should be verified against it via the GitHub API rather than via git.

This also improves the comparison: the GitHub API returns a full 40-character SHA, as does Hydra's `gitrev`, so the check is unambiguous — no prefix-matching required.

---

### Secondary: blocking network call in async context

`std::process::Command::output()` blocks the executor thread for the duration of the network round-trip. `cmd_release` is `async fn` running on Tokio. This should be `tokio::process::Command` or wrapped in `tokio::task::spawn_blocking`. Worth fixing regardless of how tag resolution is implemented.

---

### Minor issues

**`same_commit` accepts arbitrarily short prefixes** (`provenance.rs:67–75`): a 1-character string passes. In practice the filename format enforces 9+ hex chars so this can't be triggered through normal use, but a minimum-length guard makes the invariant explicit. Largely moot once both sides come from APIs returning full SHAs, but the function will still be used to compare the filename-parsed short rev against `meta.json`.

**`KNOWN_SYSTEMS` must be kept in sync manually** (`provenance.rs:21–26`): a comment linking to `nix/internal/any-darwin.nix:26` (already cited in the PR description) belongs in the source.

**`--skip-tag-check` is "deliberately awkward to reach"**: if the option is legitimate — and it is, for a release whose tag hasn't been pushed yet — it should be documented normally. The warning print on use is sufficient friction.

---

### What's good

- **Parsing the rev from the filename is worth keeping** even alongside `meta.json` and the tag check. If someone manually assembled the directory or something got corrupted, catching a mismatch between the filename-embedded rev and `meta.json` is exactly the kind of defence-in-depth that matters here.
- Parsing from the right against known system suffixes is the correct approach; the tests cover every documented trap (build counter ambiguity, dash in version, all four target systems).
- `BTreeMap` for deterministic error output is a nice touch.
- Grouping by revision so a mismatch lists every file on each side, not just the first, produces actionable errors.
- The `ReleaseOptions` struct refactor is a clean improvement regardless of this feature.

---

### Summary

The consistency checks in `provenance.rs` are correct and worth keeping as-is. The only thing that needs changing is `resolve_tag_commit`: replace the `git ls-remote` shell-out with a GitHub REST API call. `git` is unavailable in `drt`'s operating environment and the infrastructure for HTTPS requests is already there.

Crypto2099:
Thanks — the core objection is right and is fixed in 89d02eb3a. Two notes on the reasoning, since the detail matters for anyone reading this later.

**The environment, not the repository, is what broke it.** `git ls-remote <url>` operates against a remote and needs no local repository, so "designed to run without a local git repo" would not by itself have defeated it. What does defeat it is that the `ops` shell is `packages = [config.packages.drt pkgs.gnupg]` — the git *binary* is simply absent from the environment the tool is meant to run in. Same conclusion, different reason, and the second one is the one worth writing down.

**The suggested endpoint 404s.** `git/ref/refs/tags/{tag}` returns:

```
{"message": "Not Found", "status": "404"}
```

GitHub's `git/ref/{ref}` wants the ref *without* the `refs/` prefix:

```
GET /repos/input-output-hk/daedalus/git/ref/tags/11.2.0
→ ref: refs/tags/11.2.0   type: tag   sha: 4a66621d…

GET /repos/input-output-hk/daedalus/git/tags/4a66621d…
→ commit 8f9737937d346e847c9c29c965fba7aa44136612
```

That is what is implemented: resolve, and peel when `object.type == "tag"`.

**Also addressed:**

- The blocking call is gone — being async, this no longer holds a Tokio worker across a network round trip.
- `same_commit` requires seven characters before treating one revision as an abbreviation of another. Installer names carry nine, so nothing shorter came from this tool. Tests cover the short-prefix rejection.
- `KNOWN_SYSTEMS` now carries the `nix/internal/any-darwin.nix:26` citation in the source rather than only in the PR description.
- `--skip-tag-check` is documented plainly as being for a tag that has not been pushed yet, and the editorialising is removed.
- `--repo-url` became `--repo`, taking `owner/name`, which is what the API addresses. `GITHUB_TOKEN` is used when set — unauthenticated is 60 requests/hour and shared by IP, which is fine for a release but not for CI.

**Verification.** Rebuilt and run with an empty `PATH`, so no git binary was reachable at all:

```
git available? NO

=== artifacts match the tag ===
  provenance: 8f9737937 matches tag 11.2.0 (8f9737937d346e847c9c29c965fba7aa44136612)

=== artifacts from a different commit than the tag ===
Error: tag 11.2.0 points at 8f9737937…, but the installers were built from fa8258757.

=== tag that does not exist ===
Error: tag 99.9.9 does not exist in input-output-hk/daedalus.
```

Filename parsing stays, as you suggested — it is the only check that catches a directory assembled by hand or a partially-replaced artifact, which neither `meta.json` nor the tag can see.

`drt-clippy` and `treefmt` pass; 9 tests pass under the pinned toolchain.

disassembler (approved):
All concerns addressed. The implementation is correct.

The endpoint note in the code (`git/ref/tags/<tag>`, not `git/ref/refs/tags/<tag>`) is exactly right — that was an error in the review. The two-step peel (lightweight tag returns directly, annotated tag fetches the tag object) is correct, `fetch_git_object` returning `None` for 404 and `Err` for anything else is a clean pattern, and the implementation is fully async.

Good catch on the precise reason git was unavailable — "no local repo" was imprecise; the real reason is that the `ops` shell simply doesn't include the git binary. Same conclusion, but the one worth writing down.

`GITHUB_TOKEN` support is a sensible addition — unauthenticated is fine for a one-time release, but shared-IP rate limits in CI would be a problem.

PR #3376, opened 2026-08-08
fix(explorer): build explorer links for the current explorer.cardano.org layout

### LINKED ISSUE #3375: Bug: Transaction ID Links are malformed for the current explore.cardano.org
Clicking on a link in the transaction explorer (preprod edition shown) generates the following link:
https://explorer.cardano.org/txf064cf1c1255f9a696547516e0f61c32a1444a503722e91a5be815ad25a02276

Image

This generates an error via the explore.cardano.org interface

Image

The correct format should be as follows:

https://explorer.cardano.org/preprod/tx/f064cf1c1255f9a696547516e0f61c32a1444a503722e91a5be815ad25a02276

```
root_url = 'https://explorer.cardano.org'
network_discriminator = 'mainnet|preview|preprod'
discovery_type = 'address|tx'
identifier = 'address_bech32|tx_id'

explorer_link = root_url + '/' + network_discriminator + '/' + discovery_type + '/' + identifier

### PR DESCRIPTION
**Pull request by MFA-G**
Closes #3375.

## Problem

Clicking a transaction (or address) link in the wallet builds a URL for the *legacy* Cardano Explorer routes:

```
https://explorer.cardano.org/en/transaction?id=f064cf1c...
https://explorer.cardano.org/en/address.html?address=addr1...
```

The current explorer.cardano.org does not serve those routes any more, so the link lands on an error page instead of the transaction. As reported in the issue, the layout it expects is:

```
{root}/{network}/{tx|address}/{identifier}
network = mainnet | preview | preprod
```

## Fix

`getNetworkExplorerUrlByType` in `source/renderer/app/utils/network.ts` now emits the current layout:

| network | before | after |
|---|---|---|
| `mainnet` | `https://explorer.cardano.org/en/transaction?id=<tx>` | `https://explorer.cardano.org/mainnet/tx/<tx>` |
| `preprod` | `http://explorer.cardano.org/tx/<tx>` | `https://explorer.cardano.org/preprod/tx/<tx>` |
| `preview` | `http://explorer.cardano.org/tx/<tx>` | `https://explorer.cardano.org/preview/tx/<tx>` |
| `testnet` | *(unchanged)* | `https://explorer.cardano-testnet.iohkdev.io/en/transaction?id=<tx>` |
| `staging` | *(unchanged)* | `http://explorer.staging.cardano.org/tx/<tx>` |

`testnet` still points at the separate legacy host `explorer.cardano-testnet.iohkdev.io`, so it deliberately keeps the old locale-prefixed query-string routes — only the networks served by explorer.cardano.org moved to the new layout. This also removes the two `@ts-ignore ts-migrate(2322)` FIXMEs, since the type-narrowing hack they suppressed is gone.

### Secondary fix: protocol for preprod/preview

`getNetworkExplorerUri` maps every unrecognised network (including `preprod` and `preview`) to the mainnet explorer host, but `getNetworkExplorerUrl` picked the protocol from an allowlist of *network names* (`mainnet`/`testnet`/`development`). So preprod and preview resolved to the HTTPS-only mainnet host yet were linked over plain `http://`.

The protocol is now derived from the resolved **host** rather than the network name, so it cannot drift out of sync with `getNetworkExplorerUri` again:

```ts
const uri = getNetworkExplorerUri(network);
const protocol = uri === STAGING_EXPLORER_URL ? 'http://' : 'https://';
```

Staging is the only explorer host served over plain HTTP; behaviour there is unchanged.

## Validation

- `tests/common/unit/networks.spec.ts` extended with 7 cases covering mainnet/preprod/preview/testnet/staging and the unknown-network fallback, for both `tx` and `address`. The unknown-network case is what caught the `http://` bug above.
- `jest tests/common/unit/networks.spec.ts` — 11 passed
- `jest` (full suite) — **71 suites, 810 tests, all passing**, no regressions
- `tsc --noEmit` — clean, 0 errors
- `eslint source/renderer/app/utils/network.ts` — clean
- Prettier — both touched files formatted

Happy to adjust the network mapping if preprod/preview are meant to resolve to a different explorer host than mainnet.

### DIFF
--- a/source/renderer/app/components/wallet/paper-wallet-certificate/CompletionDialog.tsx
+++ b/source/renderer/app/components/wallet/paper-wallet-certificate/CompletionDialog.tsx
@@ -7,7 +7,7 @@ import SVGInline from 'react-svg-inline';
 import { Link } from 'react-polymorph/lib/components/Link';
 import { LinkSkin } from 'react-polymorph/lib/skins/simple/LinkSkin';
 import Dialog from '../../widgets/Dialog';
-import { getNetworkExplorerUrl } from '../../../utils/network';
+import { getNetworkExplorerUrlByType } from '../../../utils/network';
 import styles from './CompletionDialog.scss';
 // @ts-ignore ts-migrate(2307) FIXME: Cannot find module '../../../assets/images/clipboa... Remove this comment to see the full error message
 import iconCopy from '../../../assets/images/clipboard-ic.inline.svg';
@@ -120,9 +120,11 @@ class CompletionDialog extends Component<Props, State> {
         onClick: onClose,
       },
     ];
-    const cardanoExplorerLink = `${getNetworkExplorerUrl(
+    const cardanoExplorerLink = getNetworkExplorerUrlByType(
+      'address',
+      walletCertificateAddress,
       network
-    )}/address/${walletCertificateAddress}`;
+    );
     // Get QRCode color value from active theme's CSS variable
     const qrCodeBackgroundColor = document.documentElement
       ? document.documentElement.style.getPropertyValue(
--- a/source/renderer/app/components/wallet/paper-wallet-certificate/InstructionsDialog.tsx
+++ b/source/renderer/app/components/wallet/paper-wallet-certificate/InstructionsDialog.tsx
@@ -142,7 +142,7 @@ class InstructionsDialog extends Component<Props> {
     ];
 
     const openNetworkExplorer = () =>
-      onOpenExternalLink(getNetworkExplorerUrl(network));
+      onOpenExternalLink(getNetworkExplorerUrl());
 
     const cardanoExplorerLink = (
       <Link
--- a/source/renderer/app/config/urlsConfig.ts
+++ b/source/renderer/app/config/urlsConfig.ts
@@ -1,8 +1,6 @@
 import coingeckoConfig from './currencyConfig.coingecko';
 
 export const MAINNET_EXPLORER_URL = 'explorer.cardano.org';
-export const STAGING_EXPLORER_URL = 'explorer.staging.cardano.org';
-export const TESTNET_EXPLORER_URL = 'explorer.cardano-testnet.iohkdev.io';
 
 // Build-time configurable newsfeed URLs. Override NEWS_URL / NEWS_HASH_URL at
 // build time (via webpack EnvironmentPlugin defaults in webpack.config.js) to
@@ -55,8 +53,6 @@ export const STAGING_NEWS_HASH_URL = NEWS_HASH_HOSTNAME;
 export const CATALYST_API_URL = 'core.projectcatalyst.io';
 export const ALLOWED_EXTERNAL_HOSTNAMES = [
   MAINNET_EXPLORER_URL,
-  STAGING_EXPLORER_URL,
-  TESTNET_EXPLORER_URL,
   NEWS_HOSTNAME,
   NEWS_HASH_HOSTNAME,
   coingeckoConfig.hostname,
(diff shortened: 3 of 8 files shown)


Crypto2099:
This change seems straightforward enough and an easy lift.

Two things stand out:
1. The legacy testnet explorer is no longer functional and should be removed (may be part of a larger issue of identifying and removing this from the codebase)
2. The staging explorer is similarly non-functional and should be removed
3. Tests/behavior should be defined when no explicit network is provided (default to mainnet maybe?)
4. Tests should, if possible, make a real request to the explorer URLs with real network data (tx/address) so we can detect drift automatically in the future via CI

MFA-G:
Thanks for the review @Crypto2099 — all four points addressed in a96050e.

**1 & 2. Legacy testnet and staging explorers removed.** Both hosts are dead (`NXDOMAIN` for `explorer.cardano-testnet.iohkdev.io` and `explorer.staging.cardano.org`), so keeping special cases for them just preserved broken links. `getNetworkExplorerUri` now always returns `explorer.cardano.org`, the locale-prefixed `?id=` / `?address=` routes are gone, the `http://` fallback goes with them (nothing is served over plain HTTP anymore), and `STAGING_EXPLORER_URL` / `TESTNET_EXPLORER_URL` are dropped from `urlsConfig` including `ALLOWED_EXTERNAL_HOSTNAMES`. That shrinks `network.ts` by ~30 lines.

**3. Default behaviour is now explicit.** Added `getNetworkExplorerPath()` and an exported `DEFAULT_EXPLORER_NETWORK_PATH`. Networks the explorer doesn't serve (`staging`, `development`, `selfnode`, the retired `testnet`) and an undefined/empty `env.NETWORK` all resolve to `mainnet`, and there are tests asserting exactly that for each case rather than leaving it implicit.

While in there I also routed the two remaining ad-hoc call sites (`StakingRewardsPage`, paper-wallet `CompletionDialog`) through `getNetworkExplorerUrlByType` — they were building `${getNetworkExplorerUrl(network)}/address/${addr}` by hand, which is exactly the URL shape this PR is fixing, so they'd have kept producing network-less links. Every explorer URL is now built in one function.

**4. Drift check against the real explorer.** Added an opt-in suite that requests the URLs the wallet actually opens, using a real mainnet tx and address:

```
EXPLORER_DRIFT_CHECK=1 yarn test:jest tests/common/unit/networks.spec.ts
```

It's `describe.skip` by default so the normal suite stays offline and hermetic; CI can flip the env var on a scheduled job to detect layout drift without making PR runs flaky on explorer downtime. Verified locally against the live site:

```
The explorer URLs the wallet opens:
  ✓ resolve for a real mainnet transaction (130 ms)
  ✓ resolve for a real mainnet address (174 ms)
  ✓ resolve for the preprod and preview explorers (87 ms)
```

Full validation:
- `yarn jest tests/common/unit` — 18 passed (15 offline + 3 network, with the env var on)
- `yarn compile` (`tsc --noEmit`) — clean
- `eslint` on all touched files — 0 errors

Let me know if you'd prefer the drift check wired into a CI workflow in this PR rather than left opt-in.

disassembler (changes requested):
The fix for mainnet/preprod/preview routes is correct. Request changes on testnet handling: the PR description claims testnet is "unchanged" but the code removes \`TESTNET_EXPLORER_URL\` entirely and testnet now falls through to \`https://explorer.cardano.org/mainnet/...\`. If the testnet explorer is confirmed dead, update the description to say so and add a test documenting the deliberate fallback. If it's still live, the old routing needs to be preserved.

disassembler (approved):
Retracting my earlier request-changes — I misread the testnet handling. The code is correct: `EXPLORER_NETWORK_PATHS` explicitly maps `mainnet`, `preprod`, and `preview` to their respective paths on `explorer.cardano.org`. The old `TESTNET_EXPLORER_URL` (`explorer.cardano-testnet.iohkdev.io`) and `STAGING_EXPLORER_URL` are genuinely dead; the fallback to the mainnet path for those is intentional and the comment in the code says exactly that. The `currentLocale` removal is also correct — the new explorer layout doesn't use locale-scoped paths. Tests cover all cases and the opt-in drift check is a nice touch. Approved.

PR #3378, opened 2026-08-10
test(chain-storage): cover symlink chains, loops and state-directory nesting

Crypto2099:
The real-filesystem specs covered a single link to a directory and a link whose
target had been removed. Four further cases are ones a mocked filesystem cannot
express, because a mock decides the resolution rather than observing it.

## What is covered

| Case | Why it matters |
|---|---|
| A link pointing at another link | What moving the chain directory twice leaves behind. `realpath` walks the whole chain, so both the resolver and validation report the directory at the end of it. |
| Two links pointing at each other | `realpath` fails, so the resolver takes the same fallback a dangling link takes. Validation reports the location as not found, which is accurate: there is no directory there. |
| A link outside the state directory that resolves inside it | The nesting check runs on the resolved path, so the rejection has to follow where the link lands rather than where it sits. |
| A link that resolves onto the managed chain directory | One of the two ways a user asks for the default location back, and the branch that compares resolved paths rather than literal ones. |

## Verification

Both loop cases assert that `realpath` fails before relying on the fallback, so
neither can pass without exercising it:

```ts
expect(() => fs.realpathSync(chainPath)).toThrow();
```

The loop assertions require an absolute path rather than a throw. That result is
handed to `MithrilBootstrapService.setWorkDir()`, which has no defence against
either.

Only one of the new cases reaches `checkDiskSpace`, and it carries the longer
timeout the Windows runner needs. The rest return before the free-space read.

`yarn test:jest` goes from 820 tests to 828, all passing. `yarn lint`,
`yarn compile` and `nix fmt -- --ci` are clean.

### DIFF
--- a/source/main/utils/chainStoragePathResolver.realfs.spec.ts
+++ b/source/main/utils/chainStoragePathResolver.realfs.spec.ts
@@ -57,6 +57,41 @@ describe('resolveMithrilWorkDir against a real filesystem', () => {
     await expect(resolveMithrilWorkDir(stateDir)).resolves.toBe(target);
   });
 
+  // A user who moves the chain directory twice leaves a link pointing at a link.
+  // `realpath` walks the whole chain, so the resolver should report the directory
+  // at the end of it rather than the intermediate link.
+  it('resolves a chain of symlinks to the directory at the end of it', async () => {
+    const target = path.join(tmpRoot, 'final-target');
+    const intermediate = path.join(tmpRoot, 'intermediate-link');
+    fs.mkdirSync(target);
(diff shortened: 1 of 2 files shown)


disassembler (approved):
Precondition assertions on the loop cases (`expect(() => fs.realpathSync(...)).toThrow()`) are exactly right — without them the fallback tests pass vacuously. Nesting-via-symlink and the two managed-chain-dir reset cases are the kind of thing only a realfs test can cover. Timeout scope is correct: only the chain-of-symlinks case reaches `checkDiskSpace`. Approved.

PR #3379, opened 2026-08-10
fix(chain-storage): report an unresolvable link as not found, not unknown

Crypto2099:
Selecting a chain directory whose link target has been removed produced "Unable
to validate selected directory", the generic message, rather than "Selected
directory does not exist", which the renderer already has copy for.

## The two platforms fail at different syscalls

On POSIX the existence probe follows the link, finds nothing, and returns the
specific reason before anything else runs:

```ts
const exists = await fs.pathExists(normalizedPath);
if (!exists) {
  return { ...defaultValidation, reason: 'path-not-found', ... };
}
```

On Windows the reparse point itself satisfies that probe. The resolution on the
next line is what fails, and that error had nowhere to go but the generic catch
at the end of the function, which reports `unknown`.

Resolution failures are now classified. A missing path reports `path-not-found`;
anything else is rethrown so it keeps reaching the generic handler. An `EACCES`
on an ancestor means the location may well exist, and reporting that as not
found would send the user looking for the wrong problem. A spec covers that
distinction alongside the fix.

## Verification

The mocked spec covers the resolution failure directly, because a POSIX
filesystem will not produce that state on demand. Against unpatched code it
fails with the symptom described:

```
● validateChainStorageDirectory › returns path-not-found when the target exists but cannot be resolved

    - ObjectContaining {
    + Object {
        "isValid": false,
    +   "message": "Unable to validate selected directory.",
        "path": "/mnt/dangling",
    -   "reason": "path-not-found",
    +   "reason": "unknown",
      }
```

The real-filesystem assertion no longer needs its POSIX gate, since both
platforms now report the same reason for the same fixture.

## A looping path is the same condition

`ELOOP` is classified alongside the two missing-path codes. A path that loops
back on itself names no directory either, and it was reaching the same generic
catch and reporting the same generic message. `EACCES` stays out of the
classification for the reason it always did.

### DIFF
--- a/source/main/utils/chainStorageValidation.realfs.spec.ts
+++ b/source/main/utils/chainStorageValidation.realfs.spec.ts
@@ -49,25 +49,18 @@ const expectWritesDenied = (dir: string) => {
   expect(() => fs.accessSync(dir, fs.constants.W_OK)).toThrow();
 };
 
-// Three assertions below do not hold on Windows, for reasons that are about
-// the platform rather than the fixture. They are gated rather than adapted,
-// because two of them look like real defects and adapting the test would hide
-// the question rather than answer it:
+// One assertion below does not hold on Windows, and the reason is the platform
+// rather than the fixture: POSIX mode bits do not deny writes there, so the
+// read-only case cannot be set up at all without an ACL (`icacls /deny`). It is
+// gated rather than adapted.
 //
-//   * POSIX mode bits do not deny writes on Windows, so the read-only case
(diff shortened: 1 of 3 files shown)


disassembler (approved):
Fix is correct: inner try/catch wraps \`fs.realpath()\`, maps \`ENOENT\`/\`ENOTDIR\`/\`ELOOP\` → \`path-not-found\`, rethrows everything else. Dangling-symlink realfs test correctly promoted out of \`describeOnPosix\`. Approved.

PR #3380, opened 2026-08-10
fix(chain-storage): do not block directory validation on the free space probe

Crypto2099:
`checkDiskSpace` spawns a subprocess on Windows, and a cold start there costs
seconds. `validateChainStorageDirectory` runs it on every call, with the user
waiting on the directory picker, and awaited it without any bound:

```ts
const { free } = await checkDiskSpace(resolvedValidationPath);
if (free < requiredSpace) { ... }
```

Two consequences. A slow probe holds the dialog for as long as it takes. A
probe that throws sends the whole validation into the generic catch, so a disk
whose free space cannot be read is reported as "Unable to validate selected
directory", which is not what went wrong.

## What changed

The probe is raced against a timeout, and a probe that times out or fails
yields no figure rather than a verdict:

```ts
const free = await readFreeSpace(resolvedValidationPath, diskSpaceTimeout);

if (free != null && free < requiredSpace) {
  return { ...defaultValidation, reason: 'insufficient-space', ... };
}
```

Only a figure that was actually read can reject a directory. The selection is
otherwise accepted without a free-space figure.

This is the position `handleDiskSpace` already takes for the periodic check,
where an unreadable probe is logged and cardano-node is started anyway. The
allowance here is 5 seconds rather than the 9 that check uses, because the
periodic check has nobody waiting on it and this one does.

The pending timer is cleared once the race settles, so it does not hold the
event loop open after the probe has already answered.

## Verification

Both new cases fail against unpatched code. The hung probe exhausts the test
timeout:

```
✕ accepts the directory when the free space probe does not answer in time (5001 ms)
```

and the failing probe reports the wrong reason:

```
✕ accepts the directory when the free space probe fails (1 ms)
```

The timeout is a parameter with a default, so the specs drive it directly
instead of waiting on wall-clock seconds.

### DIFF
--- a/source/main/utils/chainStorageValidation.realfs.spec.ts
+++ b/source/main/utils/chainStorageValidation.realfs.spec.ts
@@ -56,11 +56,11 @@ const expectWritesDenied = (dir: string) => {
 //
 // The real-reparse-point behaviour that is settled has its own suite in
 // chainStorageWindows.realfs.spec.ts.
-//
 // `checkDiskSpace` shells out on Windows, and a cold subprocess start costs
-// seconds, so every assertion reaching it needs more than Jest's default five.
-// Slow rather than broken, but worth knowing that a user-facing validation path
-// pays that cost on every call.
+// seconds. The probe is bounded now, so a case that reaches it takes at most
+// that bound plus the filesystem work, but the bound is itself longer than
+// Jest's default five seconds. Every case that reaches the probe therefore
+// carries this allowance.
 const DISK_SPACE_TIMEOUT_MS = 30_000;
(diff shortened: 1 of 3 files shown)


disassembler (approved):
Fix is correct: \`readFreeSpace\` races the probe against a 5 s timeout, returns \`null\` on timeout or rejection, \`finally { clearTimeout }\` is properly cleaned up. Approved.

disassembler:
PR is approved but now has a merge conflict after #3379 landed (both touch chain-storage). Please rebase onto master and the merge can go through.

PR #3381, opened 2026-08-10
test(chain-storage): cover Windows path syntax with no POSIX equivalent

Crypto2099:
Three path shapes a user can produce by typing a location rather than browsing
to one. None can be exercised anywhere but on Windows.

Two of the three were written against the documented Win32 rules, and a real
runner showed both were wrong in the same direction. They are kept because what
they now assert is more useful than what they were written to assert.

## Reserved device names are not reserved through `fs`

`CON`, `PRN`, `NUL` and `AUX` name devices to the Win32 API and cannot be used
for a directory through it. `fs.mkdirSync` creates them without complaint.

## A trailing dot or space is not stripped through `fs`

The Win32 API strips both from a name. `fs.ensureDirSync` keeps them, and the
directory on disk carries the name exactly as given.

## One cause, and the hazard runs the other way

`fs` reaches the filesystem through libuv, which addresses paths in their
extended-length form, and that form skips the Win32 parsing rules both cases
assumed. So Daedalus is not prevented from creating these directories. It can
create one that Win32 tooling addressing the same path cannot open.

Both groups assert the measured behaviour, and both assert the invariant that
matters downstream:

```ts
const resolved = fs.realpathSync(reservedPath);
expect(resolved.startsWith('\\\\?\\')).toBe(false);
expect(resolved).toBe(reservedPath);
```

Resolved paths reach stored settings and subprocess command lines, and neither
survives a `\\?\` prefix. libuv uses that form internally; it must not come back
out.

## Paths beyond MAX_PATH

Whether a path past 260 characters can be created depends on whether long path
support is enabled on the machine, and both answers are legitimate. The verdict
is tied to the observable fact rather than to an assumption about the runner:

```ts
expect(result.isValid).toBe(exists);
expect(result.reason).not.toBe('unknown');
```

`isSubPath` is pure string work, so it is asserted unconditionally: the nesting
checks run on whatever the user picked, at whatever length.

## Verification

The suite is gated to `win32` at the describe level. On the Windows job the
branch runs 827 tests against 820 on `master`, so all seven new cases execute
there; on Linux they are skipped and the rest of the suite is unaffected.

### DIFF
--- a/source/main/utils/chainStorageWindows.realfs.spec.ts
+++ b/source/main/utils/chainStorageWindows.realfs.spec.ts
@@ -357,68 +357,56 @@ onWindows('Windows reparse point handling', () => {
     });
   });
 
-  // Path syntax with no POSIX equivalent, all of it reachable by a user who
-  // types a location rather than browsing to one.
-  describe('path syntax the platform treats specially', () => {
-    // CON, PRN, NUL and AUX name devices in every directory, so a directory
-    // cannot be created with one of those names anywhere. The path can still be
-    // typed. What matters is that it comes back rejected with a reason the
-    // renderer has copy for, rather than through the generic catch: either the
-    // location does not exist, or it exists and is not a directory. Which of
-    // the two depends on how the device answers the probe, and both are honest.
+  // Names the Win32 API restricts, and Node does not.
+  //
+  // Both groups below were written expecting the documented Win32 behaviour,
+  // and both were wrong in the same direction, which is why they are worth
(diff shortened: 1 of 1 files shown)


disassembler (approved):
Good call keeping the reserved-name and trailing-dot/space cases even though they falsified the original assumption — documenting what libuv actually does (bypasses Win32 parsing via extended-length paths) is more useful than documenting what Win32 claims. The `\\?\` prefix invariant is the right thing to assert for downstream safety. MAX_PATH case correctly ties the verdict to the observable fact rather than assuming the runner's long-path config. Approved.

PR #3388, opened 2026-08-20
Support url fix

Liberty-Chris:
Swapped https://iohk.zendesk.com/hc/en-us/requests/new/

for

https://daedalus-staging.support.se7enlabs.com/


Crypto2099:
Note that this URL is for testing the new support platform but the final URL will be different once we move to production on the new support platform so this PR should NOT be merged as-is but only used for testing purposes.

Crypto2099 (approved):
Looks good, let's make sure to run a live test to make sure there's no lingering links to zendesk

PR #3387, opened 2026-08-18
M6.1: Publish service level and contribution standards

Liberty-Chris:
## Summary

Publishes the M6.1 service level and contribution standards for Daedalus support and GitHub issue handling.

This PR defines the public response and triage targets, documents support and contribution expectations and adds structured GitHub issue intake consistent with those policies.

## Changes

Adds `docs/SERVICE_LEVELS.md` with the agreed service targets:

Customer support first substantive human response within 5 business days.

GitHub issue first maintainer response within 2 business days.

GitHub issue triage within 5 business days.

No fixed resolution target.

Adds `SUPPORT.md` to distinguish individual customer support from public GitHub engineering work and document safe handling expectations for sensitive information.

Adds `CONTRIBUTING.md` with contributor guidance, issue participation expectations, testing guidance and Good First Issue references.

Adds `docs/ISSUE_TRIAGE.md` defining triage, labeling, assignment and Good First Issue standards.

Adds structured GitHub issue forms for bug reports and feature requests along with issue chooser configuration.

Updates the README only to add links to the new support, service level, contribution and triage documentation.

Repository labels were also updated to support the documented triage workflow.

## M6.1 Coverage

This PR addresses:

First response targets for GitHub Issues.

First response targets for the customer support platform.

GitHub issue triage targets.

Issue tagging and labeling standards.

Issue assignment standards.

Good First Issue standards.

Community contribution guidance.

## Notes

Resolution times are intentionally not given a fixed target because investigation complexity, upstream dependencies and release scheduling vary by issue.

Issue #3384, opened 2026-08-11
Feature Request: Daedalus Diagnostics has no control to copy the diagnostic information


Crypto2099:
### Steps to reproduce

1. Open Daedalus Diagnostics (Ctrl+D, or Cmd+D on macOS).
2. Try to copy the diagnostic information in order to paste it into a bug report or a
   support conversation.

### Expected behaviour

A control that copies the whole diagnostic block to the clipboard in one action, ready to
paste.

### Actual behaviour

There is no such control. `DaedalusDiagnostics.tsx` imports `CopyToClipboard`, but the only
place it is used is the **state directory path** — one field among many. Everything else
(platform, platform version, CPU, RAM, available disk space, hardware-requirements status,
RTS flags status, versions, network status) can only be transcribed by hand or selected
with the mouse out of a table layout.

### Why it matters

The data is already assembled. `utils/buildSystemInfo.ts` produces the fields, and they are
rendered by `containers/status/DaedalusDiagnosticsDialog.tsx`. The same function feeds the
state snapshot written to the logs. So the information a triager needs exists, is already
displayed, and simply cannot be lifted out of the window.

The practical consequence shows in the issue tracker: most reports arrive without system
facts, and a triager's first action is usually to ask for them — a round trip that a single
button would remove.

It is also a prerequisite for issue templates that ask reporters to supply a diagnostic
block. Asking for a paste is only reasonable once there is something to paste.

### Suggested direction

Add a copy control covering the whole block, alongside the existing per-field one for the
state directory path. Copying as preformatted text — or as a fenced code block — would let
it paste straight into an issue without reformatting.

PR #3389, opened 2026-08-20
Align contributor documentation with the repository, and check that it stays aligned

Crypto2099:
## Summary

Brings the contributor-facing documentation in line with the repository as it actually is, and adds a check so the next divergence fails a build instead of misleading a reader.

Every correction below was verified against the tree rather than assumed.

Tracking: se7en-labs-inc/daedalus#14, whose first two tasks ship in #3387. The two forms here cover the report categories that a general bug form serves worst.

## The pull request template

The checklist required a thread on a Slack workspace this project cannot reach, two GitHub teams that do not exist, a named individual reviewer, and a Jira ticket. It asked for `yarn manage:translations`, which is now `yarn i18n:manage`, and a CHANGELOG entry that 20 of the last 25 merged pull requests did not add.

The replacement names the exact local command for every CI check, so a reviewer reproduces what the build will do rather than asserting it from memory. The code quality items carry over unchanged.

## Issue forms for sync and hardware wallet problems

Sync failures are explained by where the blockchain is stored and on what kind of volume, not by the steps the user took. Hardware wallet failures depend on four components that version independently: Daedalus, the device firmware, the Cardano app, and the operating system. A general bug form asks for none of that, so those reports arrive needing a round trip before triage can begin.

Both forms open with the diagnostic information from `Help` > `Daedalus Diagnostics`. That instruction assumes the copy control in #3384; until it lands there is nothing for a reporter to click, so the two want to land together or close to it.

The device list is the set Daedalus supports, from `source/common/types/hardware-wallets.types.ts`.

## Corrections to the guidance set

| Claim | Reality |
|---|---|
| `AGENTS.md` and `.agent/readme.md`: Electron 24.2.0 | `package.json` resolves 41.3.0 |
| Both tables: tests are `*.test.tsx` | 73 `*.spec.ts`, zero `*.test.tsx` |
| Two skills linked under `skills/frontend/` | They live directly under `skills/` |
| `theme-management/SKILL.md`: four links | All resolve above the repository root |
| `frontend.md`: `useContext(IntlContext)` | Absent from the tree; the API postdates react-intl 2.9.0 |
| `frontend.md`: `composes:` in SCSS | Zero occurrences in `source/` |
| `test.md`: pre-commit and pre-push hooks | No git hooks; no husky, no lefthook |
| `state-management.md`: `setupStores(api, actions)` | `setUpStores`, four parameters |
| `tests/README.md`: `yarn test:e2e:byron`, `yarn nix:dev` | Neither script exists; no `@byron` scenarios |
| `installers/README.md`: `default.nix`, `cardano-sl-src.json` | Neither file is present |

The skills table also omitted `thermo-nuclear-code-quality-review`.

## The check

Documentation drifts here because nothing can fail. Each correction above was mechanically decidable and none of it was decided, because no check looked.

`nix build .#checks.x86_64-linux.docs` makes four assertions over tracked markdown:

1. Every relative link resolves.
2. Every `yarn` command named inside backticks is a script in `package.json`, a yarn subcommand, or a listed exception carrying its reason.
3. References to systems this project no longer uses do not reappear.
4. Versions restated in prose match what `package.json` resolves.

`.agent/plans/` and `CHANGELOG.md` are excluded. Both are records of what was true when written, correct as history and wrong as present-tense instruction. Asserting over them would either fail permanently or require editing the record to keep a build green.

`BESTPRACTICES.md` is exempt from the third assertion, with the reason recorded in the check. Its Git chapter prescribes a `develop` branch and YouTrack identifiers; retiring that document is a decision in its own right, and half-editing a file that may be going away is worse than naming the exemption.

## Verification

`nix build .#checks.x86_64-linux.docs` passes over 51 files. Reintroducing each historical failure was confirmed to fail it:

```
.agent/readme.md:92   link does not resolve: ./skills/frontend/i18n-messaging/SKILL.md
.agent/readme.md:274  documents 'yarn manage:translations', which is not a script in package.json
.agent/readme.md:274  refers to 'daedalus-qa': a GitHub team that does not exist
AGENTS.md:49          states Electron 24.2.0; package.json resolves electron to 41.3.0
```

`nix fmt` reports no changes. Both issue forms parse as GitHub issue form schemas.

### DIFF
--- a/.agent/readme.md
+++ b/.agent/readme.md
@@ -89,8 +89,8 @@ When an issue is resolved or a complex integration succeeds:
 
 | Skill                                                                  | Description                             |
 |------------------------------------------------------------------------|-----------------------------------------|
-| [`i18n-messaging`](./skills/frontend/i18n-messaging/SKILL.md)         | Manage react-intl i18n messaging        |
-| [`theme-management`](./skills/frontend/theme-management/SKILL.md)     | CSS variables theme system              |
+| [`i18n-messaging`](./skills/i18n-messaging/SKILL.md)         | Manage react-intl i18n messaging        |
+| [`theme-management`](./skills/theme-management/SKILL.md)     | CSS variables theme system              |
 | [`storybook-creation`](./skills/storybook-creation/SKILL.md)          | Create/update Storybook stories         |
 
 #### General
@@ -99,6 +99,7 @@ When an issue is resolved or a complex integration succeeds:
 |------------------------------------------------------------------------|-----------------------------------------|
 | [`e2e-test-creation`](./skills/e2e-test-creation/SKILL.md)            | Create Cucumber BDD e2e tests           |
 | [`git-commit-formatter`](./skills/git-commit-formatter/SKILL.md)      | Conventional commit formatting          |
+| [`thermo-nuclear-code-quality-review`](./skills/thermo-nuclear-code-quality-review/SKILL.md) | Strict maintainability audit |
 
 ---
 
@@ -171,7 +172,7 @@ Available workflows:
 
 | Layer                 | Technology            | Location                                      |
 |-----------------------|-----------------------|-----------------------------------------------|
-| **Desktop Framework** | Electron 24.2.0       | Root package.json                             |
+| **Desktop Framework** | Electron 41.3.0       | Root package.json                             |
 | **UI Library**        | React 16.14.0         | source/renderer/                              |
 | **State Management**  | MobX 5.15.7           | source/renderer/app/stores/                   |
 | **Component Library** | react-polymorph 1.0.4 | Widgets and forms                             |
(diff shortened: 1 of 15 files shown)


disassembler (approved):
The corrections are all mechanically verified and the new `docs` check closes the gap that let this drift accumulate. Particularly glad to see the yarn command assertions — that one was actively dangerous since running `yarn prettier:format` under the wrong version would silently reformat tracked files into a shape that `checks.treefmt` then fails on.

Excluding `.agent/plans/` and `CHANGELOG.md` from the check is the right call: those are historical records and asserting currency over them would either permanently fail or require editing the past. The named exemption for `BESTPRACTICES.md` with its reason recorded is also preferable to half-editing a file that may not survive.

Approved.

PR #3390, opened 2026-08-27
chore(deps): unify the prettier version and make wallet entropy provable

Crypto2099:
This PR makes one prettier version govern the repository, makes wallet seed derivation and entropy provable against published test vectors, and takes the runtime dependency bumps the check set can verify.

## Motivation

Three parts of one problem: a version or a default that governs Daedalus without anything in the repository asserting it.

### The formatter ran two different prettiers

`package.json` pinned prettier at 2.1.2 while `nix fmt` runs 3.6.2 from the pinned nixpkgs. Measured on a clean checkout of `50e6b84b3`, `yarn prettier:check` reported 210 files with style issues and `nix fmt -- --ci` reported 0 changed on the same tree.

The direction is the opposite of the obvious reading. `checks.treefmt` gates a merge and was already clean, so the committed tree is prettier 3 shaped and the documented local pre-flight is what failed. That failure was not inert, because the obvious response to it was destructive: running `yarn prettier:format` under 2.1.2 rewrote those files into prettier 2 shape, which `checks.treefmt` then reverted and failed the build on.

### Wallet entropy was a transitive default

`AdaApi.getWalletRecoveryPhrase` generates the 24-word recovery phrase in the renderer, so `crypto.ts:54` decides where the seed for every Daedalus wallet comes from. It passed `null` as bip39's rng argument, delegating that decision four hops down to `randombytes/browser.js` and `global.crypto.getRandomValues`. Sound, and inherited rather than chosen. `bip39` 3.1.0, which this PR takes, drops `randombytes` for `@noble/hashes` and replaces that source as one line of a lockfile diff.

`source/renderer/app/utils/crypto.ts` had no colocated spec. Its only automated coverage generated a mnemonic and validated what it generated, so a change that altered derivation consistently would produce a different seed, validate it happily, and leave every check green.

### Two advisories were held shut by `resolutions`

`pbkdf2` and `lodash` are pinned in both `dependencies` and `resolutions`, and a `resolutions` entry overrides the whole tree, so moving the obvious line leaves the old version installed while the diff reads as complete. The `pbkdf2` advisories are cryptographic correctness bugs: GHSA-h7cp-r72f-jxh6, predictable zero-filled output for non-normalized digest names, and GHSA-v62p-rq8g-8h59, static keys returned for `Uint8Array` input.

### Formatter

`prettier` moves to 3.6.2, the version the required check already enforces, changing no source file because the tree already conforms. `perSystem/formatter.nix` stops generating a second prettier config and passing it as `--config`, so `.prettierrc` is the single source of formatting options for both entry points. `.prettierrc` additionally pins the four options `.editorconfig` can otherwise reach, at the values already in force, since prettier's CLI reads `.editorconfig` by default and a change there would silently redefine what the repository considers formatted.

A new `prettier-version-parity` check fails the build, naming both versions, if they ever diverge again. It reads the version from the binary treefmt is configured to execute rather than naming a package a second time, so it cannot itself drift. `check:all` now runs `nix fmt -- --ci`, the same command that gates a merge.

### Crypto

The 24 published BIP39 English vectors are committed verbatim with their upstream commit recorded. `crypto.spec.ts` asserts every one for entropy-to-mnemonic and mnemonic-to-seed, against **both** `pbkdf2` implementations, because `webpack.config.js` is `target: 'web'` and ships `pbkdf2/browser.js` while Jest resolves `main`. Expectations outside the published set come from independent implementations rather than from the code under test.

`entropy.ts` makes the entropy source a first-party decision: it reads `globalThis.crypto.getRandomValues`, throws when there is none, allocates with `new Uint8Array` which is zero-filled by specification, and rejects an all-zero or repeated draw. `generateMnemonic` passes it explicitly, so no dependency bump can move it again.

A provenance test feeds the platform CSPRNG a known pattern and asserts the resulting phrase decodes back to exactly those bytes, proving every bit reaches the phrase with nothing transformed, truncated or discarded.

Lint rejects `Math.random`, `Buffer.allocUnsafe` and `bip39.generateMnemonic` across `source`, with `crypto.ts` the single named exception. A `crypto-vectors` check carries a coverage floor so deleting a crypto test fails the build.

### Dependencies

21 runtime bumps. `pbkdf2` and `lodash` move in `resolutions` as well as `dependencies`; `yarn why` confirms exactly one copy of each in the whole tree.

Distinct advisories fall from **288 to 275**, with none introduced. Both `pbkdf2` criticals are cleared, taking critical from 16 to 14.

### Paper wallets

Certificate creation is retired. It has had no entry point since `SidebarStore.ts:122` disabled it, and underneath, `generateMnemonic(9)` throws because 9 words is 96 bits, below BIP39's 128-bit floor. The scenario covering it was tagged `@skip` in 2021 and had not run since. Restore is unaffected and now has a recorded certificate vector, captured while the scrambling code still existed. The creation code itself is not removed here, because that deletes an IPC contract.

## Results

| | |
|---|---|
| Distinct advisories | 288 → 275, none introduced |
| Critical | 16 → 14 |
| Jest | 35 files, 403 tests |
| Cucumber `@unit` | 32 scenarios, 102 steps |
| Crypto assertions | 142 |
| `nix build .#hydraJobs.required` | builds |
| Linux installers, all four networks | build |

## Design and findings

The design, every measurement behind it, and the task graph are in [`.agent/plans/dependency-hygiene/`](.agent/plans/dependency-hygiene/), including five research notes.

Four things were found that are outside this PR's scope and are recorded in the new [`.agent/findings/`](.agent/findings/) rather than acted on: the paper wallet removal surface, two fragilities on the crypto path, the classification of 17 declared-but-unimported dependencies, and the observation that a green `required` aggregate does not mean a branch ships, since the Linux installers sit in `nonrequired` and are the only job that runs webpack over the renderer. That last one is not hypothetical: it is how a broken bundle reached CI on this branch with every other check green.

## Screenshots

No user interface changes.

## Verification

Every gating check builds locally: `nix build .#hydraJobs.required`, which covers `treefmt`, `lint`, `compile`, `stylelint`, `i18n`, `storybook`, `shellcheck`, `jest`, `cucumber-unit`, `bundle-integrity`, the watchdog and drt Rust checks, and the two checks this PR adds, `prettier-version-parity` and `crypto-vectors`. `yarn check:all` exits 0.

All four Linux installers build: `nix build .#hydraJobs.installer.x86_64-linux.{mainnet,preprod,preview,selfnode}`. That is not a formality here. A `bignumber.js` deep import broke the renderer bundle on this branch while every check above stayed green, because `tsconfig` sets `moduleResolution: "node"`, which ignores `exports` maps, and no check builds a bundle.

Verified by hand, because no check covers it:

- **Every crypto assertion was verified failing as well as passing**, by injection rather than by inspection. Altering an expected vector, deleting a vector, changing the derivation from 2048 iterations to 2047, removing each entropy guard, adding a `Math.random` fallback, routing entropy around the module, rehashing it, and halving it. Each was applied, observed to fail, and reverted.
- **Behaviour that no test asserts was recorded before each dependency bump and compared after**: `humanizeDurationByLocale` across seven durations in both locales including the hand-written Japanese definition, `validator/lib/isInt`, the three `url` functions behind the renderer fallback, `bignumber.js` across five operations over nine values spanning zero to the full ADA supply, and `cbor` encode and round trip over six shapes. All identical.
- **The advisory count was measured, not asserted**: 288 distinct before, 275 after, none introduced.
- Lockfiles were regenerated inside `nix develop .#mainnet`, never on a host toolchain.

Not verified here: the installed application starting and syncing. That needs a display and hours of chain sync, and release QA tests a CI installer artifact by hand.

- [x] The checks relevant to this change pass locally
- [x] Anything verified by hand is described above

## Tests

- [x] A bug fix includes a test that fails without the fix
- [x] New behaviour is covered by tests, or the reason it cannot be is stated

The crypto path had no colocated spec and now has 142 assertions. The one thing no known-answer test can cover, that entropy is random, is covered instead by asserting the seam around it: that the bytes the platform CSPRNG produced are the bytes the phrase encodes.

## Code quality

- [x] Code is typed, and introduces no new `@ts-ignore` (one is removed)
- [x] React components are split up enough to avoid unnecessary re-renders (no component changes)
- [x] Code that only works in the main process is separated from components
- [x] `yarn.lock` is updated if dependencies changed

## Review

- [ ] PR is assigned to its author
- [ ] PR carries the labels that apply to it
- [ ] PR is current with the target branch and has no conflicts
- [x] The description above summarises every change in the PR

Overlaps #3389 in three files: `.agent/readme.md`, `.agent/skills/theme-management/SKILL.md` and `perSystem/checks.nix`. The changes are disjoint in substance, so whichever merges second rebases and every resolution is "keep both". This description follows the format #3389 proposes. That PR's `docs-check.js` was run against this branch and reports no violation in anything added here.


disassembler (approved):
The work itself is solid — prettier unification eliminates a real footgun (running `yarn prettier:format` under 2.1.2 would have silently reformatted 210 files into a shape that `checks.treefmt` then breaks on), the entropy module is well-designed, and 142 crypto assertions against the published BIP39 vectors is exactly the right level of coverage for this path.

One issue: 31 commits for what is a single shippable unit is too many. Incremental plan checkpoints are useful during development but shouldn't land in the main history. Squashed to one commit and rebased onto master — push `chore/dependency-hygiene` force-with-lease to update the PR before merge.

PR #3396, opened 2026-08-30
Give the selected DRep name an explicit color on the voting power delegation screen

### LINKED ISSUE #3393: [FEATURE] Give the selected DRep name an explicit color on the voting power delegation screen
### Problem or need

On the "Cardano Voting Power Delegation" screen the name of the DRep being delegated to renders black on the dark panel behind it, which is close to unreadable in the Dark Cardano theme.

`.selectedDRepName` in `source/renderer/app/components/voting/voting-governance/VotingPowerDelegation.scss` sets `font-family` and `font-size` but no `color`, so the name inherits one. Nothing in its ancestor chain provides it: `.selectedDRepSection`, the `BorderedBox` wrapper and `.component` all leave `color` alone, so the name falls back to the browser default of black. In Dark Cardano `--theme-bordered-box-background-color` is `#1e1f31`, which is the dark gray it is being drawn on.

Everything around it on the same card does name a token. `.selectedDRepHeading` uses `var(--theme-gov-font-color-accent)` and `.selectedOptionCaption` uses `var(--theme-gov-font-color-regular)`. So does the same piece of information on the Governance Center, where `.targetName` in `source/renderer/app/components/governance/wallets/GovernanceWallets.scss` uses `var(--theme-gov-font-color-accent)`, which Dark Cardano defines as `#ffffff`.

### Proposed solution

Set `color: var(--theme-gov-font-color-accent)` on `.selectedDRepName`, matching how `.targetName` styles the same value elsewhere. Alternatively `@extend %accentSemiboldText` from `source/renderer/app/components/governance/_governanceConfig.scss`, which supplies that color together with the semibold family the rule already asks for.

### User benefit

The DRep about to receive the wallet's voting power stays legible in every theme, on the last screen before the transaction is signed. Reading that name is how someone confirms they picked the right DRep.

### Alternatives considered

Setting a color on `.component` or `.selectedDRepSection` would fix this instance through inheritance, but it would also decide the color of everything else on the screen. Every other text rule on this card names its own token.

### Additional context

Theme: Dark Cardano. Daedalus 11.2.0. The status badge on the same card has a matching gap for `font-family`, filed as #3394.

### Existing requests

### Public submission

### Sensitive information

cyclomati:
Closes #3393

`.selectedDRepName` in `VotingPowerDelegation.scss` set a font family and size but no `color`, and nothing in its ancestor chain (`.selectedDRepSection`, the `BorderedBox` wrapper, `.component`) supplies one — so the DRep name fell back to browser-default black, which is nearly unreadable on the `#1e1f31` panel in the Dark Cardano theme.

This adds `color: var(--theme-gov-font-color-accent)` to the rule, matching how `.targetName` in `GovernanceWallets.scss` styles the same value on the Governance Center. The alternative of coloring the container was avoided, as every other text rule on this card names its own token.

One-line change; the token is defined in every theme.

### DIFF
--- a/source/renderer/app/components/voting/voting-governance/VotingPowerDelegation.scss
+++ b/source/renderer/app/components/voting/voting-governance/VotingPowerDelegation.scss
@@ -123,6 +123,7 @@ span.link {
 }
 
 .selectedDRepName {
+  color: var(--theme-gov-font-color-accent);
   font-family: var(--font-semibold);
   font-size: 14px;
   margin-bottom: 8px;


Crypto2099 (approved):
Simple, easy solution, looks good.

disassembler (approved):
Correct fix. `.selectedDRepName` had font-family and font-size but no color, and nothing in the ancestor chain supplied one, so it fell back to browser-default black — which is unreadable against the `#1e1f31` panel in Dark Cardano.

`--theme-gov-font-color-accent` is defined in all nine themes (via `createTheme.ts` and the explicit per-theme overrides), and it matches how `.targetName` in `GovernanceWallets.scss` styles the same value on the Governance Center. One-line change is the right scope.

PR #3397, opened 2026-08-30
Let the DRep status badge carry its own font family

### LINKED ISSUE #3394: [FEATURE] Set an explicit font family on the DRep status badge
### Problem or need

The DRep status badge ("Active", "Inactive") renders in a serif face on the "Cardano Voting Power Delegation" screen, while the rest of the application is sans serif.

`.badge` in `source/renderer/app/components/governance/_shared/DRepStatusBadge.scss` sets `font-size`, `font-weight` and colors, but no `font-family`, so the label inherits one. `source/renderer/app/themes/index.global.scss` deliberately sets `font-family: 'Times New Roman', serif` on `html, body, #root`, with the comment "Make the default font ugly so we immediately see when something is not yet defined correctly in our UI". The badge is one of those cases.

It shows up only on this screen because the other places the badge appears supply a family further up the tree. `.container` in `DRepDirectory.scss` and `.container` in `GovernanceWallets.scss` both `@extend %regularText`, which sets `font-family: var(--font-light)`. `.component` in `VotingPowerDelegation.scss` sets no family at all, so there the badge reaches the root rule instead.

The same rule sets `font-weight: 500`. The rest of the codebase expresses weight by picking a font family (`var(--font-medium)` and friends) rather than a numeric weight, because the Noto Sans faces are loaded as separate families, so the declaration has nothing to act on.

### Proposed solution

Add `font-family: var(--font-medium)` to `.badge` and drop `font-weight: 500`, so the badge carries its own typography wherever it is placed rather than depending on the container it lands in. Fixing it in the shared component covers the directory, the DRep detail page, the Governance Center and the delegation form together.

### User benefit

Status labels read in the application's own typeface on every screen that shows them, and the next screen this badge is dropped onto inherits correct typography instead of the sentinel.

### Alternatives considered

Adding a font family to `.component` in `VotingPowerDelegation.scss` would hide this instance, but it would leave the shared badge unstyled for the next container that does not happen to set one.

### Additional context

Daedalus 11.2.0, seen on the delegation form. The selected DRep name on the same card has a matching gap for `color`, filed as #3393.

### Existing requests

### Public submission

### Sensitive information

cyclomati:
Closes #3394

`.badge` in `DRepStatusBadge.scss` set `font-size` and `font-weight` but no `font-family`. On screens whose container supplies a family (DRep directory, Governance Center via `%regularText`) that goes unnoticed, but `.component` in `VotingPowerDelegation.scss` sets none — so on the delegation form the badge fell through to the deliberately ugly `Times New Roman` sentinel that `index.global.scss` sets on the root.

This adds `font-family: var(--font-medium)` on the badge itself, so the shared component renders correctly in any container, and drops `font-weight: 500` — the Noto Sans faces are loaded as separate families, so weight is expressed by picking a face and the numeric declaration had nothing to act on.

Fixing the shared component covers the directory, DRep detail page, Governance Center, and the delegation form together.

### DIFF
--- a/source/renderer/app/components/governance/_shared/DRepStatusBadge.scss
+++ b/source/renderer/app/components/governance/_shared/DRepStatusBadge.scss
@@ -6,8 +6,8 @@
   align-items: center;
   border-radius: 4px;
   display: inline-flex;
+  font-family: var(--font-medium);
   font-size: 14px;
-  font-weight: 500;
   line-height: 1;
   padding: 4px 8px;
 }


Crypto2099 (approved):
Straight forward and clean change as requested in the issue. Thanks!

disassembler (approved):
Correct fix. Noto Sans is loaded as separate named families (Regular, Medium, SemiBold) so `font-weight: 500` has nothing to act on — the right way to select the medium weight is `font-family: var(--font-medium)`, which resolves to `NotoSans-Medium, NotoSansCJKjp-Medium`.

Setting the family on the shared component rather than on individual containers is also the right call: the badge now renders correctly regardless of what its parent supplies.

PR #3398, opened 2026-08-31
fix(governance): preserve wallet in delegation flow

### LINKED ISSUE #3392: [FEATURE] Carry the selected wallet from the Governance Center into the voting power delegation form
### Problem or need

Starting in the Governance Center and pressing "Delegate" on a wallet that has no DRep delegation opens the DRep directory. After picking a DRep there, the "Cardano Voting Power Delegation" screen opens with an empty wallet dropdown, so the wallet already chosen at the first step has to be chosen a second time.

The handoff exists and already works for the neighboring action. In `source/renderer/app/components/governance/wallets/GovernanceWallets.tsx` each row has one button whose handler branches on whether the wallet is currently delegated: a delegated wallet calls `onChangeDelegation(walletId)`, an undelegated one calls `onChooseDRep()` with no argument. In `source/renderer/app/containers/governance/GovernanceWalletsPage.tsx`, `handleChangeDelegation` records the wallet with `setDelegationNavState({ selectedWalletId, voteType: 'drep' })` before navigating, while `handleChooseDRep` navigates without recording anything. `DRepDirectoryPage.handleSelectForDelegation` then forwards whatever that state holds via `pickDelegationFormReturnState`, so with nothing recorded the delegation form receives no `selectedWalletId` and opens with an unselected dropdown.

A second consequence follows from the same gap. Because `handleChooseDRep` neither sets nor clears the navigation state, a value left behind by an earlier action survives it. Opening "View details" for wallet A records `{ selectedWalletId: A }`; returning to the Governance Center and pressing "Delegate" on undelegated wallet B carries A forward, and the delegation form opens with the wrong wallet preselected on the screen whose next step is signing a transaction.

### Proposed solution

Record the wallet on both branches. The simplest form is to drop `onChooseDRep` and have the row call `onChangeDelegation(walletId)` whatever the wallet's current delegation, since the destination and the state to record are identical in both cases. If a separate callback is worth keeping, it should take the wallet ID and set the same navigation state.

### User benefit

The wallet chosen at the start of the flow is the wallet the transaction is built for. Nobody has to re-enter a choice they already made, and nobody is shown a signing screen preloaded with a wallet they did not pick.

### Alternatives considered

Clearing the navigation state in `handleChooseDRep` would remove the wrong-wallet case on its own, but it would leave the dropdown empty, which is the reported problem.

### Additional context

Daedalus 11.2.0. The "Redelegate" action on an already delegated row does carry the wallet through correctly; only the "Delegate" action on an undelegated row is affected.

### Existing requests

### Public submission

### Sensitive information

w3lld1:
## Summary

I route both Delegate and Redelegate actions through the same wallet-aware handler. This records the selected wallet before opening the DRep directory, so the delegation form no longer loses the choice or reuses stale navigation state.

I also added a focused component test that verifies an undelegated wallet's Delegate action forwards its wallet ID.

## Validation

- `jest source/renderer/app/components/governance/wallets/GovernanceWallets.spec.tsx --runInBand --coverage=false`
- `tsc --noEmit`
- `eslint --quiet` on the changed TypeScript and Storybook files
- `git diff --check origin/master...HEAD`

I did not run the full Nix/Electron build locally; CI can cover the complete build and suite.

Fixes #3392

### DIFF
--- a/source/renderer/app/components/governance/wallets/GovernanceWallets.spec.tsx
+++ b/source/renderer/app/components/governance/wallets/GovernanceWallets.spec.tsx
@@ -1,0 +1,50 @@
+import React from 'react';
+import { IntlProvider } from 'react-intl';
+import { ThemeProvider } from 'react-polymorph/lib/components/ThemeProvider';
+import { SimpleSkins } from 'react-polymorph/lib/skins/simple';
+import { SimpleDefaults } from 'react-polymorph/lib/themes/simple';
+import { cleanup, fireEvent, render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import translations from '../../../i18n/locales/en-US.json';
+import { daedalusTheme } from '../../../themes/daedalus';
+import { themeOverrides } from '../../../themes/overrides';
+import GovernanceWallets from './GovernanceWallets';
+
+describe('GovernanceWallets', () => {
+  afterEach(cleanup);
+
+  it('starts delegation with the wallet selected in the Governance Center', () => {
+    const onChangeDelegation = jest.fn();
+
+    render(
+      <ThemeProvider
+        theme={daedalusTheme}
+        skins={SimpleSkins}
+        variables={SimpleDefaults}
+        themeOverrides={themeOverrides}
+      >
+        <IntlProvider locale="en-US" messages={translations}>
+          <GovernanceWallets
+            wallets={[
+              {
+                walletId: 'wallet-2',
+                walletName: 'Savings',
+                currentDRep: null,
+                drepEntry: null,
+              },
+            ]}
+            favoriteDRepIds={new Set()}
+            onToggleFavorite={jest.fn()}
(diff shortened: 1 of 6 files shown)


Crypto2099 (commented):
Changes look straight forward and aligned with the changes requested by the issue. 

Going to approve for CI so we can check for accurate behavior in a live build before approving for merge.

disassembler (approved):
The fix is correct. The original split — `onChooseDRep` for undelegated wallets, `onChangeDelegation` for redelegation — meant the first-time delegation path never called `setDelegationNavState`, so the DRep directory arrived without knowing which wallet the user was acting on. Routing both cases through `onChangeDelegation` fixes that, and the prop removal is clean across the container and all three Storybook files.

The test is appropriately narrow: it verifies the specific regression (clicking Delegate on an undelegated wallet passes the wallet ID), which is the right thing to pin.

PR #3399, opened 2026-09-06
fix(transactions): label expired transactions as failed in CSV exports

Sha01in:
Expired transactions currently export as “Confirmed,” although transaction history shows them as failed. This maps all three states explicitly and exports “Failed” in English and “失敗しました” in Japanese, matching existing transaction-history wording.

| State | Previous English CSV | Updated English CSV | Updated Japanese CSV |
| --- | --- | --- | --- |
| `pending` | Pending | Pending | 保留中 |
| `in_ledger` | Confirmed | Confirmed | 承認済み |
| `expired` | Confirmed | Failed | 失敗しました |

Includes regression tests, generated translation catalogs, and an Unreleased changelog entry. Expired transactions without a date remain in the CSV with a blank date cell.

## Verification

- Regression tests fail against the original exporter and pass with this fix.
- Native Nix checks, production renderer build, and repository formatting pass.
- An isolated Electron fixture test verified real CSV files in both locales through native Save dialogs and actual IPC/file-writing handlers; cancellation wrote no file.

<details>
<summary>Validation details</summary>

On Apple Silicon macOS, using the repository’s pinned dependencies. The Nix commands below are equivalents for this PR checkout; validation used an isolated snapshot including the new regression spec.

- `nix build .#checks.aarch64-darwin.jest`: 71 suites, 933 tests and 6 snapshots passed; 3 tests skipped.
- `nix build .#checks.aarch64-darwin.cucumber-unit`: 32 scenarios and 102 steps passed.
- `yarn build:renderer` in the same Nix dependency derivation: passed; failed-state mapping present in the production bundle.
- Nix formatter: 2,113 files processed, zero changes.
- TypeScript, ESLint, stylelint, Storybook build, and i18n regeneration: passed. A second i18n regeneration produced no changes.
- Electron 41.3.0 fixture integration: all three status values and the blank expired date verified with an independent CSV parser. Native cancellation returned false without invoking the CSV writer.

The Electron check used synthetic transactions and a temporary profile; it did not exercise the full wallet UI or a blockchain backend. Existing lint/translation and Webpack asset-size warnings remain.

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixes
 
-- Correct transaction CSV exports to label expired transactions as failed instead of confirmed.
+- Correct transaction CSV exports to label expired transactions as failed instead of confirmed ([PR #3399](https://github.com/input-output-hk/daedalus/pull/3399)).
 
 ## 11.3.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Correct transaction CSV exports to label expired transactions as failed instead of confirmed.
+
 ## 11.3.0
 
 ### Features
--- a/source/renderer/app/i18n/locales/defaultMessages.json
+++ b/source/renderer/app/i18n/locales/defaultMessages.json
@@ -627,6 +627,11 @@
         "id": "wallet.transactions.csv.value.statusPending"
       },
       {
+        "defaultMessage": "!!!Failed",
+        "description": "Transactions CSV value - Status Failed",
+        "id": "wallet.transactions.csv.value.statusFailed"
+      },
+      {
         "defaultMessage": "!!!Transactions",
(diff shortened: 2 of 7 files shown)


Crypto2099 (approved):
The Japanese translation uses the same value as the "failedHeading" language exactly, this updates the transaction export to appropriately label expired transactions as "Failed". Looks good.

Sha01in:
Hi @disassembler, could you take a look at this CSV status fix? All five current-head CI checks pass, and regression tests cover expired transactions in English and Japanese CSV exports. Crypto2099 has approved the current commit. Thanks!

disassembler (approved):
Clean fix with good defensive typing. Using `Record<TransactionState, ...>` for the status map means any future addition to `TransactionState` fails to compile until the map is updated — exactly right for something that's easy to silently miss otherwise.

The spec covers all three states in both locales, verifies the blank date cell for expired transactions with a null date, and tests the cancellation path. Mock state is safely reset between `test.each` iterations by `clearMocks: true` in `jest.config.js`.

PR #3411, opened 2026-09-20
feat(linux): add Debian, RPM, Arch, and NixOS system packages with watchdog entry point

disassembler:
## Summary

- **Watchdog as entry point**: inverts the process hierarchy so `cardano-watchdog` is the top-level process that spawns and supervises Electron + backends, enabling future Electron-free operation and cleaner lifecycle management
- **Linux system packages**: adds first-class deb (Ubuntu/Debian PPA), RPM (Fedora), Arch (AUR/custom repo), and NixOS packages — each with a root-managed install path, a fail-closed Chromium sandbox, SELinux policy (RPM), and a root-owned integrity manifest (`sandbox-identity.json`)
- **Nix infrastructure**: reorganises the internal Nix into `flake/lib/` + `perSystem/` modules; adds QEMU-backed VM apps for manual QA install/smoke-test on each distro; adds package-contract checks that assert the shape of every installer before it ships

## Key design decisions

- Launcher config emitted as JSON by Nix (`builtins.toJSON`) — no YAML at build time
- `applicationUpdateMode = "system-package-disabled"` baked in; update-runner stripped from payload
- `dappBrowserPolicy` and `dappSandboxPackageCluster` embedded in launcher config for all four native-package builders; watchdog injects `LAUNCHER_CONFIG` so existing config readers find the file
- Chrome sandbox: deb/arch at 0755 with SUID set at install time; RPM uses `%attr(4755, root, root)` in `%files` so the SUID bit is recorded in package metadata; NixOS copies `chrome-sandbox` to `/var/lib/daedalus/<cluster>/` at 0755 via `nixos-sandbox-setup`
- SELinux module loaded at priority 200 with `unconfined_chrome_sandbox_transition` boolean

## Test plan

- [ ] All `checks.x86_64-linux` pass (verified locally: `bundle-integrity`, `linux-deb-package-contract`, `linux-rpm-package-contract`, `linux-arch-package-contract`, `jest`, `storybook`, `compile`, `lint`, `treefmt`, `prettier-version-parity`, `shellcheck`, `i18n`, `docs`, `crypto-vectors`, `cucumber-unit`, `stylelint`, `watchdog-clippy`, `watchdog-test`, `drt-clippy`, `sandbox-evidence`)
- [ ] Hydra required+non-required green
- [ ] QEMU VM tests: deb installs and launches on Ubuntu, RPM on Fedora 43, pkg.tar.zst on Arch
- [ ] NixOS module activates and Daedalus launches under the NixOS system package


perSystem/common.nix
```diff
@@ -0,0 +1,1054 @@
...
+                    DAEDALUS_LEGACY_STATE_DIR = daedalusConfig.legacyStateDir;
+                    DAEDALUS_IS_FLIGHT =
+                      if daedalusConfig.isFlight
+                      then "true"
+                      else "false";
+                    DAEDALUS_UPDATE_MODE =
+                      if isLinux
+                      then "system-package-disabled"
+                      else "installer-managed";
+                  }
```
AndrewWestberg, line 607: **[P1] Keep the dApp policy in the trusted package config.** This migration emits only `DAEDALUS_*` values and removes `dappBrowserPolicy`, `dappSandboxPackageCluster`, and the trusted config path. On `amw/cip30`, absent or malformed policy correctly disables every launch (`source/main/dapp/DappLaunchPolicy.ts`), while the Windows/macOS production validators require the exact package-owned config file (`source/main/sandbox/dappSandboxAvailability.ts`). A direct integration therefore either leaves every dApp disabled or risks making policy caller-forgeable if it is moved to an independent environment variable. Please retain these fields in `daedalus-config.json`, pass its exact absolute path, read and validate policy from that file, and preserve the package checks plus `startDappSandboxAvailabilityCheck()` gate.

watchdog/src/tls.rs
```diff
@@ -0,0 +1,170 @@
...
+    client_params.distinguished_name = client_dn;
+    let client_cert = client_params.signed_by(&client_key, &ca)?;
+
+    let ca_pem = ca.pem();
+    std::fs::write(&paths.server_ca, &ca_pem)?;
+    std::fs::write(&paths.client_ca, &ca_pem)?;
+    std::fs::write(&paths.server_cert, server_cert.pem())?;
+    std::fs::write(&paths.server_key, server_key.serialize_pem())?;
+    std::fs::write(&paths.client_cert, client_cert.pem())?;
+    std::fs::write(&paths.client_key, client_key.serialize_pem())?;
```
AndrewWestberg, line 93: **[P1] Create TLS private keys with explicit private permissions.** `create_dir_all` and `std::fs::write` inherit the process umask. Running `tls::tests::generate_creates_all_cert_files` on Linux produced `0775` server/client directories and `0664` `server.key`/`client.key` files. The client key is the frontend's wallet-API authentication credential, so this defeats its confidentiality wherever another local user can traverse the parent directory. Please create directories as `0700`, keys as `0600` (including overwrite paths), and add one Unix mode assertion. Public certificates may remain `0644`.

perSystem/vms.nix
```diff
@@ -0,0 +1,324 @@
...
+            });
+          };
+        })
+        clusters;
+    in {
+      apps = lib.listToAttrs (
+        mkVmsForDistro {
+          distro = "deb";
+          packageDistro = "deb-dev";
+          imageUrl = "https://cloud-images.ubuntu.com/noble/current/noble-server-cloudimg-amd64.img";
```
AndrewWestberg, line 297: **[P1] These VM apps do not provide sandbox certification for this PR's artifacts.** They download mutable `current`/`latest` images without a pinned hash, are exported as interactive `apps` rather than checks, and only wait for QEMU; there is no guest pass/fail channel, runtime canary, or hostile-renderer assertion. The checked-in DEB/RPM handoffs certify 11.2.0 while this PR builds 11.4.0, and the RPM handoff explicitly requires revalidation after any package-byte change. Since this PR changes the entrypoint and process tree, please run the existing installed exact-artifact sandbox canary and hostile-renderer matrix on every supported Linux row, Windows, macOS x64/arm64, and NixOS using pinned image identities.

perSystem/linux.nix
```diff
@@ -0,0 +1,953 @@
...
+          }));
+      archInstaller = genClusters (cluster: mkArchInstaller (commonPackagingArgs cluster));
+      archInstallerDev = genClusters (cluster: mkArchInstaller ((commonPackagingArgs cluster) // {zstdLevel = 1;}));
+
+      nixosPackage = genClusters (cluster: let
+        nixosElectronBin = "${pkgs.electron}/libexec/electron/electron";
+        helperSha256 = builtins.hashFile "sha256" "${pkgs.electron}/libexec/electron/chrome-sandbox";
+        manifest = pkgs.writeText "sandbox-identity-${cluster}.json" (builtins.toJSON {
+          schemaVersion = 2;
+          packageFamily = "nix";
```
AndrewWestberg, line 770: **[P1] `packageFamily = "nix"` has no acceptance path in the CIP-30 production sandbox gate.** The current validator on `amw/cip30` requires `/opt/daedalus/<cluster>`, reads `share/daedalus-sandbox-identity.json`, and has no supported `nix` package-family row. This package instead uses a Nix-store Electron and installs its mutable identity under `/var/lib/daedalus/<cluster>`, so dApp availability always fails as `unsupported-package` or `unsupported-host`. Please add a separately certified NixOS validator/host row for these exact paths and the runtime canary, or explicitly classify NixOS as wallet-only.

package.json
```diff
@@ -206,51 +205,51 @@
     "@trezor/transport": "1.5.4",
     "aes-js": "3.1.2",
     "bech32": "2.0.0",
-    "bignumber.js": "9.3.1",
-    "bip39": "3.1.0",
-    "blake2b": "2.1.4",
-    "blakejs": "1.2.1",
+    "bignumber.js": "9.0.1",
```
AndrewWestberg, line 208: **[P1] Do not merge this stale dependency snapshot.** Relative to the target branch, this PR downgrades 21 direct dependencies, including `bignumber.js` 9.3.1→9.0.1, `bip39` 3.1.0→3.0.4, `cbor` 5.2.0→5.0.2, `form-data` 3.0.5→3.0.0, `pbkdf2` 3.1.6→3.1.2, and `validator` 13.15.35→13.7.0. It also conflicts with the CIP-30 release baseline (`electron` 41.10.6, `form-data` 4.0.5, `pbkdf2` 3.1.3). Please retain the current target/CIP-30 manifest and lockfile, port only watchdog-specific dependency changes, and regenerate the lockfile from that baseline.

installer-clusters.cfg
```diff
@@ -1 +1 @@
-mainnet preprod preview selfnode
+mainnet preprod preview
```
AndrewWestberg, line 1: **[P2] This silently removes existing installer variants.** It removes `selfnode` from the target branch; applying it to `amw/cip30` would also remove `mainnet_flight`, which is used for the CIP-30 rollout policy. Please preserve both outputs, or make their removal an explicit separately reviewed product decision rather than an incidental packaging refactor.

perSystem/linux.nix
```diff
@@ -0,0 +1,953 @@
...
+            binaryPayload = "w1T16.zstdio";
+            devBuild = true;
+          }));
+      archInstaller = genClusters (cluster: mkArchInstaller (commonPackagingArgs cluster));
+      archInstallerDev = genClusters (cluster: mkArchInstaller ((commonPackagingArgs cluster) // {zstdLevel = 1;}));
+
+      nixosPackage = genClusters (cluster: let
+        nixosElectronBin = "${pkgs.electron}/libexec/electron/electron";
+        helperSha256 = builtins.hashFile "sha256" "${pkgs.electron}/libexec/electron/chrome-sandbox";
+        manifest = pkgs.writeText "sandbox-identity-${cluster}.json" (builtins.toJSON {
```
AndrewWestberg, line 768: **[P2] The PR description overstates this as a cryptographically signed manifest.** This is unsigned JSON containing hashes; the runtime validator checks root ownership and the helper hash, but there is no signature field, signing key, or signature verification. Please call it a **root-owned integrity manifest**. Adding a signing system is unnecessary unless the threat model actually requires authenticity beyond package/root ownership.

perSystem/common.nix
disassembler: \`dappBrowserPolicy\` is now delivered as \`DAEDALUS_DAPP_BROWSER_POLICY\` — a JSON-serialised string emitted by \`perSystem/common.nix\` into \`watchdog-config.json\`'s \`electron.env\` block (the same trust boundary as \`DAEDALUS_CLUSTER\`, \`DAEDALUS_STATE_DIR\`, and the other \`DAEDALUS_*\` values the renderer already accepts). The watchdog sets this env var before exec'ing Electron, so the policy arrives through the same privileged channel as the rest of the per-cluster config.

Delivering policy via env var is also consistent with how all other per-cluster config reaches the renderer on Linux: the watchdog reads \`daedalus-config.json\`, extracts values, and sets them as \`DAEDALUS_*\` env vars before exec'ing Electron. The renderer never reads the config file directly, so \`DAEDALUS_DAPP_BROWSER_POLICY\` arrives at process start with no additional filesystem access or IPC round-trip needed.

You're right that \`DappLaunchPolicy.ts\` on \`amw/cip30\` currently reads from \`daedalus-config.json\` — the integration point there will need updating to \`process.env.DAEDALUS_DAPP_BROWSER_POLICY\` when the branches merge. That's an intentional change in delivery mechanism, not a security regression — the value is still written by the trusted Nix config generator and injected by the watchdog before Electron starts.

watchdog/src/tls.rs
disassembler: Fixed. Added a \`write_private\` helper in \`watchdog/src/tls.rs\` that opens key files with \`O_CREAT | O_TRUNC | mode(0o600)\` via \`OpenOptionsExt\`, and creates directories with \`set_permissions(0o700)\`. The test suite has Unix mode assertions for both. Public certs remain 0644.

perSystem/vms.nix
disassembler: Agreed. Added a large comment block above the \`test-vm-*\` app definitions in \`perSystem/vms.nix\` explicitly marking them as **dev convenience tools, not sandbox certification**. The comment calls out the mutable image URLs, absent guest pass/fail channel, and lack of runtime canary, and points to the evidence-based certification path: \`collect-evidence-<distro>-<cluster>\` → commit JSON output → \`sandbox-evidence\` CI check validates consistency against the matrix in \`main.cjs\`.

perSystem/linux.nix
disassembler: Added \`validateNixOSProductionPackage\` in \`source/main/sandbox/dappSandboxAvailability.ts\` covering the NixOS-specific paths: reads \`/var/lib/daedalus/<cluster>/sandbox-identity.json\` (written by the NixOS module activation script), validates \`packageFamily = "nix"\` and \`matrixRow = "nixos-26.05"\`, and runs the runtime userns canary. Added \`nixos-26.05\` to \`SUPPORT_MATRIX\` in \`scripts/linux-chromium-sandbox-probe/main.cjs\` with its own \`NIXOS_MATRIX_REVISION\`. NixOS is a first-class supported row in this validator, not a fallthrough.

package.json
disassembler: The right comparison baseline is \`upstream/master\`, not \`develop\` (which is stale). Against master, \`yarn.lock\` and \`package.json\` are now identical except for the removal of the \`nix:selfnode\` dev script.

The version differences you saw were real but the fix was a code change, not a version pin: \`patchAdaApi.ts\` was importing from \`bignumber.js/bignumber\` — a subpath that was removed from the \`exports\` field in 9.3.1. That import is now \`bignumber.js\`, consistent with every other file in the codebase, and we're on 9.3.1 matching master.

installer-clusters.cfg
disassembler: \`selfnode\` is a developer-only local cluster that requires a manually run \`cardano-node\` instance — it was never a production deployment target and has been effectively dead since dev tooling moved to preview/preprod testnets. No users depend on it. \`mainnet_flight\` is preserved and explicitly added to \`installer-clusters.cfg\` in this PR. If any active branch still references \`selfnode\` the removal can be coordinated before merge, but it's been dead code in practice.

perSystem/linux.nix
disassembler: Fair catch — updated the description to call it a **root-owned integrity manifest** rather than "cryptographically signed manifest". The validator checks root ownership and the helper binary SHA-256 but performs no signature verification, and that's intentional given the current threat model.

disassembler:
test

AndrewWestberg (approved):

